PR #4, opened 2020-11-12
TILA-82 Static analysis configuration and dependencies.

aprenfors:
Most of the changes in requirements are downstream from adding pytest dependency.


setup.cfg
```diff
@@ -0,0 +1,23 @@
+[flake8]
+exclude = migrations
```
aprenfors, line 2: Added some sensible defaults here, feel free to discuss what these values should be, like line length for example etc.

aprenfors:
I'll run the actual formatting in separate merge request so it's easier to see the actual manual changes here.

tommimanbytes (approved):

PR #6, opened 2020-11-13
TILA-45 | Resolve deprecation warnings

otiala:
- Implement REST/GQL API's
- Resolve deprecation warnings


aprenfors (commented):
To me this looks very clean and easy to read, but since i'm a bit out of touch with django best practices better if someone else reviews too. There is a conflict in requirements.

aprenfors:
This is more of a discussion starter about coding standards, than comment on this pull request. How do people feel about using type hints? How about using some fluent assertion library for testing, for example assertpy or something similar, imo it makes tests more readable, easier to do soft asserts when needed etc.

aprenfors (approved):
Looks clean to me

PR #9, opened 2020-11-17
TILA-94 | Application model and supporting models.

aprenfors:
This still needs some work, mainly maybe some of stuff could be in more general module instead of applications for better flow of dependencies, but let's create this pull request to have some discussion whether this is even close to what's expected or not of application model.


applications/models.py
```diff
@@ -1,3 +1,132 @@
...
+
+class Application(Applicant):
+
+    description = models.fields.TextField(
+        verbose_name=_("Description"), max_length=1000, blank=True, null=True
+    )
 
-# Create your models here.
+    reservation_purpose = models.ForeignKey(
+        to="reservations.ReservationPurpose",
```
aprenfors, line 127: This feels a bit hacky, with regular import there is a circular dep. Is there better way to handle this, like should these common stuff be in separate module or is this ok?

applications/models.py
```diff
@@ -1,3 +1,132 @@
+import datetime
+
 from django.db import models
+from django.contrib.auth.models import User
+from django.utils.translation import gettext_lazy as _
+from rest_framework.exceptions import ValidationError
+
```
aprenfors, line 7: Some stuff in here could maybe be abstract instead of separate tables.

reservations/migrations/0002_auto_20201118_0912.py
```diff
@@ -0,0 +1,50 @@
+# Generated by Django 3.0.10 on 2020-11-18 09:12
```
tommimanbytes, line 1: I suggest we use same system as we used with Respa with migration naming.

Leave the number at start as is and after that give it a describing name for the migrations for example:

0002_add_fields_to_recurring_reservations.py

applications/contact_information.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
```
tommimanbytes, line 1: Maybe we should have one file per app for abstract base models? If we add many of these base / abstract models then file structure gets pretty messy. Also usually in Django projects I have seen filenames describes type of entities the file has (for example models, views, serializers) rather than name of the class it contains.

I suggest something like base_models.py

tommimanbytes (changes requested):

tommimanbytes (approved):

PR #10, opened 2020-11-18
TILA 94 | event model

aprenfors:


aprenfors:
This is missing area related information. Saw that Tommi was doing something related to that in another branch, so maybe wait for that before implementing so it'll be uniform.

requirements.txt
```diff
@@ -7,8 +7,8 @@ click==7.1.2
 Django==3.0.10
 django-environ==0.4.5
 django-extensions==3.0.9
+django-recurrence==1.10.3
 djangorestframework==3.11.1
-django-environ==0.4.5
```
aprenfors, line 11: There were some duplicate things in requirements, so these removals are from that. Probably from some merge conflict or something where pip freeze wasn't used.

tommimanbytes:
> This is missing area related information. Saw that Tommi was doing something related to that in another branch, so maybe wait for that before implementing so it'll be uniform.

I made a pull request of the changes to districts, spaces etc. They may not be perfect yet for milestone of december but we can  merge it for now so we can finish this PR.

tommimanbytes (commented):
Otherwise this PR looks good to me but have you confirmed that django-recurrence library is coherrent enough to be used here? I checked it out and it looked nice, but it isn't maintained very often and there are over 40 issues reported.

If you think it's all good its ok by me. I'm just usually pretty careful when using model fields (which determines how db is created) from 3rd party packages because they could cause some problems in future I think.

aprenfors:
> Otherwise this PR looks good to me but have you confirmed that django-recurrence library is coherrent enough to be used here? I checked it out and it looked nice, but it isn't maintained very often and there are over 40 issues reported.
> 
> If you think it's all good its ok by me. I'm just usually pretty careful when using model fields (which determines how db is created) from 3rd party packages because they could cause some problems in future I think.

At least the library is something to start with so we don't need to create our own implementation. If there are problems with it it should be relatively simple to get rid of it, since it's just a django wrapper for dateutil.rrule. And if there is problems with even that, it uses rfc2445 standard for the data format https://tools.ietf.org/html/rfc2445 so there could be alternative implementations or write one from scratch.

tommimanbytes:
> > Otherwise this PR looks good to me but have you confirmed that django-recurrence library is coherrent enough to be used here? I checked it out and it looked nice, but it isn't maintained very often and there are over 40 issues reported.
> > If you think it's all good its ok by me. I'm just usually pretty careful when using model fields (which determines how db is created) from 3rd party packages because they could cause some problems in future I think.
> 
> At least the library is something to start with so we don't need to create our own implementation. If there are problems with it it should be relatively simple to get rid of it, since it's just a django wrapper for dateutil.rrule. And if there is problems with even that, it uses rfc2445 standard for the data format https://tools.ietf.org/html/rfc2445 so there could be alternative implementations or write one from scratch. So the data looks something like this
> 
> ` DTSTART;TZID=US-Eastern:19970902T090000 RRULE:FREQ=WEEKLY;UNTIL=19971224T000000Z;INTERVAL=2;BYDAY=1,3`

Okay, nice. Just checked that this is a considered decision. Let me know when this is ready for final review and merge.

tommimanbytes (approved):

PR #11, opened 2020-11-19
TILA-87 | Changes and additions to space models

tommimanbytes:
Made changes and additions to space models to support district and neighborhood systems of Helsinki. Space can now be directly related to districts and neighborhoods.


spaces/models.py
```diff
@@ -3,7 +3,45 @@
 
 
 class District(models.Model):
+    TYPE_MAJOR_DISTRICT = "major_district"
+    TYPE_DISTRICT = "district"
+    TYPE_SUB_DISTRICT = "sub_district"
+
+    TYPE_CHOICES = (
```
aprenfors, line 10: Just curious if these are used as way to restrict the parent child hierarchy? Like only major district can be parent of district and so on.

spaces/models.py
```diff
@@ -3,7 +3,45 @@
...
+        (TYPE_MAJOR_DISTRICT, _("Major district")),
+        (TYPE_DISTRICT, _("District")),
```
aprenfors, line 12: So i assume these are pretty much set in stone? Like that there only exists district, neighborhood etc? So there is no need to make this super flexible and just use one space/place model, and with attributes in the model control whether it's a district, neighborhood or rentable space or whatever so we could have pretty much infinite hierarchy?

tommimanbytes: Districts, neighborhood, real estate and building are models for mostly search, filter and information display purposes. Major districts, Districts and Subdistricts can be used to filter reservation units according to Jira documents. 

[Subdivisions_of_Helsinki](https://en.wikipedia.org/wiki/Subdivisions_of_Helsinki). According to this wiki page neighborhoods and districts are different system. They usually overlap but sometimes don't.

**I guess neighborhood could also be one type of district model. I will change it so.**

Space is only entity here that can be attached to reservation unit. Space can be a whole building, room of building, part of room of building, outside area etc.

spaces/models.py
tommimanbytes: This is pretty much fixed hierarchy in district system of Helsinki. Validation you described would represent the real world scenario and could be implemented but its probably not mandatory feature because districts are probably created once (programmatically maybe) and then only changed if changes happens like once in 10 years or something.

spaces/models.py
tommimanbytes: If you want to further discuss this we can take slack call or google meet. Its pretty hard to talk about this thoroughly here.

aprenfors: No need to, this is enough information 👍

spaces/models.py
aprenfors: Yeah, didn't mean this would necessarily need validation at least in the current PR

aprenfors (approved):

PR #12, opened 2020-11-19
TILA-88 | Data model for application periods

tommimanbytes:
This is a model that represents a round of applying for certain reservation units for certain periods of time.


applications/models.py
```diff
@@ -83,6 +84,44 @@ class Organisation(models.Model):
     )
 
 
+class ApplicationRound(models.Model):
```
aprenfors, line 87: Round is a bit finglish, maybe something like ApplicationPeriod or maybe ApplicationCycle if you want to preserve something close to "kierros"

applications/models.py
```diff
@@ -83,6 +84,44 @@ class Organisation(models.Model):
...
+    )
+
+    reservation_period_begin = models.DateField(
+        verbose_name=_("Reservation period begin"),
+    )
+    reservation_period_end = models.DateField(
+        verbose_name=_("Reservation period end"),
+    )
+
+    purposes = models.ManyToManyField(
```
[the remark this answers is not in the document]
tommimanbytes: It has been mentoined in discussions that there will be certain purposes for the application rounds so I thought that this could be used for displaying them and also maybe filtering reservation units with these purposes when staff is creating the application round.

I thought that it would be better practice to choose purposes for the application period and then choose reservation units for it rather than first choosing from all reservation units which may have many purposes and then determine which of those purposes to use for this application round.

applications/models.py
tommimanbytes: Yeah, this was initially ApplicationPeriod but I changed it to ApplicationRound to include kierros in the name so there would be similarity in discussions and code concepts. I didn't think that ApplicationRound is too stupid sounding but if you think so I can change it back to ApplicationPeriod for example.

aprenfors: Yeah, if you want to include kierros, then i think Cycle is closer to the Finnish meaning of kierros in this context than round.

applications/models.py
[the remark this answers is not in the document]
aprenfors: Yeah, that makes sense.

applications/models.py
aprenfors: But i'll approve this, change it you want. Can be changed later if it bothers people.

aprenfors (approved):

applications/models.py
tommimanbytes: I changed it to application period. Now merging.

PR #13, opened 2020-11-20
Tila 95 | application api

aprenfors:
These, especially application event and recurrence still need some work, but didn't want to mix model changes into this merge request so we can keep this relatively small. Namely there is at least acceptable time frame missing from the recurrence information. 

ApplicationEvents now produce this kind of data with all the recurrences for the recurring event listed. 

```javascript
[
    {
        "id": 1,
        "recurrences": [
            {
                "id": 1,
                "priority": 200,
                "recurrence": [
                    "1997-09-02T09:00:00Z",
                    "1997-09-03T09:00:00Z",
                    "1997-09-04T09:00:00Z",
                    "1997-09-05T09:00:00Z",
                    "1997-09-06T09:00:00Z",
                    "1997-09-07T09:00:00Z",
                    "1997-09-08T09:00:00Z",
                    "1997-09-09T09:00:00Z",
                    "1997-09-10T09:00:00Z",
                    "1997-09-11T09:00:00Z"
                ]
            }
        ],
        "num_persons": 12,
        "age_group": null,
        "ability_group": null,
        "num_events": 12,
        "duration": "00:00:23",
        "application": 1
    }
]
```


api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+        for field in self.fields:
+            recurrence[field] = getattr(data, field)
+
+        recurrence["recurrence"] = data.recurrence.between(
+            self.get_start_date(),
+            self.get_end_date(),
+        )
+        return recurrence
+
+    def get_start_date(self) -> datetime:
```
aprenfors, line 116: So this and get end date have some problems. The start (context) is set in the ApplicationEventViewSet#get_request_params

It works perfectly when i call /applications/ to fetch all the applications but when i do applications/id to fetch single application. I get like 4 extra times that end up in 

return MINIMUM_TIME

branch even start is set correctly in ApplicationEventViewSet. The first call to get_start_date has start date in context, but for some reason there are some extra calls to this method where the context is not set. Even with those extra calls the api works just fine and returns recurrences only between given start and end dates. Can't figure out where those extra calls come. 

And as i said there are extra call only when fetching single one, not when fetching all. And in both cases api returns correctly.

aprenfors:
This is missing tests. I'll write them in new merge request or here once i have verified with you guys that this makes some sense.

api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+
+    queryset = Application.objects.all()
+
+    serializer_class = ApplicationSerializer
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
+
+
+class RecurrenceReadSerializer(serializers.ModelSerializer):
```
aprenfors, line 104: Everything until this point seems relative straightforward, so from here on is stuff i'm not really sure on. If there is a better way to do this, please inform me, i have never worked with django rest framework, so might have missed something very obvious.

api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+        if end is not None:
+            end = parse(end)
+
+        return {"start": start, "end": end, "read": self.is_read()}
+
+    def retrieve(self, request, pk=None):
+        queryset = ApplicationEvent.objects.all()
+        user = get_object_or_404(queryset, pk=pk)
+        serializer = ApplicationEventSerializer(user, context=self.get_request_params())
+        return Response(serializer.data)
```
indigane, line 176: Is this only for passing context? If so, it's better to override only the `get_serializer_context` method of `ModelViewSet`.

I'll also note that unless there's a good reason not to, you should always get queryset with `self.get_queryset()` or get the object with `self.get_object()` etc.

http://cdrf.co helps a lot as well.

api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def perform_create(self, serializer):
+        serializer.save()
+
+
+class AddressSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Address
+
+        fields = "__all__"
```
indigane, line 47: Not sure if this is work-in-progress, but note that using `__all__` will make everything in the database table readable _and_ writable in the API by default (at least in the case of the `ModelViewSet` below). This may cause problems for example if someone adds new backend-only fields and is not aware that they're automatically made public and possibly writable.

api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+    queryset = Application.objects.all()
+
+    serializer_class = ApplicationSerializer
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
+
+
+class RecurrenceReadSerializer(serializers.ModelSerializer):
+    def to_representation(self, data):
```
indigane, line 105: The second parameter for `to_representation()` is `instance`, so using `data` confused me for a while, as that's what's used in `to_internal_value()` where it's a `dict`. Here `instance` should be a model instance.

api/applications_api.py
indigane: If we use `RecurrenceField` elsewhere as well, we should consider writing a custom serializer field for it: https://www.django-rest-framework.org/api-guide/fields/#custom-fields instead of handling the field in the serializer that uses it.

aprenfors: Ok, i'll take a look into that.

api/applications_api.py
aprenfors: 👍

api/applications_api.py
aprenfors: Yeah, leftover from some quick testing, will fix.

tommimanbytes (changes requested):

api/applications_api.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class OrganisationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Organisation
+
+        fields = ["id", "name", "identifier", "year_established"]
+
+
+class OrganisationViewSet(viewsets.ModelViewSet):
```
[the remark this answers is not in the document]
aprenfors: I guess it depends how we want to do this, partially how the process will go, like will it be more like a wizard you can continue later with. And partially how we want to actually implement the api, to my understanding django rest is kind of tightly coupled to underlaying model structure, and api proposal by Oskari says you should not write nested writhing serializers. It's probably possible to flatten the serializer and deserialize from that to models though. 

But i guess it's just a decision, which way to do.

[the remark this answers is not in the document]
aprenfors: Looks like the process seems to flow with steps, the video in tilanvaraus tiimi channel. Whether we want to save between each step or not is of course another thing.

tommimanbytes (approved):

PR #19, opened 2020-11-27
Tila 58 | devops pipeline

aprenfors:
Docker configuration for production and development environments.


Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Served by whitenoise middleware
+ENV STATIC_ROOT /srv/static
+
+ENV PYTHONUNBUFFERED True
+
+RUN apt-get update && apt-get install -y postgresql-client gdal-bin
+
+RUN pip install --no-cache-dir uwsgi
+# Sentry CLI for sending events from non-Python processes to Sentry
```
siren, line 21: I am just curious: is the sentry CLI actually used somewhere?

siren (approved):
I don't understand about the internals, but externally the docker-compose still works and starts the api -> works for me :+1:

Dockerfile
aprenfors: Not yet, but we'll use sentry in future so might as well.

PR #22, opened 2020-11-30
TILA-90 | Reservation unit filters

otiala:
Create the following filters and necessary tests for reservation unit API:
- purpose
- application_period
- district
- max_persons


aprenfors (commented):
The implementation itself looks fine to me. Maybe things like Studiokompleksi, testi2 should be in english even though they are not really code, and purposes could have more descriptive names that foo and bar if those are actually used somewhere in future.

aprenfors (approved):

aprenfors:
Oh, looks like there are some merge conflicts in here.

api/reservation_units_api.py
```diff
@@ -1,9 +1,28 @@
...
+
+
+class ReservationUnitFilter(filters.FilterSet):
+    purpose = filters.ModelMultipleChoiceFilter(
+        field_name="purposes", queryset=Purpose.objects.all()
+    )
+    application_period = filters.ModelMultipleChoiceFilter(
+        field_name="application_periods", queryset=ApplicationPeriod.objects.all()
+    )
+    location = filters.ModelChoiceFilter(
```
tommimanbytes, line 20: Location model is created mostly for storing location data for various different spaces apps models (RealEstate, Building, Space).

Using this as relation filter is not that useful. You would have a dropdown full of addresses or something?

At the moment the location model is kind of work in progress. It should later contain also coordinate data. It could also be a base model that is inherited by models that uses it (RealEstate, Building, Space).

More useful for now would be to filter by districts. They are hierarchical structure of districts in Helsinki. Users could then filter by major district, district, sub district or neighborhood. Check here:  https://en.wikipedia.org/wiki/Subdivisions_of_Helsinki 


**TLDR:** Filter by location model is not useful or relevant. Maybe later you can filter by distance to your current location and this could be used when we add coordination data to app. **More relevant would be ability to filter by district (and maybe building and real estate)**

tommimanbytes (changes requested):

api/reservation_units_api.py
tommimanbytes: Thinking more about this maybe it would be somewhat useful if there would be like typeahead style of filter for location. Like you would start writing address, it offers an address with typeahead and clicking it would filter reservation units with spaces with that location (or reservation units with spaces whose building or real estate relation has this location)

However district filtering is most relevant for MVP I think. User path:
1. Look for application periods
2. Find application period that fits your needs
3. See filtered list of reservation units for that application period
4. Filter with district (of any hierarchy level) that your organisation operates in

otiala: Good point. I refactored location filter to a district filter, which takes hierarchy into account. District seems to be used as a filter option in designs.

tommimanbytes (approved):

PR #23, opened 2020-12-01
Tila 95 | application api refinement

aprenfors:


api/tests/test_application_api.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pytest
+from django.utils import timezone
+from rest_framework.reverse import reverse
+
+from applications.models import Application
+
+
+@pytest.mark.django_db
+def test_application_create(purpose, reservation_purpose, organisation, person, application_period, user_api_client):
```
aprenfors, line 11: Ok, here i have a bit of a problem. My model defines user as not null, so for some reason this fails if i don't make it not null. 

Probably not clear from this merge request where the problem is, but the relevant things are

https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/54b1df89899c49d877c87d0e137d9e12613254bc/api/applications_api.py#L75

The application serializer and viewset in there. If i set the user in the model to nullable then it works perfectly and the viewset sets the user from the request. 

Since i'm not too familiar with django rest framework, what's the proper way to handle this?

api/tests/test_application_api.py
```diff
@@ -0,0 +1,68 @@
...
+    purpose,
+    reservation_purpose,
+    organisation,
+    person,
+    application_period,
+    user_api_client,
+    user,
+):
+    assert Application.objects.count() == 0
+    data = {
```
tommimanbytes, line 21: So the idea is that in the application creation process we first create separate organization, contact person etc and only after that we make another call to API to create application with these? 

So its possible that we have organisations and contact persons in db without application relation for them?

aprenfors: I think organization probably should be separate table since it came up that it in the future might come from some external source etc. Contact person could probably be part of the application itself, doesn't look like it has much reusability, i can refactor that in separate merge request.

tommimanbytes (approved):

PR #27, opened 2020-12-04
TILA-105 | Initialize django-modeltranslations

otiala:
Translates the following models and fields:

- `ReservationUnit: name`
- `Purpose: name`
- `Space: name`
- `Resource: name`
- `Service: name`


tommimanbytes:
FYI in Respa there is a base serializer created that takes care of showing translations available in a object like this:
`
"name": {
    "fi": "foo",
    "en": "bar",
    "sv": "foobar"
}
`

I think we should use it or something similar here so we don't have to specify every translated field in serializer `fields` list.

otiala:
> FYI in Respa there is a base serializer created that takes care of showing translations available in a object like this:
> `"name": { "fi": "foo", "en": "bar", "sv": "foobar" }`
> 
> I think we should use it or something similar here so we don't have to specify every translated field in serializer `fields` list.
> 
> The base serializer in respa is in
> `resources/api/base.py`

I think this is only because how `django-parler` works, it does not serve all the available languages out of the API, but rather only the language requested, I think this is a bit unnecessary with `django-modeltranslation`.

tilanvarauspalvelu/settings.py
```diff
@@ -0,0 +1,144 @@
+import os
```
aprenfors, line 1: Hmm, looks like something is wrong here, this overrides stuff that should be in development, maybe something wrong with some merge from dev or something.

aprenfors (changes requested):

tilanvarauspalvelu/settings.py
tommimanbytes: I think this hasn't been rebased after changing name of project and project folder now appears as new becaus it has different name. Please rebase to current develop and push this branch again @otiala

aprenfors: Oh right, missed that this still has the old folder name.

tommimanbytes:
> FYI in Respa there is a base serializer created that takes care of showing translations available in a object like this:
> `"name": { "fi": "foo", "en": "bar", "sv": "foobar" }`
> 
> I think we should use it or something similar here so we don't have to specify every translated field in serializer `fields` list.
> 
> The base serializer in respa is in
> `resources/api/base.py`

Looks like you didn't use same style as in respa and translated fields now comes as their own fields from API. I think this can be pretty ugly sometimes if there is like 5 translated fields in a model. This creates 15 fields to root of the JSON object. Thats also a reason why I suggested we get translated fields as objects. Whats does @aprenfors and @indigane think about this?

aprenfors:
> Looks like you didn't use same style as in respa and translated fields now comes as their own fields from API. I think this can be pretty ugly sometimes if there is like 5 translated fields in a model. This creates 15 fields to root of the JSON object. Thats also a reason why I suggested we get translated fields as objects. Whats does @aprenfors and @indigane think about this?

So this currently outputs this as?

`{
    foo: "bar",
    name_fi: "nimi",
    name_en: "name",
    name_sv: "namn"
}`

i think what Tommi proposes so "name": { "fi": "foo", "en": "bar", "sv": "foobar" } looks cleaner to me and is semantically better when we clearly have one name field.

otiala:
Will chose this. I cherrypicked changes to a new branch.

PR #28, opened 2020-12-07
TILA-87 | Add data models and API changes to support reservation unit images

tommimanbytes:
We use UrlField for image url for now to prevent need for decision of media file storage for decembers demo.


api/reservation_units_api.py
```diff
@@ -25,10 +26,17 @@ class ReservationUnitFilter(filters.FilterSet):
     )
 
 
+class ReservationUnitSerializer(serializers.ModelSerializer):
```
aprenfors, line 29: Duplicate class name, might get a bit confusing, should probably be (reservationunit)imageserializer

tommimanbytes: Yes, accident. Will fix.

indigane (approved):

PR #30, opened 2020-12-11
TILA-87 | Made changes and additions to support newly designed reservation unit…

tommimanbytes:
… api changes. Also formatting changes with format.sh script.

Added location and max_persons fields for reservation units to support design for the demo.

Changed one-to-one relation fields of Location model in Spaces app for django admin inlines to be supported (they are supported only if fk field is in inline model).


reservation_units/models.py
```diff
@@ -35,6 +35,18 @@ class ReservationUnit(models.Model):
     def __str__(self):
         return "{}".format(self.name)
 
+    def get_location(self):
+        # For now we assume that if reservation has multiple spaces they all have same location
+        spaces = self.spaces.all()
+        if len(spaces) > 0 and hasattr(spaces[0], "location"):
```
aprenfors, line 41: Is it safe to assume if one (first)space has not location, then none of them have? Or would it be safer to do something like.

`return next((obj for obj in spaces if hasattr(obj, "location")), None)`

Would get rid of the hard coded index too which is in my opinion a bit ugly.

siren:
General suggestion: when implementing api changes, would it be possible to update fixtures/cases.json so that it demonstrates what kind of content is expected for the new fields/entities?

tommimanbytes:
> General suggestion: when implementing api changes, would it be possible to update fixtures/cases.json so that it demonstrates what kind of content is expected for the new fields/entities?
Good suggestion. Will update them. How practical this is as general rule depends on how developers use their local dbs. If people fill their db with all kinds of random test data, fixtures are also affected. This could be handled by having two dbs, one for fiddling around and one that has only somewhat clean data for fixtures.

Anyways. will update them for this pr.

aprenfors (approved):

PR #31, opened 2020-12-11
Tila 95 | application model refinements

aprenfors:


api/applications_api.py
```diff
@@ -57,39 +43,97 @@ def perform_create(self, serializer):
 
 
 class OrganisationSerializer(serializers.ModelSerializer):
+
+    id = serializers.IntegerField(allow_null=True, required=False)
+
+    def to_internal_value(self, data):
```
aprenfors, line 49: Here is some kind of proof of concept code, which may or may not be useful. When using writing nested serializers uniqueness can't be determined by rest framework automatically it seems, so we skip uniqueness check when id matches the one in database.

api/applications_api.py
```diff
@@ -57,39 +43,97 @@ def perform_create(self, serializer):
...
+        return person
+
+    def handle_organisation(self, organisation_data) -> Union[Organisation, None]:
+        organisation = None
+        if organisation_data is not None:
+            organisation = Organisation(**organisation_data)
+            organisation.save()
+        return organisation
+
+    def create(self, validated_data):
```
aprenfors, line 104: Here is another question, seems like handling updating and creating the nested objects is kind of annoying, is there some easier way to do this. Rather not create new row every time.

(Although i think at least probably contact person and maybe even organisation could be part of the application table itself, but anyway)

tommimanbytes: Maybe we could check in handle functions that if data has ID, then it is and existing object and should be updated rather than created. This kinda seems as clean as it gets to me with nested serializers.

api/applications_api.py
```diff
@@ -57,39 +45,90 @@ def perform_create(self, serializer):
 
 
 class OrganisationSerializer(serializers.ModelSerializer):
+
+    id = serializers.IntegerField(allow_null=True, required=False)
```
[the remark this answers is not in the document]
aprenfors: It seems to be when using nested write serializers, if it's not in there then we are creating a new organisation row every time we save an application. So how this currently works it updates the organisation row when application's organisation has an id, if not then creates a new one. Without the fields definition djago seems to strip the id field from the nested data, so the data is always without the id field unless specifically defined here. 

But maybe i'm missing something and there is simpler way to do it.

api/applications_api.py
aprenfors: Basically that how it works now. If it has an id it'll be updated, if not a new one is created. Maybe could use get_or_create or something to make it clearer?

aprenfors: No, actually get_or_create is bit of bad idea, but maybe some other way would be clearer.

tommimanbytes (approved):

PR #35, opened 2020-12-15
TILA-139 | Very simple basic or session authentication required for wri…

aprenfors:
…te operations.

Permissions can be fully turned off with an environment variable.


tommimanbytes (approved):

aprenfors:
Actually making a small change and not adding this to default env vars, because this is temporary thingy.

PR #45, opened 2021-01-04
Add tests for handling nested data update, save and delete.

aprenfors:
Some tests for existing code to verify nested data update/save/delete


otiala (approved):
Looks good to me!

PR #47, opened 2021-01-07
TILA-210 Add equipment to reservation unit. Reservation filtering by state

aprenfors:
This contains also kind of naive implementation of reservation fetches by reservation unit. The hierarchy isn't well defined atm, so currently this can just fetch reservations for the unit, does not take care of hierarchy. At least now this has tests for that so it might be easier to implement without breaking stuff. If it looks like it might not be beneficial to have it here can be removed. 

Terms of use are now just a field in reservation unit, even though the respa example had them in separate table. With current knowledge it doesn't really look like a separate table is needed (since payment stuff isn't really defined). Should be easy to refactor into separate table if needed in the future.


tommimanbytes (approved):
Looks good to me!

PR #48, opened 2021-01-11
TILA-81 | Add audit loggin to local database

aprenfors:
TILA-81 Enable logging to elastic apm, configure logger for it.

TILA-81 Document environment variables.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,20 +2,24 @@ aniso8601==7.0.0
 appdirs==1.4.4
 asgiref==3.3.0
 attrs==20.3.0
+auditlog3==1.0.1
 black==20.8b1
 certifi==2020.12.5
 click==7.1.2
 coverage==5.3
 Django==3.0.10
+django-auditlog==0.4.7
 django-cors-headers==3.6.0
 django-environ==0.4.5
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -39,7 +43,7 @@ pyparsing==2.4.7
 pytest==6.1.2
 pytest-cov==2.10.1
 pytest-django==4.1.0
-python-dateutil==2.8.1
+python-dateutil==2.6.0
```
nicobav, line 46: What caused this downward bump and should it be worth of mentioning here?

aprenfors: There was a version conflict for the dateutil dependency required by djago auditlog and what was used previously. This is how pip resolved the conflict and everything seems to work. I didn't find any real problems with just forcing this to 2.8.1 but it produces a nasty looking warning and would probably need more thorough testing.

nicobav: Alrighty :+1:  It would be nice to have a comment in the requirements that we're sticking to this version by a reason.

tommimanbytes (approved):

nicobav (approved):

PR #49, opened 2021-01-13
TILA-215 | Replace `ReservationUnitImage` field `image_url` with `image field

nicobav:
There is possible to store images to `ReservationUnit`.
There was initially made possibility to give the image url but no image file itself.
This commit replaces the `ReservationUnitImage` field `image_url`  with `image` field which should contain the actual image file.

Further development like validation store location etc. should be discussed in workshops.

TILA-215


tommimanbytes (approved):
Approving this because its definitely needed at some point. Although do we already have support for hosting image files in our in infra / configs? My concern is that we can't demo the UI with images as it progresses if we don't have support for serving media files in our infra yet. Has there been any discussions about where uploaded image files will be hosted?

nicobav:
> Approving this because its definitely needed at some point. Although do we already have support for hosting image files in our in infra / configs? My concern is that we can't demo the UI with images as it progresses if we don't have support for serving media files in our infra yet. Has there been any discussions about where uploaded image files will be hosted?

No discussions yet. The demoing is  a good point. I could leave the `image_url` there so that UI can use it in demoing until we get solved the storage location. What do you think? @tommimanbytes

aprenfors:
> Approving this because its definitely needed at some point. Although do we already have support for hosting image files in our in infra / configs? My concern is that we can't demo the UI with images as it progresses if we don't have support for serving media files in our infra yet. Has there been any discussions about where uploaded image files will be hosted?

Yeah, we should have configs for uploaded images/files, at least media folder etc are configured for prod type deployments.

nicobav:
I'll close this due to the fact that I will publish other PR for this with thumbnails and avoid the rebase hassle with this old thing.

PR #50, opened 2021-01-14
TILA-216 | Allow saving application in parts

tommimanbytes:
Allow null / blank values in several fields. Also added blank=True to many relation fields in addition to null=True, because you can't save empty value in Django Admin if blank=False.


nicobav (approved):
LGTM :ok_hand:

api/applications_api.py
```diff
@@ -140,22 +140,24 @@ class ApplicationEventSerializer(serializers.ModelSerializer):
     )
 
     application_id = serializers.PrimaryKeyRelatedField(
-        queryset=Application.objects.all(), source="application"
```
aprenfors, line 143: I think this application_id  should be required, i think it'll lead to pretty much these just disappearing when fetching an application with it's events. Creating these doesn't really make sense without the application existing.

tommimanbytes: > I think this application_id should be required, i think it'll lead to pretty much these just disappearing when fetching an application with it's events. Creating these doesn't really make sense without the application existing.

Hmm, yeah, I was thinking this only from nested write perspective where application ID was set automatically. Wanted to make it possible to create an application with events in one POST call and in that case there is no application ID yet. But as you said it will probably cause some problems when using ApplicationEvent endpoint independently. Also probably there isn't a use case where whole application isn't already saved before saving application events. I'll change that.

PR #52, opened 2021-01-21
Tila 280 | Add initial model for Unit

nicobav:
Adds model for Unit meaning as in premises/office(s)

Initial model for representing Unit.


nicobav:
Based on this is coming the importer.

spaces/models.py
```diff
@@ -36,6 +37,44 @@ def __str__(self):
...
+    short_description = models.CharField(
+        verbose_name=_("Short description"), max_length=255, blank=True, default=""
+    )
+    web_page = models.URLField(
+        verbose_name=_("Homepage for the unit"), max_length=255, default=""
+    )
+    email = models.EmailField(verbose_name=_("Email"), max_length=255)
+    phone = models.CharField(verbose_name=_("Telephone"), max_length=255)
+    address = models.ForeignKey(
+        "applications.Address",  # Avoid circulars. Should address be moved to spaces?
```
nicobav, line 66: I think `Address` could somewhat general purpose class that could be used there and there when addresses are involved. 
What do you think?

tommimanbytes: There is ´Location´ model in spaces app that includes address at the moment. It will also include coordinates (issue already in backlog). I think ´Unit´ should use the ´Location´  model as coordinates of Units could have some uses in the system (calculating distances etc).

tommimanbytes: Location is a model that is designed to be used as OneToOne relation to various models which has location data.

nicobav: Excellent! :ok_hand:

nicobav: So I added coordinates to `Location` class and referenced `Unit` in there also.

spaces/models.py
```diff
@@ -167,6 +200,18 @@ class Location(models.Model):
         related_name="location",
         on_delete=models.CASCADE,
     )
+    unit = models.OneToOneField(
```
aprenfors, line 203: Is this some django convention for referencing one to one relations. To me it would be more natural to define these one to one relations the other way around, like adding location to unit instead of adding unit to location. 

Location doesn't really need to know what model references it and the database isn't normalized when doing it this way?

tommimanbytes: I also think it would be more natural to define these in related models, not in Location. They are currently defined in the Location because for some strange reason Django Admin doesn't support one-to-one relations as inline forms without custom templates if relation is defined other way around.  So if you want to be able to fill location in same form as you are creating Unit for example then the relation has to be defined like its defined currently.

If someone has at some point time to investigate a convenient way to show inline location form in django admin and also defining location in the related models (space, unit etc) it would be nice imo too. But changing them now would break the ability to fill location inline in django admin and you would have to create location objects separately and then go to attach them to related models.

aprenfors (approved):

spaces/models.py
nicobav: I agree with you that the Location OneToOne should be defined in the model that needs it. Also having it this way simply touching non-existent location will blow-up. I could try to check this issue.

nicobav:
I think we'll merge this now and deal the refactor of the models later.

PR #54, opened 2021-01-22
TILA-279 | Document API endpoints as OpenAPI using DRF spectacular

tommimanbytes:
I also disabled some endpoints that clearly aren't ready to be open, lacks tests etc. I think it's best to open endpoints as we need and use them and prevent making endpoints for every model if we don't know for sure that they are needed separately. It keeps documentation clean and makes understanding the API easier.

I hope review from as many developers as possible, because understanding who the API works is crucial for whole team.

### DIFF
--- a/api/application_period_api.py
+++ b/api/application_period_api.py
@@ -16,6 +16,29 @@ class ApplicationPeriodSerializer(serializers.ModelSerializer):
             "reservation_period_end",
             "purposes",
         ]
+        extra_kwargs = {
+            "name": {
+                "help_text": "Name that describes this event.",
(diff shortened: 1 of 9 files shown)


tommimanbytes:
You can check the documentation in your local dev env by running this branch and going to one of these addresses:

OpenAPI doc file - http://localhost:8000/openapi/
Documentation UI in swagger format -  http://localhost:8000/swagger-ui/
Documentation UI in redoc format  -  http://localhost:8000/redoc/

aprenfors (approved):

PR #56, opened 2021-01-27
Add source code folder to dev env volumes in docker-compose 

nicobav:
This will reflect the changes made to code to the container without needing to rebuild.


nicobav:
okay, seems that the build fails for imports (isort) (out of this PR's scope). I'll provide a fixing commit to that as well in this PR.

aprenfors (approved):
Nice, I'm by no means expert in docker, so if you find anything else that seems off that's great.

PR #59, opened 2021-01-28
TILA-281 | Units importer

nicobav:
Adds units importer to code. 
Units are intended to be imported from https://www.hel.fi/palvelukarttaws/rest/v4/unit/.  The data source isn't restricted so this works for other sources as well.


aprenfors (approved):
Looks very clean to me.

tommimanbytes (approved):

PR #61, opened 2021-01-29
TILA-345 | Added model for service sector. Also linked reservation units to Units.

tommimanbytes:
Added initial model for ServiceSector (liikunta, nuoriso etc). This will be used for many purposes but I created if for now for making it possible for me to create user role / permission system POC implementation. 

I also linked reservation units to unit with foreign key. Unit admin/managers will be creating and editing reservation units for their unit and there must be a relation so we can determine which Units admins/managers can edit which reservation units.

Format.sh also made some changes.


spaces/models.py
```diff
@@ -37,6 +37,26 @@ def __str__(self):
...
 
+class ServiceSector(models.Model):
+    """
+    Model representation of Service Sector that contains and manages
+    units and application periods.
+    """
+
+    name = models.CharField(verbose_name=_("Name"), max_length=255)
+    units = models.ManyToManyField(
+        "Unit", verbose_name=_("Units"), related_name="departments"
```
aprenfors, line 48: related name seems to be wrong/inconsistent

tommimanbytes: Fixe'd

aprenfors (approved):

PR #63, opened 2021-01-29
Tila 299 | allocation object structure

aprenfors:
This creates objects that will be used in automatic allocation and implements mapping Applications to Allocation Data. 

There will be more stuff added in the future, this is the bare minimum to start allocating applications.

### DIFF
--- a/reservations/allocation_models.py
+++ b/reservations/allocation_models.py
@@ -1,0 +1,126 @@
+import datetime
+
+from django.utils import timezone
+
+from applications.models import Application, ApplicationEvent
+from reservation_units.models import ReservationUnit
+
+
+class AvailableTime(object):
+    def __init__(self, start: int, end: int):
+        self.start = start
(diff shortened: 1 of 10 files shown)


reservations/allocation_models.py
```diff
@@ -0,0 +1,126 @@
...
+        period_start: datetime.date,
+        period_end: datetime.date,
+        times: [AvailableTime],
+    ):
+        self.id = id
+        self._period_start = period_start
+        self._period_end = period_end
+        self.available_times: [AvailableTime] = times
+
+    def addTime(self, start: datetime, end: datetime):
```
nicobav, line 32: :camel: :eyes:

aprenfors: But camel case is superior... Changing python conventions one method at a time.

nicobav (approved):

PR #64, opened 2021-02-02
Register district to API and add translations

otiala:
Cherrypicked to another branch so I don't have to manually fiddle with migration conflicts.

- Register `District` to API, register `name` as a translatable field
- Confirm that all the necessary filters exist and work as intended


api/space_api.py
```diff
@@ -52,6 +52,22 @@ class Meta:
         }
 
 
+class DistrictSerializer(TranslatedModelSerializer):
+    class Meta:
+        model = District
+        fields = ["id", "district_type", "name", "parent"]
```
aprenfors, line 58: If parent is id field, should be named parent_id in the serializer according to contributing docs.

aprenfors: maybe same for district_type?

aprenfors (approved):

PR #66, opened 2021-02-03
TILA-285 | Application and ApplicationEvent statuses

nicobav:
New statuses for `Application` and `ApplicationEvent`


aprenfors:
Couple of failing tests.

aprenfors (approved):

api/applications_api.py
```diff
@@ -203,6 +203,8 @@ class ApplicationEventSerializer(serializers.ModelSerializer):
         help_text="List of reservation units applied for this event with priority included.",
     )
 
+    status = serializers.CharField(help_text="Status of this application")
```
siren, line 206: Should this be 'Status of applicationEvent'? 

It would also be nice from API users perspective to have some additional documentation such as the valid values and when to use which. Currently it is documented (in swagger) as:

"status* | string Status of this application"

Additionally I think the error message could also be improved. For example now I have a broken UI at hand and all the api tells me is '["Invalid status"] 

Does this refer to status of the application or applicationEvent?

applications/migrations/0011_add_new_application_statuses.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('applications', '0010_allow_null_for_application_fields'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='applicationstatus',
+            name='status',
+            field=models.CharField(choices=[('draft', 'Draft'), ('in_review', 'In review'), ('review done', 'Review done'), ('allocating', 'Allocating'), ('allocated', 'Allocated'), ('validated', 'Validated'), ('approved', 'Approved'), ('declined', 'Declined'), ('cancelled', 'Cancelled')], max_length=20, verbose_name='Status'),
```
siren, line 16: review done -> review_done ?

api/applications_api.py
nicobav: Good spots :+1:  
Yes, this considers the ApplicationEvent. 
I'll make PR for fixing these issues

PR #68, opened 2021-02-03
TILA-360 Some refactoring into incoming data to allocation models.

aprenfors:
TILA-327 Restrict allocation by units selected in the applicationa and those available in application period.

Refactored allocation data to take application period as a parameter instead of an application. 

Restrict allocation to units that are selected for each application (and those available in application period)


reservations/allocation_solver.py
```diff
@@ -72,38 +86,44 @@ def print_solution(self):
...
     def solve(self):
         model = cp_model.CpModel()
 
         selected = {}
-        for space in self.spaces:
-            for allocation_event in self.allocation_events:
-                for occurence_id, occurence in allocation_event.occurrences.items():
+        for allocation_event in self.allocation_events:
+            for occurence_id, occurence in allocation_event.occurrences.items():
+                for space_id, space in suitable_spaces_for_event(
```
aprenfors, line 98: This space restriction could be also done with constraints on the model, while this might be a bit less readable, i think it might be smart to do it this way instead so we limit the amount of data we pass into the solver as early as possible to improve performance. This would be a hard constraint in the model anyway, so doing it this way won't affect the solutions.

nicobav (approved):

PR #69, opened 2021-02-05
Tila 361 | restrict allocation by capacity

aprenfors:


reservations/tests/test_allocation_solver.py
```diff
@@ -49,3 +49,94 @@ def test_non_matching_unit_in_application_and_application_period_can_not_be_allo
     solution = solver.solve()
 
     assert len(solution) == 0
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "multiple_applications",
```
nicobav, line 56: this is a beast of a parameter :sweat_smile:

nicobav (approved):

reservations/tests/test_allocation_solver.py
aprenfors: Yeah, it's a bit much. Maybe should create objects instead of doing it like this, would probably be more readable and easy to construct.

PR #73, opened 2021-02-09
TILA-285 | status enhancements

nicobav:
Replaces `ApplicationStatus` "approved" to "handled" and adds more descriptive help texts for statuses.

Fixes typo in `ApplicationStatus` "review done" to "review_done"


aprenfors (approved):

api/applications_api.py
```diff
@@ -205,7 +205,7 @@ class ApplicationEventSerializer(serializers.ModelSerializer):
         help_text="List of reservation units applied for this event with priority included.",
     )
 
-    status = serializers.CharField(help_text="Status of this application")
+    status = serializers.CharField(help_text="Status of this application event")
```
siren, line 208: Is it possible to also document the values with this framework?

In openapi these are called enums (https://swagger.io/docs/specification/data-models/enums/).

aprenfors: I'd imagine if you use ChoiceField (or MultipleChoiceField, but i guess this is just single choice)  instead of CharField those get documented? Which imo would be correct choice anyway.

nicobav: very good, I'll change that

aprenfors (approved):

nicobav:
woosh, got sweating 'cause all of the conflicts :sweat_smile:

api/applications_api.py
siren: Looking good now 👍

PR #75, opened 2021-02-10
TILA-388 | Fixed fixtures

tommimanbytes:
Minor bug in application round status string representation and updated data structure image


nicobav (approved):
Looks good :+1:
Special points for updating the models image :cookie:

PR #77, opened 2021-02-10
changes to the Dockerfile

jaana-ibm:
Changes made:
 - added subscription-manager in order to enable the codeready-builder-for-rhel-8-x86_64-rpms repository
 - added chgrp and chmod command for /tvp folder


aprenfors:
Seems to introduce problems when running with docker-compose which passes password and username as build args, i can take a look into that.

aprenfors (approved):

PR #80, opened 2021-02-15
TILA-396 | organization active members

nicobav:
Adds active members to organisation and to API (nested in application)


kieferbonk (approved):

api/applications_api.py
```diff
@@ -67,6 +73,9 @@ class Meta:
             "year_established": {
                 "help_text": "Year when the organisation was established.",
             },
+            "active_members": {
+                "help_text": "Active space user person count of the organization.",
```
aprenfors, line 77: Active space user person count?

nicobav: :laughing:  yes
I tried to point out perhaps the obvious and surely sounds like a "rabbit hole". 
Perhaps just  "Number of active persons in the organization"?

aprenfors: Yeah, that sounds good :)

nicobav: Fix'd

aprenfors (approved):

PR #81, opened 2021-02-15
TILA-384 Getting application events by basket.

aprenfors:
TILA-384 Filter by purpose age group and order number.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -239,6 +239,21 @@ class ApplicationRound(models.Model):
     def get_status(self):
         return self.statuses.last()
 
+    def get_application_events_by_basket(self):
+        handled_application_events: [ApplicationEvent] = {}
+        for basket in sorted(
+            self.application_round_baskets.all(), key=lambda k: k.order_number
+        ):
+            handled = [
+                event
+                for events in handled_application_events.values()
+                for event in events
+            ]
+            handled_application_events[
+                basket.id
+            ] = basket.get_application_events_in_basket(handled)
+        return handled_application_events
+
     def __str__(self):
         return "{} ({} - {})".format(
             self.name, self.reservation_period_begin, self.reservation_period_end
@@ -293,6 +308,18 @@ class ApplicationRoundBasket(models.Model):
     order_number = models.PositiveSmallIntegerField(
         verbose_name=_("Order number"), default=1, null=False, blank=True
     )
+
+    def get_application_events_in_basket(self, applicationevents):
+        events = ApplicationEvent.objects.filter()
+
+        if self.purpose is not None:
+            events = events.filter(purpose=self.purpose)
+        if len(self.age_groups.all()) > 0:
+            events = events.filter(age_group__in=self.age_groups.all())
+        if len(applicationevents) > 0:
+            events = events.exclude(id__in=map(lambda x: x.id, applicationevents))
+
+        return list(events.all())
 
     def __str__(self):
         return "{} ({})".format(self.name, self.application_round.name)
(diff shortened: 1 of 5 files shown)


applications/models.py
```diff
@@ -239,6 +239,21 @@ def set_status(self, status, user=None):
     def get_status(self):
         return self.statuses.last()
 
+    def get_application_events_by_basket(self):
+        handled_application_events: [ApplicationEvent] = {}
+        for basket in sorted(
+            self.application_round_baskets.all(), key=lambda k: k.order_number
```
nicobav, line 245: This is merely a taste issue; `self.application_round_baskets.order_by("order_number")` would be the django way 

obviously performance varies between ordering in memory and by the db but I think this case we have quite low data mass for baskets per application rounds so it does not matter.

applications/models.py
```diff
@@ -239,6 +239,21 @@ def set_status(self, status, user=None):
...
+        ):
+            handled = [
+                event
+                for events in handled_application_events.values()
```
nicobav, line 249: Is the return value of `.values()` needed here (it is queryset dict)? or would just `.all()` be good here (which is queryset with objects)?

applications/models.py
```diff
@@ -294,6 +309,18 @@ class ApplicationRoundBasket(models.Model):
         verbose_name=_("Order number"), default=1, null=False, blank=True
     )
 
+    def get_application_events_in_basket(self, applicationevents):
```
nicobav, line 312: so the parameter of this excludes the given ApplicationEvents from the result?

applications/models.py
```diff
@@ -294,6 +309,18 @@ class ApplicationRoundBasket(models.Model):
...
+        events = ApplicationEvent.objects.filter()
```
nicobav, line 313: this includes all the ApplicationEvents. Should it be just the ones in the basket?

aprenfors: We don't know which events are in the basket since it's purely based on the filtering by purpose, age group etc.

applications/models.py
aprenfors: Yeah, that looks better.

applications/models.py
aprenfors: I'll check it out.

applications/models.py
aprenfors: Yeah, through i renamed this better already but i'll give it more descriptive name.

applications/models.py
aprenfors: But yeah, actually there is a bug here! Should restrict by application round.

applications/models.py
aprenfors: With the current way it's done we need to use values, not a queryset dict as such, it's a dict with basket id as key and instances of events that belong to that basket as values. Maybe there is a better way to structure this?

applications/models.py
```diff
@@ -294,6 +307,20 @@ class ApplicationRoundBasket(models.Model):
...
+        events = ApplicationEvent.objects.filter(
+            application__application_round=self.application_round
+        )
+
+        if self.purpose is not None:
+            events = events.filter(purpose=self.purpose)
+        if len(self.age_groups.all()) > 0:
+            events = events.filter(age_group__in=self.age_groups.all())
+        if len(excluded_events) > 0:
+            events = events.exclude(id__in=map(lambda x: x.id, excluded_events))
```
nicobav, line 320: Not a change request; this could also be written as: `events = events.exclude(id__in=excluded_events.values_list("id", flat=True))` if the excluded_events is a queryset.

nicobav (approved):

PR #82, opened 2021-02-15
Fixes to application API

otiala:
This PR is waiting for frontend implementation.


api/applications_api.py
```diff
@@ -460,7 +509,7 @@ def handle_events(self, appliction_instance, event_data):
         event_ids = []
         for event in event_data:
             event["application"] = appliction_instance
-            if "id" not in event or ["id"] is None:
+            if "id" not in event or event["id"] is None:
```
aprenfors, line 512: Nice catch 👍

aprenfors (approved):

applications/migrations/0018_organisation_core_business.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('applications', '0017_application_billing_address'),
+        ('applications', '0018_organisation_active_members'),
```
nicobav, line 9: I have bad experience with same migration numbers.

PR #89, opened 2021-02-18
Register coordinates field in LocationSerializer

otiala:
- Register coordinates field in LocationSerializer
- Declare SRID
- Serve coordinates in nested object to be more explicit about lon/lat


tommimanbytes (approved):
Is the purpose of this issue (TILA-391) to create only read-only support for coordinates? If so, then looks fine by me.

PR #93, opened 2021-02-19
Tila 384 allocation by basket

aprenfors:
There are no functional changes here, just refactored django models out of allocation models for less tight coupling and created an intermediate builder to handle initializing AllocationData

Also refactored test cases to use assertpy.

### DIFF
--- a/reservations/allocation_data_builder.py
+++ b/reservations/allocation_data_builder.py
@@ -1,0 +1,134 @@
+import datetime
+from typing import Dict, List
+
+from django.utils import timezone
+
+from applications.models import ApplicationRound
+from reservation_units.models import ReservationUnit
+from reservations.allocation_models import (
+    AllocationBasket,
+    AllocationData,
+    AllocationEvent,
+    AllocationSpace,
+)
+
+
+class AllocationDataBuilder(object):
+    def __init__(
+        self, application_round: ApplicationRound, included_baskets: [int] = []
+    ):
+        self.period_start: datetime.date = application_round.reservation_period_begin
+        self.period_end: datetime.date = application_round.reservation_period_end
+        self.application_round = application_round
+        self.included_baskets = included_baskets
+        self.baskets = {}
+
+    def get_allocation_data(self):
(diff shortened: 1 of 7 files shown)


reservations/allocation_data_builder.py
```diff
@@ -0,0 +1,134 @@
...
+                        occurrences=application_event.get_all_occurrences(),
+                        period_start=self.period_start,
+                        period_end=self.period_end,
+                        baskets=event_baskets.get(application_event.id),
+                        space_ids=space_ids,
+                        begin=application_event.begin,
+                        end=application_event.end,
+                        min_duration=application_event.min_duration,
+                        max_duration=application_event.max_duration
+                        if application_event.max_duration is not None
```
nicobav, line 54: do we need explicit `None` check here? 
For example using  `if application_event.max_duration` to also check against zero (0) value => so what if the max_duration is zero, should it be count or not as max_duration?

reservations/tests/test_allocation_model.py
```diff
@@ -115,11 +123,28 @@ def test_mapping_application_round_baskets(
...
+        application_round=application_round_with_reservation_units
+    ).get_allocation_data()
+
+    assert_that(data.baskets).contains_key(
+        application_round_basket_one.id, application_round_basket_two.id
+    )
+
+    assert_that(data.baskets[application_round_basket_one.id]).has_id(
+        application_round_basket_one.id
+    ).has_order_number(
```
nicobav, line 136: this dynamic assertion is nice :sunglasses:

nicobav (approved):

reservations/allocation_data_builder.py
aprenfors: There is validation that checks that max duration is none or larger than min duration, so there should be no need to check for zeros.

reservations/tests/test_allocation_model.py
aprenfors: Yeah, i kind of find magic like this a bit disturbing, could be hard to debug, but seems to work.

PR #96, opened 2021-02-24
TILA-409 | Application api parameters

nicobav:
Adds parameters for `ApplicationStatus` (status) and `ApplicationRound` (application_round) to application api

Adds new field to `Application` called "cached_latest_application_status" where the latest application status value is stored. This makes queries lot easier and faster.

PR #97, opened 2021-02-25
TILA-426 | ApplicationEvent api data enrichment

nicobav:
Adds purpose name, ability group name and reservation unit details into application event data returned by api

### DIFF
--- a/api/applications_api/serializers.py
+++ b/api/applications_api/serializers.py
@@ -233,7 +233,12 @@ class ApplicationEventSerializer(serializers.ModelSerializer):
         queryset=Purpose.objects.all(),
         source="purpose",
         allow_null=True,
-        help_text="Id of the usa purpose for this event.",
+        help_text="Id of the use purpose for this event.",
+    )
+
+    purpose = serializers.SerializerMethodField(
+        method_name="get_purpose_name",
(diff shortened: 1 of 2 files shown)


nicobav:
Depends on changes in https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/96

api/applications_api/serializers.py
```diff
@@ -298,6 +315,14 @@ class Meta:
             },
         }
 
+    def get_purpose_name(self, obj):
```
aprenfors, line 318: Not 100% related to this PR, but purpose name doesn't seem to be unique, which it most likely should be. While this is read only thingy, i think when used like this should be maybe used only on unique values to be super clear. 

But no need to do anything in this pr, more or less for discussion.

aprenfors (approved):

api/applications_api/serializers.py
nicobav: Yes, I think the purposes should be perhaps be pre defined => it could be just enum also

aprenfors: Not sure if it can be an enum, might need some flexibility. Anyway this looks fine to me.

PR #100, opened 2021-02-26
Tila 384 basket allocation priority

aprenfors:


reservations/allocation_solver.py
```diff
@@ -202,38 +245,64 @@ def constraint_by_capacity(self, model: cp_model.CpModel, selected: Dict):
...
-            for space_id, space in suitable_spaces_for_event(
-                event, self.spaces
-            ).items():
-                model.Add(
-                    sum(
-                        selected[(space_id, event.id, occurrence_id)]
-                        for occurrence_id, occurrence in event.occurrences.items()
-                    )
-                    <= event.events_per_week
+        all_events = {}
```
aprenfors, line 248: This needs to be refactored so we don't need this logic in the solver itself. 

Since this needs some refactoring into allocation models, i'll do it in separate PR, so the logic changes and model changes are separate and easier to read.

nicobav (approved):

PR #103, opened 2021-03-02
Add description field for ReservationUnit

otiala:


aprenfors (approved):

reservation_units/models.py
```diff
@@ -40,7 +40,9 @@ def __str__(self):
 
 class ReservationUnit(models.Model):
     name = models.CharField(verbose_name=_("Name"), max_length=255)
-
+    description = models.TextField(
+        verbose_name=_("Description"), max_length=512, blank=True, default=""
```
nicobav, line 44: is the `max_length` something that is wanted explicitly? Asking for curiosity since the `max_length` of the `TextField` only is used for auto generated form field but not used in [database level](https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.TextField)

otiala: It's not explicitly needed, but I added it for consistency. Also I'm pretty sure DRF uses it in serializer-level validation.

PR #104, opened 2021-03-02
Add pytest factory boy and few tests

nicobav:
Adds few tests against  `ReservationUnit.check_reservation_overlap` with factories.


reservation_units/tests/factories.py
```diff
@@ -0,0 +1,82 @@
+from factory import SubFactory, post_generation
```
aprenfors, line 1: 👍 These factories looks really nice.

aprenfors (approved):

spaces/tests/test_unit_importer.py
```diff
@@ -92,7 +92,7 @@ def test_importer_creates_a_units_with_correct_values(self, mock_response):
 
         unit_too = Unit.objects.get(service_map_id=2)
         self.check_unit_values_vs_data_values(unit_too, data[1])
-        location_too = Location.objects.get(unit=2)
```
nicobav, line 95: have no idea why this passed the tests before and why this failed now

nicobav: now I know. The `Location.objects.get(unit=[integer])` where the integer refers to pk of the fk reference. TIL.

PR #105, opened 2021-03-04
TILA-284 | ReservationUnit images as files

nicobav:


tilavarauspalvelu/urls.py
```diff
@@ -17,3 +19,6 @@
     path("redoc/", SpectacularRedocView.as_view(url_name="schema")),
     path("graphql", GraphQLView.as_view(graphiql=True)),
 ]
+
+if settings.DEBUG:
+    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
nicobav, line 24: :see_no_evil:

tommimanbytes (approved):
Looks nice and clean solution to me.

siren:
I am adding my comments also here in case they were missed.

Should we add another image size in addition to the suggested thumb nail size. if you look at the reservation unit page there are those thumbnail images and then images that are  of with 384 px. If we use the smalle rones we would need to stretch the images and if we use the original images then the load time will be huge (the images in used varaamo are several megs in size)

nicobav:
> I am adding my comments also here in case they were missed.
> 
> Should we add another image size in addition to the suggested thumbnail size. if you look at the reservation unit page there are those thumbnail images and then images that are of with 384 px. If we use the smaller ones we would need to stretch the images and if we use the original images then the load time will be huge (the images used varaamo are several megs in size)

Yes, good :+1:  I've added the 384px image as "medium" now.

aprenfors (approved):

PR #108, opened 2021-03-08
TILA-346 | Initial implementation of roles and API permissions

tommimanbytes:
Finally somewhat ready implementation of roles and permissions for tilavarauspalvelu. Includes the permission system, implementation of permissions for each API endpoint, loads of test and contributing.md info for permissions. Documentation will be done separately to JIRA.


api/reservations_api.py
```diff
@@ -122,6 +135,19 @@ class ReservationViewSet(viewsets.ModelViewSet):
     def perform_create(self, serializer):
         serializer.save(user_id=self.request.user.pk)
 
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        user = self.request.user
+        return queryset.filter(
```
aprenfors, line 141: Hmm, do we need to restrict visibility of reservations. Those are needed for showing availability of reservation unit. Although i guess this api/viewset shows information that shouldn't be visible there and probably needs a separate api endpoint.

api/tests/test_application_api.py
```diff
@@ -323,3 +298,186 @@ def test_application_review_invalid(
     }
     response = user_api_client.post(reverse("application-list"), data, format="json")
     assert response.status_code == 400
+
+
+@pytest.mark.django_db
+def test_application_api_permissions(
```
aprenfors, line 304: i'd maybe split these into multiple tests instead of asserting so many things in one test, but that can be refactored later.

api/tests/test_application_round_api.py
```diff
@@ -79,17 +86,134 @@ def test_application_round_when_basket_orders_dont_overlap(
...
+    service_sector_application_manager_api_client,
+    unit_admin_api_client,
+):
+    # Normal user can't create application rounds
+    response = user_api_client.post(
+        reverse("application_round-list"),
+        data=valid_application_round_data,
+        format="json",
+    )
+    assert response.status_code == 403
```
aprenfors, line 135: A bit long and multi assert test here too.

aprenfors (commented):
Read about half of this, will continue after eating. Generally looks good, i find tests a bit hard to read since they test so many things in same test, would probably split those into smaller tests, also easier to debug/see what's wrong if they start failing when you can see if one assert or all of them fail etc. 

But this can be refactored in separate pr.

api/reservations_api.py
tommimanbytes: There is a task in JIRA about availablility of reservation unit from API and I was thinking when creating it that maybe we could retrieve availability in some different way than as list of reservations. In Respa we used reservation objects but with most of the information taken out of it for privacy reasons. 

However I thought at some point that it wasn't the best way to handle possibly hundreds of reservation objects and their possible buffer periods or some other things that block reserving for certain period. But I'm not sure if there is better way and you could very well be right about this.

However some information in reservations could be very private and shouldn't be showed to everyone. We don't have an implementation to filter that kind of information from public yet so because of that I made reservation visible for users themselves and admins etc only. 

I can of course change this for now because looks like there is not much private information in reservations yet.

api/tests/test_application_api.py
tommimanbytes: Yeah, I thought about that as valid option but then again I thought it wouldn't be nice either if in same file as other tests there would be 20 permissions tests at the end. We could of course make new test files for each endpoints permissions tests or something. 

I agree that its not beautiful as is either.

permissions/helpers.py
```diff
@@ -0,0 +1,273 @@
...
+        map(
+            lambda role: role.service_sector,
+            user.service_sector_roles.filter(role__permissions__permission=permission),
+        )
+    )
+
+
+def can_view_reservation(user: User, reservation: Reservation) -> bool:
+    permission = "can_view_reservations"
+    reservation_units = reservation.reservation_unit.all()
```
aprenfors, line 196: Mostly duplicate code here and the next method, might be ok so the permissions are super independent.

api/tests/test_application_api.py
nicobav: I agree with splitting to smaller (multiple) tests. Easier to read / make changes if needed etc.

nicobav (approved):
Quite a chunk :hot_face: Looks good to my evening eyes :+1: 
Those tests could be refactored but I'll give my approval with or without.

api/reservations_api.py
aprenfors: Probably leave as is, since we want to get the availability information from some place else anyway eventually, this kind of forces/reminds to implement that instead of relying on this.

permissions/helpers.py
```diff
@@ -0,0 +1,273 @@
...
+        role__permissions__permission=required_permission
+    ).exists()
+
+
+def can_manage_general_roles(user: User) -> bool:
+    permission = "can_manage_general_roles"
+    return has_general_permission(user, permission) or is_superuser(user)
+
+
+def can_manage_service_sector_roles(user: User, service_sector: ServiceSector) -> bool:
```
aprenfors, line 61: The might be some premature optimization in here with ordering, i'd guess is_superuser is the lightest of these to check, but as said more or less premature.

tommimanbytes: I swapped lightest checks to beginning

api/tests/test_application_round_api.py
tommimanbytes: Changed to separate tests

api/tests/test_application_api.py
tommimanbytes: Changed to separate tests

aprenfors (approved):

api/reservations_api.py
tommimanbytes: Leaving as is

PR #116, opened 2021-03-12
TILA-358 | Helusers integration, new user model and API for getting users

tommimanbytes:
This PR adds Tunnistamo authentication to the project.

Exact list of changes / new stuff:
- Added django-helusers library and its dependencies
- Added new user model inherited from helusers one
- Enabled JWT authentication. Frontend now logins in Tunnistamo, gets JWT token and uses it to authenticate to backend. Basic and session authentication are enabled in debug mode.
- Enabled Django Admin authentication with Tunnistamo. Known issue: For same SSO user two local user objects are created if same user uses Django Admin and API with JWT token.
- Added API endpoint for users. Admins can see every user, normal users only own. You can also get only current authenticated user object from `v1/users/current/`. This can be useful for frontend to check which roles user has etc. Roles are included with user objects.
- Updated readme.md authentication section

**BECAUSE USER MODEL CHANGES WITH THIS PR, ALL DATABASES NEEDS TO BE CREATED AGAIN. MIGRATIONS TO EXISTING DATABASES WON'T WORK DIRECTLY**


api/tests/test_users_api.py
```diff
@@ -0,0 +1,60 @@
...
+):
+    response = general_admin_api_client.get(
+        reverse("user-list"),
+        format="json",
+    )
+
+    assert response.status_code == 200
+    user_id_list = list(map(lambda x: x["id"], response.data))
+
+    assert user.id in user_id_list
```
aprenfors, line 37: Could use assertpy assertions for more fluent assertions, like 

assert_that(user_id_list).contains_only(user.id, user_2.id, general_admin.id)

But we can  refactor later if you're running out of hours.

aprenfors:
some lint failure with build.

api/tests/test_users_api.py
tommimanbytes: Changed in a few places

aprenfors (approved):

PR #120, opened 2021-03-18
Tila 452 allocation start api

aprenfors:


api/allocation_api.py
```diff
@@ -0,0 +1,123 @@
...
+
+class AllocationInProcessException(Exception):
+    pass
+
+
+class UnsuitableStatusForAllocationException(Exception):
+    pass
+
+
+class AllocationRequestSerializer(serializers.ModelSerializer):
```
aprenfors, line 28: So for @kieferbonk 

This is a way to start allocating specifiv allocation round (and selected baskets)

Basically at minimum you only need application_round_id, 

This can also track whether it has completed and when, maybe we should actually rename completed to something else, like succeeded? Basically currently if there is no end date it has not completed allocating, and the boolean completed means if false some exception happened (they shouldn't) if true allocation completed successfully. 

When starting the allocation it also marks application round's allocating to true and sets it back to false when completed (or failed)

api/tests/test_allocation_api.py
```diff
@@ -0,0 +1,77 @@
+import pytest
+from assertpy import assert_that
+from rest_framework.reverse import reverse
+
+from applications.models import ApplicationRoundStatus
+
+
+@pytest.mark.parametrize(
+    "round_status",
+    [ApplicationRoundStatus.REVIEW_DONE, ApplicationRoundStatus.ALLOCATED],
```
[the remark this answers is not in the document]
kieferbonk: This is how I'm seeing it as well.

allocation/models.py
```diff
@@ -0,0 +1,36 @@
...
+
+class AllocationRequest(models.Model):
+
+    start_date = models.DateTimeField(
+        verbose_name=_("Start time"), null=False, blank=True
+    )
+
+    end_date = models.DateTimeField(verbose_name=_("End time"), null=True, blank=True)
+
+    completed = models.BooleanField(verbose_name=_("Completed"), null=True, blank=True)
```
nicobav, line 18: should we add `default=False` and make this non nullable? Would make it easier to query for not completed ones.

api/allocation_api.py
```diff
@@ -0,0 +1,123 @@
...
+        t = threading.Thread(
+            target=self.start_allocation, args=[allocation_request], daemon=True
+        )
+        t.start()
+        return allocation_request
+
+    def update(self, instance, validated_data):
+        return AllocationRequest.objects.get(pk=instance.id)
+
+    def start_allocation(self, allocation_request: AllocationRequest):
```
nicobav, line 87: Should we put logic in this method under some facade method to allocation code api?

nicobav (approved):
Looks good, some minor comments not necessary change requests. :thumbsup

allocation/models.py
aprenfors: Yeah, that sounds better.

api/allocation_api.py
aprenfors: Yeah, i think that'd probably be better place and not pollute serializer with this.

PR #123, opened 2021-03-22
TILA-112 | Translations

nicobav:


aprenfors:
Looks like there are test failures, but other than that looking good.

nicobav:
> Looks like there are test failures, but other than that looking good.

Yup, I'm on it

jmp:
I'll close this very old translation-related PR since it should be obsolete. Let's re-open if needed. 🙂

PR #125, opened 2021-03-26
TILA-507 | Restrict applicant to save the application after period ends

nicobav:
Adds freezegun dependency for testing time / date specific cases (like this).


permissions/helpers.py
```diff
@@ -166,7 +169,11 @@ def can_manage_service_sectors_applications(
 def can_modify_application(user: User, application: Application) -> bool:
     return (
         is_superuser(user)
-        or application.user == user
+        or (
+            application.user == user
+            and datetime.now(tz=UTC)
```
aprenfors, line 174: Would using 

`timezone.get_default_timezone()`

be safer since it'd always get the timezone from settings (although it's UTC atm, but i guess could theoretically change)

nicobav: good point - I'll do the change :+1:

aprenfors (approved):

PR #126, opened 2021-03-26
TILA-327 Internal api for fetching opening hours for a reservation unit.

aprenfors:


requirements.txt
```diff
@@ -63,15 +66,18 @@ py==1.9.0
...
+pyjwkest==1.4.2
 PyJWT==2.0.1
 pyparsing==2.4.7
 pyrsistent==0.17.3
 pytest==6.1.2
 pytest-cov==2.10.1
 pytest-django==4.1.0
-pytest-factoryboy===2.1.0
+pytest-factoryboy==2.1.0
+pytest-mock==3.5.1
```
nicobav, line 80: I think we could use the unittest.mock(?)

api/tests/test_opening_hour_api.py
```diff
@@ -0,0 +1,41 @@
...
+from rest_framework.reverse import reverse
+
+from opening_hours.hours import TimeElement
+
+
+@pytest.mark.django_db
+def test_fetching_opening_hours_for_reservation_unit(
+    user_api_client, reservation_unit, mocker
+):
+    mocker.patch(
```
nicobav, line 14: you could use the unittest mock and add a decorator; `@mock.patch("api.hauki_api.get_opening_hourse", return_value=something)`

allocation/allocation_data_builder.py
```diff
@@ -80,6 +80,7 @@ def get_all_dates(self):
         return dates
 
     def get_space(self, unit: ReservationUnit):
+
```
nicobav, line 83: this air could be removed

requirements.txt
aprenfors: Yeah, this is actually probably unnecessary.

nicobav (approved):
:ship:

PR #127, opened 2021-04-06
TILA-439| Allocation result api

nicobav:
and allocation solution mapping to db objects.


applications/allocation_result_mapper.py
```diff
@@ -0,0 +1,48 @@
...
+                    },
+                )
+            except Exception:
+                logger.exception(
+                    "AllocationResultMapper: error occurred while creating event results."
+                )
+
+            try:
+                application_event_schedule = ApplicationEventSchedule.objects.get(
+                    pk=allocated_event.occurrence_id  # Correct?
```
aprenfors, line 33: Yeah, this should be correct. Need to do something about the naming conventions to make them more clear, like we use occurence and space in allocation models, and something different elsewhere, so it's a bit confusing.

aprenfors (approved):

applications/allocation_result_mapper.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def to_events(self):
+        for allocated_event in self.allocated_events:
+            try:
+                application_event_schedule = ApplicationEventSchedule.objects.get(
+                    pk=allocated_event.occurrence_id
+                )
+                ApplicationEventScheduleResult.objects.update_or_create(
+                    application_event_schedule=application_event_schedule,
+                    allocated_reservation_unit_id=allocated_event.space_id,
```
aprenfors, line 21: Should this reservation unit be in defaults? It can change from one unit to another.

nicobav: good point :+1:  Actually my initial thought was it would create a new one (result) but the logic does not handle deletion so I'll add it to defaults.

aprenfors (approved):

PR #129, opened 2021-04-07
Application round service sector can be null

aprenfors:


aprenfors:
@kieferbonk This may not fix the issue you're having, and if it doesn't let's check your role permissions.

nicobav (approved):

PR #133, opened 2021-04-08
Add applicationschedule to admin as event's inline

nicobav:


aprenfors (approved):
Some import order failure, but other than that looks good.

PR #135, opened 2021-04-09
Create cities and update to reference city table.

aprenfors:


aprenfors:
There is probably way to do this without pure sql, but it seemed a bit complicated. But if you have an idea how to do it easily, please suggest.

nicobav (approved):

PR #147, opened 2021-04-20
Tila 534 restricted spaces

aprenfors:


nicobav:
Should we have the "don't take this reservation unit into account" in the allocation logic also? Or does that come in separate PR?

aprenfors:
> Should we have the "don't take this reservation unit into account" in the allocation logic also? Or does that come in separate PR?

Yeah, i'll do allocation part in separate pr.

nicobav (approved):

PR #149, opened 2021-04-21
Tila 546 | Application event status permission

nicobav:
Also adds user as read_only in application status api
 - sets the user from request

### DIFF
--- a/api/applications_api/serializers.py
+++ b/api/applications_api/serializers.py
@@ -21,7 +21,6 @@ from applications.models import (
     Organisation,
     Person,
     Recurrence,
-    User,
 )
 from reservation_units.models import Purpose, ReservationUnit
 from reservations.models import AbilityGroup, AgeGroup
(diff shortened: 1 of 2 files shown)


api/applications_api/serializers.py
```diff
@@ -667,6 +665,10 @@ class Meta:
         read_only_fields = ["id", "timestamp"]
 
     def create(self, validated_data):
+        request = self.context.get("request")
         instance = ApplicationEventStatus(**validated_data)
+        if request:
+            instance.user = request.user
+        instance.user = self.request.user
```
aprenfors, line 672: Isn't this already set in the previous line?

nicobav: good spot :+1:  that's poor cleanup by me.

nicobav: should be fixed now

aprenfors (approved):

PR #150, opened 2021-04-21
Fix conflicting migrations

nicobav:


aprenfors (approved):

aprenfors:
Maybe we should add a step to merge main into pr when doing ci build so it would catch these?

nicobav:
> Maybe we should add a step to merge main into pr when doing ci build so it would catch these?

that would be nice. Not sure how to detect those until running migrations after other PR is merged which caused the conflict :thinking:

aprenfors:
> that would be nice. Not sure how to detect those until running migrations after other PR is merged which caused the conflict 🤔

That's true, would catch at least some percentage of conflicts. I guess you could do something creative like auto revert master on fail of pr merge or something, but sounds hard and hazardous.

PR #151, opened 2021-04-23
Tila 546 | Allocation results API permission changes

nicobav:
First of all makes the allocation results api end point as readonly. 
Second, makes the allocation results viewable only for service sector and general admin. => service sector id is required parameter.


aprenfors (approved):
👍

PR #152, opened 2021-04-23
Tila 534 units in allocation

aprenfors:


applications/allocation_result_mapper.py
```diff
@@ -20,22 +20,23 @@ def to_events(self):
                 application_event_schedule = ApplicationEventSchedule.objects.get(
                     pk=allocated_event.occurrence_id
                 )
-                ApplicationEventScheduleResult.objects.update_or_create(
+
+                ApplicationEventScheduleResult.objects.create(
```
nicobav, line 24: I think changing this to create only can introduce duplicate results :thinking:  I noticed that at start of the allocation this should get deleted => Would it still be safer to stick with the update_or_create?

applications/allocation_result_mapper.py
```diff
@@ -20,22 +20,23 @@ def to_events(self):
...
+                    allocated_end=allocated_event.end,
+                    allocated_day=application_event_schedule.day,
                 )
+
                 ApplicationEventStatus.objects.create(
                     status=ApplicationEventStatus.ALLOCATED,
                     application_event=application_event_schedule.application_event,
                 )
-            except ApplicationEventScheduleResult.DoesNotExist:
+            except Exception:
```
nicobav, line 38: this wide exception catching will get some nags in some code scanners. Perhaps most widest acceptable error could be `Error` from django.db.utils

aprenfors: Yeah it's a bit nasty. Logic here is that if this fails to anything the users are basically locked out of starting any allocations for the application round, and currently we don't have any mechanism to open it up other that fixing the database.

applications/allocation_result_mapper.py
aprenfors: Maybe it would be safer. Was thinking if we want it to fail on duplicates, because it's basically an error (not 100% sure how django handles one to one situations where more than 2 refer to it, does it actually fail on create or just pass through without error)?

nicobav: django throws an error if duplicate entry is tried to save to db, so perhaps my duplicate result worry wasn't that a big deal. we

allocation/allocation_runner.py
```diff
@@ -4,9 +4,14 @@
...
 from allocation.models import AllocationRequest
 from applications.allocation_result_mapper import AllocationResultMapper
+from applications.models import ApplicationEventScheduleResult
 
 
 def start_allocation(allocation_request: AllocationRequest):
+    ApplicationEventScheduleResult.objects.filter(
+        accepted=False,
+        application_event_schedule__application_event__application__application_round=allocation_request.application_round,  # noqa: E501
+    ).delete()
```
nicobav, line 14: I think it would be nice if the deletion of the schedule result would be in mapper class?

aprenfors: Yeah, i think that would be a better place, so those get deleted only when we actually get a solution.

aprenfors: Although then we need to add some logic for skipping the accepted events in AllocationDataBuilder but that's probably fine.

aprenfors: Never mind, i'm stupid it's already handled in ApplicationEvent#get_not_scheduled_occurrences ...

nicobav (approved):

PR #154, opened 2021-04-27
TILA-545 | Create reservations

nicobav:
Adds reservation creation to application event status api when status  `APPROVED` is created.
Adds new permission for unit managers to be able to create `VALIDATED` status. 
Only unit managers can add VALIDATED status and service sector managers can also create APPROVED statuses which causes the reservation creation.


nicobav:
Woosh, this went through few permission iterations and I had some difficulties to fully get the permissions idea.

nicobav:
Woosh, this went through few permission iterations and I had some difficulties to fully get the permissions idea.

applications/utils/reservation_creation.py
```diff
@@ -0,0 +1,63 @@
...
+
+def create_reservation_from_schedule_result(result, application_event):
+    recurring_reservation = RecurringReservation.objects.create(
+        user=application_event.application.user,
+        application=application_event.application,
+        age_group=result.application_event_schedule.application_event.age_group,
+        ability_group=result.application_event_schedule.application_event.ability_group,
+    )
+
+    reservation_date = application_event.begin
```
aprenfors, line 31: I think there is a bug here, running this with 

```
day = 0
reservation_date = datetime.date(2021, 4, 27)
```

returns 2021-04-28 when it should return 2021-05-03 which is the next monday after 2021-04-27

I think something like this should be correct

```
def next_matching_weekday(date, weekday):
    days_ahead = weekday - date.weekday()
    if days_ahead < 0: 
        days_ahead += 7
    return date + datetime.timedelta(days_ahead)

print(next_matching_weekday(datetime.date(2021, 4, 27), 0))
```

applications/utils/reservation_creation.py
```diff
@@ -0,0 +1,63 @@
...
+    if not reservation_date.weekday() == result.allocated_day:
+        reservation_date = reservation_date + datetime.timedelta(
+            days=abs(reservation_date.weekday() - result.allocated_day)
+        )
+    interval = 14 if application_event.biweekly else 7
+    while reservation_date < application_event.end:
+        try:
+            reservation = Reservation.objects.create(
```
aprenfors, line 39: There probably should be a check for overlaps

There is a ReservationUnit#check_reservation_overlap

which could be used.

api/tests/test_application_status_api.py
```diff
@@ -0,0 +1,243 @@
...
+        response = self.manager_api_client.post(
+            reverse("application_event_status-list"),
+            data={
+                "status": ApplicationEventStatus.APPROVED,
+                "application_event_id": self.application_event.id,
+            },
+        )
+        assert_that(response.status_code).is_equal_to(201)
+        assert_that(RecurringReservation.objects.count()).is_equal_to(1)
+        assert_that(Reservation.objects.count()).is_equal_to(2)
```
aprenfors, line 243: Could be good to add test for the date thingy.

applications/utils/reservation_creation.py
aprenfors: actually there is already basically this in 

date_util.py#next_or_current_matching_weekday

nicobav: very nicely catched :fishing_pole_and_fish:

nicobav: this should be now fixed.

applications/utils/reservation_creation.py
nicobav: added overlap logic to create the reservation as DENIED status if overlaps.

api/tests/test_application_status_api.py
nicobav: added the test

nicobav:
added new commit " Adds RecurringReservations to have fk ref to ApplicationEvent " since there is need in the applicant views to show which event caused which reservations

applications/utils/reservation_creation.py
aprenfors: 👍 denied sounds better than cancelled that i suggested in the meeting.

aprenfors (approved):

PR #160, opened 2021-05-03
Tila 578 occurrence constraint

aprenfors:


allocation/tests/test_allocation_solver.py
```diff
@@ -625,3 +625,108 @@ def test_should_allocate_when_event_num_persons_is_none(
 
     assert len(solution) == 1
     assert solution[0].begin == datetime.time(hour=10, minute=0)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
```
nicobav, line 631: this looks awesome :joy:

nicobav (approved):

allocation/tests/test_allocation_solver.py
aprenfors: Yeah should really write class based/use factories, but this is kind of copy pasta 🤦 Maybe there will be some time to refactor at some point.

PR #161, opened 2021-05-03
TILA 572 | Check opening hours from HAUKI when creating reservations

nicobav:


applications/tests/test_reservation_creation.py
```diff
@@ -0,0 +1,245 @@
...
+            start_dt.month,
+            start_dt.day,
+            12,
+            0,
+            tzinfo=get_default_timezone(),
+        )
+        assert_that(start_dt).is_equal_to(expected_res_begin)
+        assert_that(end_dt).is_equal_to(self.reservation_end)
+
+    def test_reservation_times_are_none_when_the_reservation_unit_is_not_open_in_reservation_time(
```
aprenfors, line 193: KInd of same here. I think maybe we should create the reservation with the times from result, and just mark it cancelled if it doesn't fit within opening times.

applications/tests/test_reservation_creation.py
```diff
@@ -0,0 +1,245 @@
...
+    ):
+        scheduler = ReservationScheduler(
+            self.res_unit.id, self.reservation_begin, self.reservation_end
+        )
+        start_dt, end_dt = scheduler.get_reservation_times_based_on_opening_hours()
+
+        assert_that(start_dt).is_equal_to(self.reservation_begin)
+        assert_that(end_dt).is_equal_to(self.reservation_end)
+
+    def test_reservation_time_is_one_hour_reduced_from_reservation_time_when_closes_before_reservation_end(
```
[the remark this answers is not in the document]
nicobav: Yes, that might be the case. I thought that the customer would have the possiblity to use the space at least some time if it wouldn't fit within the opening hours since allocation won't allocate to same time afaik?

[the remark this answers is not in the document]
aprenfors: Ok, looks like this is good as it is.

applications/tests/test_reservation_creation.py
aprenfors: I think in this case we should just use the times requested and create the reservation as DENIED

nicobav: Added the logic and a test for it.

api/tests/test_application_status_api.py
```diff
@@ -274,18 +292,16 @@ def test_correct_dates_are_used_for_reservations(self):
...
         res_date = next_or_current_matching_weekday(
             self.application_event.begin, self.result.allocated_day
         )
-        res_dt = datetime.datetime.combine(res_date, self.result.allocated_begin)
+        res_dt = datetime.datetime.combine(res_date, self.result.allocated_begin, tzinfo=get_default_timezone())
         ReservationFactory(
             begin=res_dt,
             end=res_dt + datetime.timedelta(hours=12),
             reservation_unit=[self.result.allocated_reservation_unit],
+            state=STATE_CHOICES.CREATED
```
nicobav, line 304: Found a reason why this was flickering :see_no_evil:

aprenfors (approved):

aprenfors:
Looks like there is still a lint error, but other than that looks good.

nicobav:
> Looks like there is still a lint error, but other than that looks good.

oh, unused import from removing the test skipping based on __version__

PR #162, opened 2021-05-04
TILA-573 | Recurring reservation api

nicobav:


api/reservations_api.py
```diff
@@ -156,6 +164,71 @@ def get_queryset(self):
         )
 
 
+class RecurringReservationSerializer(serializers.ModelSerializer):
+    reservations = ReservationSerializer(many=True)
+
+    class Meta:
+        model = RecurringReservation
+        fields = [
+            "application",
```
aprenfors, line 173: Little bit inconsistent field naming compared to other serializers. When returning just id or list of ids, should be named appliation_id, foobar_ids etc.

nicobav: renamed as they should :+1:

aprenfors (approved):

PR #163, opened 2021-05-05
TILA-586 Enable removing weekly reductions.

aprenfors:


api/applications_api/filters.py
```diff
@@ -8,3 +8,13 @@ class ApplicationFilter(filters.FilterSet):
         field_name="application_round", queryset=ApplicationRound.objects.all()
     )
     status = filters.BaseInFilter(field_name="cached_latest_status")
+
+
+class ApplicationEventWeeklyAmountReductionFilter(filters.FilterSet):
+    application_event_id = filters.NumberFilter(method="filter_application_event")
```
nicobav, line 14: `filter.ModelChoiceFilter` would be nice built-in filter to use here.  If you use it, the method below can be removed.

nicobav (changes requested):

api/applications_api/filters.py
nicobav: I know there's some other filtering made with NumberFilter / method (by me) => those should be refactored at some point :hourglass:

aprenfors: Yeah, i'll change this

nicobav (approved):

PR #164, opened 2021-05-05
Fix application filtering.

aprenfors:
Don't allocate declined or cancelled applications.


aprenfors:
This kind of doesn't solve the actual problem in prod database, although this was a bug too. Just kind of happens to solve it by accident, there is another bug with multi day schedules.

nicobav (approved):

PR #167, opened 2021-05-07
TILA-571 | Application event aggregate data

nicobav:
Similarly like the application aggregate data. 
The event aggregate data creation is called at the end of the application aggregated data creation. 
 - It's run in its own thread to avoid API response latency (same could be done to application aggregate data as well)
 - The above is not that nice solution since the operation can be only be monitored through logging - not that bad, but we might want some manageable background operation solution (celery, django-rq, django-cron etc.)
 
 
 **Deployment notes** 
 -  Need to run the `event.create_aggregate_data()` creation for every event of application which has IN_REVIEW status or higher


nicobav:
hmph, some test failures when running all together => have to get rid of dependencies with each other since they pass when running single.

nicobav:
> hmph, some test failures when running all together => have to get rid of dependencies with each other since they pass when running single.

This is done so its ok to review.

applications/models.py
```diff
@@ -770,6 +778,71 @@ def get_all_occurrences(self):
...
+            events_count = len(schedule.get_occurences().occurrences)
+            total_events.append(events_count)
+
+            time_dt_begin = datetime.timedelta(
+                hours=schedule.begin.hour, minutes=schedule.begin.minute
+            )
+            time_dt_end = datetime.timedelta(
+                hours=schedule.end.hour, minutes=schedule.end.minute
+            )
+            event_duration = time_dt_end - time_dt_begin
```
aprenfors, line 794: I don't think begin and end can be used to determine duration, they are basically the start and end time of the "suitable" time range. So i think you should just get the duration from the application event min duration, which should be maybe good enough estimate.

aprenfors (changes requested):

applications/models.py
nicobav: good point :thinking:  I'll change that

nicobav: this is changed

aprenfors (approved):

PR #171, opened 2021-05-10
Tila 592| ScheduleResultAggregateData

nicobav:


applications/models.py
```diff
@@ -818,6 +824,62 @@ def create_aggregate_data(self):
...
+            )
+            total_amount_of_events.append(amount_of_events)
+            total_events_duration.append(
+                (
+                    amount_of_events
+                    * schedule.application_event_schedule_result.allocated_duration
+                ).total_seconds()
+            )
+
+        total_reservations = sum(
```
aprenfors, line 850: Do we actually need to care about events per week at this point? If there are these results and count of them exceeds the amount of events per week, which shouldn't happen but anyway, isn't the higher amount basically the correct answer?

nicobav: Well now that you pointed that out that it shouldn't happen that results exceeds the events per week amount, we shouldn't care the events per week and can trust the results.

nicobav: its more of an scenario before allocation when we try to guess what would happen in allocation in most optimistic case

aprenfors: Right, but the schedule results don't exist before allocation, so isn't the estimation handled in ApplicationEvent#create_aggregate_data

nicobav: yes exactly. Was just pointing out that your comment was correct => we should care only the result data. I'll change this.

aprenfors (approved):

applications/models.py
nicobav: Changed

PR #174, opened 2021-05-11
Refactor get_opening_hours resource_id parameter

nicobav:
Remove the need for giving HAUKI_ORIGIN_ID as part of the resource_id in
get_opening_hours function.
Moves the HAUKI_ORIGIN_ID to be added to resource_id when doing the
query to HAUKI

### DIFF
--- a/applications/utils/reservation_creation.py
+++ b/applications/utils/reservation_creation.py
@@ -52,7 +52,7 @@ def create_reservation_from_schedule_result(result, application_event):
         )
 
         reservation_scheduler = ReservationScheduler(
-            result.allocated_reservation_unit.id, res_start, res_end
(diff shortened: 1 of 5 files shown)


applications/utils/reservation_creation.py
```diff
@@ -152,7 +152,7 @@ def can_continue_to_next_day(self, date, time):
     def get_opening_hours(self):
         # TODO: Cache opening hours per unit.
         return get_opening_hours(
-            self.reservation_unit_id,
+            self.reservation_unit.unit.id,
```
aprenfors, line 155: Should be reservation_unit.uuid

aprenfors (approved):

PR #175, opened 2021-05-14
Tila 554 | ApplicationRound aggregate data

nicobav:


reservations/models.py
```diff
@@ -98,7 +104,31 @@ class STATE_CHOISE_CONST(object):
 STATE_CHOICES = STATE_CHOISE_CONST()
 
 
+class ReservationQuerySet(models.QuerySet):
```
aprenfors, line 107: Very nice method names 👍

reservations/models.py
```diff
@@ -98,7 +104,31 @@ class STATE_CHOISE_CONST(object):
...
+    def total_duration(self):
+        return self.annotate(duration=F("end") - F("begin")).aggregate(
+            total_duration=Sum("duration")
+        )
+
+    def within_application_round_period(self, app_round: ApplicationRound):
+        return self.filter(
```
aprenfors, line 114: One could maybe wonder if reservation starts before the res period but ends within the res period should be partly part of used capacity, but doesn't really matter, since we don't really need super exact numbers here.

aprenfors (approved):

nicobav:
:facepalm:  forgot the api representation => added now.

aprenfors (approved):

PR #185, opened 2021-05-17
TILA-502 | Reservation unit capacity api

nicobav:
Adds method to reservation unit api which gets the capacity within given parameters


reservations/models.py
```diff
@@ -116,6 +116,12 @@ def within_application_round_period(self, app_round: ApplicationRound):
             end__lte=app_round.reservation_period_end,
         )
 
+    def within_period(self, period_start, period_end):
```
aprenfors, line 119: I guess you could call this method from within_application_round_period to reuse this, but this is quite simple, so not really necessary.

aprenfors (approved):

reservations/models.py
nicobav: changed this => merging

PR #188, opened 2021-05-18
Fixed extra comma after option.

aprenfors:


aprenfors:
Looks like there is something weird going on with apt mirror, build failure not related to this pr, main fails also.

nicobav (approved):

PR #195, opened 2021-05-21
TILA-622 | Reservation and Recurring reservation api adds

nicobav:


api/reservations_api.py
```diff
@@ -85,6 +160,22 @@ class Meta:
             },
         }
 
+    def get_begin_weekday(self, instance):
+        return instance.begin.weekday()
+
+    def get_application_event_name(self, instance):
```
aprenfors, line 166: Not related to this pull request, but i think this whole recurring reservations seems quite unnecessary, could just link reservations straight to application events imo. 

But let's see about refactoring that at some time.

aprenfors (approved):

api/reservations_api.py
nicobav: yes, that is true, this `RecurringReservation`  might be unnecessary boilerplate model in the middle.

aprenfors: Yeah, it's from way back from initial models, when we didn't even have any actual recurring reservations and stuff.

nicobav: we could think of the v2 api at near future and drop this then.

PR #201, opened 2021-05-27
TILA-640 | Add summary info from allocation results to application round api

nicobav:
Adds allocation result aggregate data summary info to application round
api. Renders as part of the app round's aggregated data.

### DIFF
--- a/api/application_round_api.py
+++ b/api/application_round_api.py
@@ -1,9 +1,11 @@
 import logging
 
 from django.conf import settings
+from django.db.models import Sum
 from rest_framework import permissions, serializers, viewsets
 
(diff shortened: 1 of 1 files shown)


nicobav:
Actually going to do some changes and it this data straight to application round serializer.

aprenfors:
Was wondering if this is something we want to show to everyone since app round api is basically open for anyone, but i guess it makes sense, since you could use it to show how likely you'll get a reservation etc., and doesn't contain anything  secret.

aprenfors (approved):

kieferbonk (approved):

PR #202, opened 2021-05-31
TILA-629 | Make partial update possible on application round

nicobav:


aprenfors (approved):

nicobav:
@aprenfors I added one new commit here concerning permissions

kieferbonk:
+1

permissions/api_permissions.py
```diff
@@ -170,7 +170,9 @@ def has_permission(self, request, view):
         if request.method in permissions.SAFE_METHODS:
             return True
 
-        service_sector_id = request.data.get("service_sector_id")
+        service_sector_id = request.data.get("service_sector_id", None)
+        if not service_sector_id and view.detail:
```
aprenfors, line 174: Hmm, what's the case where we need this special case?

nicobav: When updating only single field (round's status), even though its detail it will bump first here and then go to the object permission check. TBH, I'm no sure for the whole existing service_sector check. But perhaps that's something to consider in the permission refactor.

aprenfors (approved):

PR #208, opened 2021-06-07
TILA-584 Api to get reservation unit calendar urls.

aprenfors:
Application event ical url and reservation unit ical url.


api/ical_api.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        buffer.seek(0)
+
+        return FileResponse(buffer, as_attachment=True, filename="file.ics")
+
+
+def reservation_unit_calendar(reservation_unit: ReservationUnit):
+    cal = Calendar()
+
+    cal.add("version", "2.0")
```
nicobav, line 130: should the version come from some variable / settings?

api/ical_api.py
```diff
@@ -0,0 +1,187 @@
...
+
+    reservations = Reservation.objects.filter(
+        recurring_reservation__application_event=application_event
+    )
+    for reservation in reservations:
+        ical_event = Event()
+        ical_event.add("summary", application_event.name)
+        ical_event.add("dtstart", reservation.begin)
+        ical_event.add("dtend", reservation.end)
+        ical_event.add("dtstamp", datetime.datetime.now() + datetime.timedelta(hours=1))
```
nicobav, line 182: why this timestamp needs to be one hour ahead of current time?

tilavarauspalvelu/settings.py
```diff
@@ -169,6 +169,7 @@ def get_git_revision_hash() -> str:
     HAUKI_ORIGIN_ID=(str, "tvp"),
     CSRF_TRUSTED_ORIGINS=(list, []),
     MULTI_PROXY_HEADERS=(bool, False),
+    ICAL_HASH_SECRET=(str, "qhoew923uqqwee"),
```
nicobav, line 172: perhaps just return empty if not in env.

applications/models.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import math
+import uuid as uuid
```
nicobav, line 4: `import uuid`

api/ical_api.py
```diff
@@ -0,0 +1,187 @@
...
+from rest_framework.reverse import reverse
+from rest_framework.viewsets import ViewSet
+
+from applications.models import ApplicationEvent
+from permissions.api_permissions import ReservationUnitCalendarUrlPermission
+from reservation_units.models import ReservationUnit
+from reservations.models import Reservation
+
+
+class ExampleSerializer(serializers.ModelSerializer):
```
nicobav, line 22: :construction_worker_man:

nicobav (changes requested):

api/ical_api.py
aprenfors: 🤦

api/ical_api.py
aprenfors: I guess it could be good.

api/ical_api.py
aprenfors: It shouldn't, good catch. No idea why i put it like this.

aprenfors:
Wonder why that test started failing... Shouldn't really be anything in this that changes it, but oh well.

api/ical_api.py
aprenfors: Setting might be a bit heavy handed, maybe just a constant would be better.

nicobav (approved):
:ship:

PR #210, opened 2021-06-08
TILA-369 | API validation tuning

nicobav:
status validations for application, application round

Address / contact information required for application.


api/applications_api/serializers.py
```diff
@@ -426,7 +426,7 @@ class ApplicationSerializer(serializers.ModelSerializer):
     contact_person = PersonSerializer(
         help_text="Contact person information for the application",
         read_only=False,
-        allow_null=True,
+        allow_null=False,
```
aprenfors, line 429: I don't think we can set this to not allowing null, since the application is done in "steps", so  we don't have the possibility to add contact person until later. Probably need to validate this in validate_for_review

aprenfors (changes requested):

api/applications_api/serializers.py
nicobav: is it so? I thought this would be initial when user wants to make the application

nicobav: I'll move this to `validate_for_review` :+1:

aprenfors: Yeah, i think you save the first version of application with pretty much nothing

aprenfors (approved):

PR #213, opened 2021-06-10
Bump django to 3.1.10

nicobav:


aprenfors:
Maybe missing a merge from main or something, since it complains about ical dependency missing.

requirements.txt
```diff
@@ -126,7 +126,6 @@ graphql-core==2.3.2
     #   graphql-relay
 graphql-relay==2.0.1
     # via graphene
-icalendar==4.0.7
```
nicobav, line 129: ahh, this one got removed by pip-compile

aprenfors: Looks like it's my bad, it's missing from req.in 🤦

nicobav: I added it there now

aprenfors (approved):

PR #215, opened 2021-06-15
Tila 584 application event UUID

aprenfors:


api/ical_api.py
```diff
@@ -136,12 +159,19 @@ def application_event_calendar(application_event: ApplicationEvent) -> Calendar:
     cal = Calendar()
 
     cal.add("version", ICAL_VERSION)
-    organisation_name = (
-        application_event.application.organisation.name
+
+    applicant_name = (
```
mingfeng, line 163: It could be only me, I find it difficult to follow the logic with one-liner if-elses, might be better to just use if-else blocks. Not sure if I understand the logic correctly, but just for comparison:

```python
organisation = application_event.application.organisation
contact_person = application_event.application.contact_person

if organisation:
    application_name = organisation.name
elif contact_person:
    application_name = f"{contact_person.first_name} {contact_person.last_name}"
else:
    application_name = ""
```

mingfeng: If the contract_person is a standard user model, you could do `contact_person.get_full_name()`

aprenfors: It's not an user, rather something that can be filled in by the person filling the application.

mingfeng (approved):
Looks superb 👍  :shipit:

PR #220, opened 2021-06-23
TILA-652 Add round applications sent status to application round.

aprenfors:


api/application_round_api.py
```diff
@@ -225,6 +229,17 @@ def get_aggregated_data(self, instance):
         allocation_result_dict.update(instance.aggregated_data_dict)
         return allocation_result_dict
 
+    def get_applications_sent(self, instance: ApplicationRound):
+        not_sent = next(
+            filter(
```
nicobav, line 234: This works. Although I would do this with queryset like: `instance.applications.filter(cached_latest_status__in=(ApplicationStatus.IN_REVIEW, ApplicationStatus.REVIEW_DONE)).exists()`
Benefit for queryset way is that it makes a SQL query with where statement with the statuses rather than looping them in memory.

aprenfors: Perfect, i was actually going to ask if there is a way to do it with query, couldn't find a way to do it with status.latest easily, missed this cached status altogether 👍

nicobav (approved):

PR #222, opened 2021-06-28
TILA-555 | Bandit and safety checks

nicobav:


aprenfors:
Looks like the safety check is failing with command not being found.

nicobav:
:facepalm:  fixed that one

aprenfors (approved):

PR #223, opened 2021-07-01
TILA-707 Graphql fetch api for reservation units.

aprenfors:
Simple snapshot sanity test case to build on.

### DIFF
--- a/api/graphql/tests/snapshots/snap_test_reservation_units.py
+++ b/api/graphql/tests/snapshots/snap_test_reservation_units.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# snapshottest: v1 - https://goo.gl/zC4yUc
+from __future__ import unicode_literals
+
+from snapshottest import Snapshot
+
+snapshots = Snapshot()
+
+snapshots["ReservationUnitCapacityTestCase::test_getting_reservation_units 1"] = {
+    "data": {"reservationUnits": {"edges": [{"node": {"name": "Test name"}}]}}
+}
--- a/api/graphql/tests/test_reservation_units.py
+++ b/api/graphql/tests/test_reservation_units.py
@@ -1,0 +1,35 @@
+import json
+
+import snapshottest
+from assertpy import assert_that
+from graphene_django.utils import GraphQLTestCase
+from rest_framework.test import APIClient
+
+from reservation_units.tests.factories import ReservationUnitFactory
+
+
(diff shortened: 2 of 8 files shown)


api/graphql/tests/snapshots/snap_test_reservation_units.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# snapshottest: v1 - https://goo.gl/zC4yUc
+from __future__ import unicode_literals
+
+from snapshottest import Snapshot
```
nicobav, line 5: I guess we cannot test permissions with this or can we?

api/graphql/tests/test_reservation_units.py
```diff
@@ -0,0 +1,35 @@
...
+
+import snapshottest
+from assertpy import assert_that
+from graphene_django.utils import GraphQLTestCase
+from rest_framework.test import APIClient
+
+from reservation_units.tests.factories import ReservationUnitFactory
+
+
+class ReservationUnitCapacityTestCase(GraphQLTestCase, snapshottest.TestCase):
```
nicobav, line 11: Test class naming could be like `ReservationUnitTestCase`

nicobav (changes requested):
Looks nice - one test naming thingy

api/graphql/tests/test_reservation_units.py
aprenfors: Ooops, copy pasta.

api/graphql/tests/snapshots/snap_test_reservation_units.py
aprenfors: Not with this one. 

Took the snapshot test dep, so it's not so tedious to write asserts. As you probably noticed there is only one super simple test here, i'll write more, but wanted to get this to review so you have time to review it before you go to vacation.

nicobav: yup, this looks nice. Was asking because I'm trying to write some permission related tests atm to gql.

nicobav:
as general note - if the schema file is getting huge I think we could refactor the schema file and split it to multiple files when that happens.

aprenfors:
> as general note - if the schema file is getting huge I think we could refactor the schema file and split it to multiple files when that happens.

Yep. It's already a bit large. Could refactor the types to separate file(s).

nicobav (approved):

PR #231, opened 2021-07-06
TILA-606 Configure celery for running backend tasks.

aprenfors:
Configures celery for running backend tasks. 

Note that this isn't production ready, specifically docker related stuff, since we are using filesystem backend. 

But this gives use a way to start implementing backend tasks with celery without doing changes to infrastructure, and configuration is done so that moving this to redis or similar message queue should be very simple.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,6 @@ coverage.xml
 .pytest_cache/
 
 .mypy_cache/
+
+# Celery filesystem queue
+broker/
--- a/.env.example
+++ b/.env.example
@@ -209,3 +209,49 @@ SENTRY_ENVIRONMENT=local-development-unconfigured
 #by anything so you can subscribe to them from calendars,
 #we need to hash those, so you can't access those just by uuid
 #ICAL_HASH_SECRET=
+
+
+# CELERY
+
+# Set true if you want to use celery to run backend tasks. Otherwise they will be run synchronously.
+# Default is true
+#CELERY_ENABLED=True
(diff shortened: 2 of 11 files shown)


.dockerignore
```diff
@@ -3,4 +3,7 @@
 coverage.xml
 .pytest_cache/
 
-.mypy_cache/
\ No newline at end of file
+.mypy_cache/
+
+# Celery filesystem queue
+broker/
```
klemmari1, line 9: No newline at the end of file.

.env.example
```diff
@@ -208,4 +208,50 @@ SENTRY_ENVIRONMENT=local-development-unconfigured
...
+#CELERY_FILESYSTEM_BACKEND=(bool, True),
+
+# Folder to use for celery queue out. Defaults to ./broker/queue/
+#CELERY_QUEUE_FOLDER_OUT=
+
+# Folder to use for celery queue in. Defaults to ./broker/queue/
+#CELERY_QUEUE_FOLDER_IN=
+
+# Folder to use for storing processed tasks. Defaults to ./broker/processed/
+#CELERY_PROCESSED_FOLDER=
```
klemmari1, line 257: No newline at the end of file.

klemmari1 (approved):
Just two small comments about the newlines. Do as you please with them.

Otherwise looks good to me!

.env.example
aprenfors: 👍

PR #236, opened 2021-08-09
TILA-718 | permission classes for gql 

nicobav:


permissions/helpers.py
```diff
@@ -273,6 +275,15 @@ def can_view_reservation(user: User, reservation: Reservation) -> bool:
     )
 
 
+def can_view_reservations(user):
```
aprenfors, line 278: This is probably too broad permission, but maybe this'll be done later since there is another stub right below?

nicobav: I think we could and probably need to show reservations for all users (authenticated) but just not include the user information. Probably different gql queries for those :thinking:

aprenfors: Yeah, probably something like minimalized dto that can be used for reserved dates/times etc.

aprenfors (approved):

PR #238, opened 2021-08-11
TILA-727 | Add gql query to query reservation units by unit

nicobav:
This is now basing the TILA-726 branch for diff's sake. When that gets merged can change the base to main.

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -1,3 +1,4 @@
+import django_filters
 import graphene
 from graphene import Field, relay
 from graphene_django.forms.mutation import DjangoModelFormMutation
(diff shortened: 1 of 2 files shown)


api/graphql/schema.py
```diff
@@ -37,9 +43,17 @@ class AllowAuthenticatedFilter(AuthFilter):
     permission_classes = (AllowAuthenticated,)
 
 
-class ReservationUnitsFilter(AuthFilter):
+class ReservationUnitsFilter(AuthFilter, django_filters.FilterSet):
     permission_classes = (ReservationUnitPermission,)
 
+    unity = ModelInFilter(field_name="unit", queryset=Unit.objects.all())
```
aprenfors, line 49: Unity?

nicobav: :see_no_evil:  typo => i'll fix

nicobav: oh, actually this is completely unnecessary here => was testing FilterSets with bad results => I'll remove this `FilterSet` experiment here

nicobav: Okay, actually got the `FilterSet` class to work.

nicobav:
Added new `FilterSet` class for `ReservationUnitFIlter` => naming the `FilterSet` class as `ReservationUnitFilterSet` (usually naming convention is just `NNNFilter` but it would be confusing with the django filter connections :exploding_head: 
Anyhow the unit filtering works now nicely within the `reservationUnits` filter

aprenfors (approved):

PR #241, opened 2021-08-20
TILA-735 | next avalable reservation time for reservation unit

nicobav:
Kinda initial solution since this is quite slow. Restricted this to be only in reservation_units_by_pk query.

TILA-735


reservation_units/utils/reservation_unit_reservation_scheduler.py
```diff
@@ -0,0 +1,140 @@
+import datetime
+
+import pytz
+from pytz import UTC
+
+from opening_hours.utils.opening_hours_client import OpeningHoursClient
+
+
+class ReservationUnitReservationScheduler:
+    MARCH = 3
```
aprenfors, line 10: Seems to be unused

reservation_units/utils/reservation_unit_reservation_scheduler.py
```diff
@@ -0,0 +1,140 @@
...
+        self.reservation_unit = reservation_unit
+
+        if self.reservation_unit.max_reservation_duration:
+            self.reservation_duration = (
+                self.reservation_unit.max_reservation_duration.total_seconds() / 3600
+            )
+        else:
+            self.reservation_duration = 1
+
+        self.start_time = datetime.datetime.now(tz=pytz.UTC) + datetime.timedelta(
```
aprenfors, line 23: Hmm, what's happening with utz + 2 hours?

nicobav: yup, actually should be the default time zone :flushed: => I'll fix this

aprenfors (approved):

PR #242, opened 2021-08-23
TILA-730 | Terms of use to gql api 

nicobav:
also reservation unit mutations.

Adds permission checker class for classes based on `SerializerMutation`


aprenfors (approved):

nicobav:
@aprenfors tested this before merging and  noticed some permission problems => fixed those in two newer commits

aprenfors (approved):

PR #257, opened 2021-09-06
TILA-738 | Opening hours gql

nicobav:
Makes possible to query opening hours for reservation unit and unit. 
  - adds fetching periods also.

TILA-738 TILA-750

### DIFF
--- a/api/graphql/tests/test_reservation_units.py
+++ b/api/graphql/tests/test_reservation_units.py
@@ -1,10 +1,16 @@
+import datetime
 import json
+from unittest import mock
 
 import snapshottest
 from assertpy import assert_that
+from django.conf import settings
+from django.test import override_settings
 from graphene_django.utils import GraphQLTestCase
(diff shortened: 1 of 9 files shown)


api/graphql/tests/test_reservation_units.py
```diff
@@ -104,3 +110,134 @@ def test_should_error_when_not_found_by_pk(self):
...
+                    periods=[
+                        1,
+                    ],
+                ),
+            ],
+        },
+    ]
+
+
+def get_mocked_periods():
```
aprenfors, line 185: Quite a lot of test data setup but i guess it's pretty unavoidable.

aprenfors (approved):

api/graphql/tests/test_reservation_units.py
nicobav: yup its kinda bloating but, we expect this kind of data to be fetched

PR #260, opened 2021-09-07
Use bearer header instead of jwt.

aprenfors:


nicobav (approved):
Is the `Bearer:` what UI is using already with tunnistamo?

aprenfors:
> Is the `Bearer:` what UI is using already with tunnistamo?

Yeah.

PR #262, opened 2021-09-09
TILA-769 | Unit mutation enchancements

nicobav:
Also finetunes permissions checks for auth serializer mutations to raise permission error.

### DIFF
--- a/api/graphql/units/unit_serializers.py
+++ b/api/graphql/units/unit_serializers.py
@@ -3,5 +3,15 @@ from api.units_api.serializers import UnitSerializer
 
 
 class UnitUpdateSerializer(UnitSerializer, PrimaryKeyUpdateSerializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


api/graphql/units/unit_serializers.py
```diff
@@ -3,5 +3,15 @@
 
 
 class UnitUpdateSerializer(UnitSerializer, PrimaryKeyUpdateSerializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["tprek_id"].required = False
```
aprenfors, line 8: How does the validation work with all of these set to not required?

nicobav: This case I don't see the problem when the tprek is the master (I actually don't know why we're editing the unit data in tvp :flushed: ) so this is just an extension to modify unit data.
Of course we might add some fields afterwards to unit level and so on but those optional marked fields are optional in gql update (create not even possible atm).

aprenfors (approved):

PR #268, opened 2021-09-14
TILA-750 | Read origin id from hauki

nicobav:
Previously we were using the resource id from hauki as origin id from
hauki but those two are separate ids and that caused some problems in
OpeningHoursClient.
This fixes that so that the origin id is read which is the id that tvp
uses for resources.

TILA-750


nicobav:
Need to check those test failures

nicobav:
@aprenfors this is ready now

aprenfors (approved):

PR #272, opened 2021-09-16
TILA-776 | Resource mutations and query changes

nicobav:


api/graphql/resources/resource_serializers.py
```diff
@@ -0,0 +1,99 @@
...
+from api.graphql.base_serializers import (
+    PrimaryKeySerializer,
+    PrimaryKeyUpdateSerializer,
+)
+from api.resources_api import ResourceSerializer
+from resources.models import Resource
+
+
+class ResourceCreateSerializer(ResourceSerializer, PrimaryKeySerializer):
+    location_type = serializers.CharField(
```
aprenfors, line 12: Should this be ChoiceField, might get better validation?

nicobav: I got some weird error if this was `ChoiceField` (or was not redefined at all): `AssertionError: Found different types with the same name in the schema: location_type, location_type.`

But your comment in mind => have to write validation check for location type

aprenfors: Looks like it might be a bug with having type in field name 😮 

https://github.com/graphql-python/graphene-django/issues/185#issuecomment-314130927

nicobav: I tested this with changing the model name (migrations and all) to not contain type but it still nagged about it. It could be something related with `ChoiceField` go and tell :shrug:  
Anyhow I added the validations for the infamous `location_type`

aprenfors (approved):

PR #273, opened 2021-09-17
Opening hours fetch fix

nicobav:
There were found two "unwanted features" which caused opening hours and nextavailableSlot apis to not work properly.

### DIFF
--- a/opening_hours/hours.py
+++ b/opening_hours/hours.py
@@ -165,7 +165,7 @@ def get_opening_hours(
         for opening_hours in day_data_in["opening_hours"]:
             day_data_out = {
                 "resource_id": day_data_in["resource"]["id"],
-                "origin_id": day_data_in["resource"]["origins"]["origin_id"],
+                "origin_id": day_data_in["resource"]["origins"][0]["origin_id"],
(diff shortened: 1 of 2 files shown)


opening_hours/hours.py
```diff
@@ -165,7 +165,7 @@ def get_opening_hours(
         for opening_hours in day_data_in["opening_hours"]:
             day_data_out = {
                 "resource_id": day_data_in["resource"]["id"],
-                "origin_id": day_data_in["resource"]["origins"]["origin_id"],
+                "origin_id": day_data_in["resource"]["origins"][0]["origin_id"],
```
aprenfors, line 168: Hmm wonder if this will cause trouble if querying both tvp and tprek opening hours at once, but i guess the way it's done currently it shouldn't happen.

aprenfors (approved):

opening_hours/hours.py
nicobav: yes, you cannot use the function to fetch two origins

nicobav: Don't se the use case for it now but surely we could implement such feature enhancement later if its needed.

PR #278, opened 2021-09-21
Upgrade  sqlparse, pillow and elasticapm

nicobav:
dependabot warned about sqlparse and pillow. Elastic apm had new 5.x version.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -31,7 +31,7 @@ psycopg2==2.8.6
 pytest-cov==2.10.1
 pytest-django==4.1.0
 pytest-factoryboy==2.1.0
-elastic-apm==5.10.0
+elastic-apm==5.10.1
 sentry-sdk==0.19.5
 snapshottest
 social-auth-app-django==4.0.0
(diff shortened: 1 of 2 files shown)


aprenfors (approved):
Wonder if we could bump django version with the upgraded elastic apm, iirc. the old one didn't work with 3.1

nicobav:
> Wonder if we could bump django version with the upgraded elastic apm, iirc. the old one didn't work with 3.1

Yes that would be nice to upgrade django to 3.2 - actually we have to do it this year since 3.1 is getting old end of this year. I did some minor testing and digging for upgrading to 3.2 and it requires to upgrade to latest elastic-apm 6.x and fixing some code stuff as well so it requires own ticket imo.

aprenfors:
> Yes that would be nice to upgrade django to 3.2 - actually we have to do it this year since 3.1 is getting old end of this year. I did some minor testing and digging for upgrading to 3.2 and it requires to upgrade to latest elastic-apm 6.x and fixing some code stuff as well so it requires own ticket imo.

Yeah, didn't think it should be done here.

PR #280, opened 2021-09-23
TILA-781 | Reservation unit mutations enchancements and new fields

nicobav:
TILA-711 TILA-781


jmp (approved):
Looks good to me! Nice comprehensive test cases! 👍

PR #284, opened 2021-09-28
TILA-737: Add ability to get reservations for reservation units

jmp:
This PR adds `reservations` field for reservation units in the GraphQL API.

This allows getting reservations e.g. for a specific reservation unit:

```graphql
{
  reservationUnitByPk(pk: 1) {
    reservations {
      begin
      end
      state
    }
  }
}
```
Currently the main use case is to get all reservations for the given reservation unit for a specific time range (e.g. for a week).

For this purpose, the reservations can be filtered by timestamps and state:

```graphql
{
  reservationUnitByPk(pk: 1) {
    reservations(from: "2020-01-01T00:00:00Z", to: "2020-01-07T00:00:00Z", state: "CREATED") {
      begin
      end
      state
    }
  }
}
```

This would return all the reservations for the reservation unit `1` with state `CREATED` state and whose begin or end time is between `2020-01-01` and `2020-01-07`.


nicobav (approved):
LGTM :+1:

PR #287, opened 2021-09-30
Tila 739 graphql translations

aprenfors:


nicobav:
> UnitType and UnitByPkType (why do we have two types for the same data?):
ReservationUnitType, ReservationUnitByPkType (again why the duplication?):

@siren  Those `ByPkTypes` are there because they define extra fields (opening hours) which are not available for mass queries due to the potential high risk of (accidental) misuse of the queries to cause slowness in the system.

api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -82,10 +83,20 @@ def resolve_keyword_groups(self, info):
...
+
 class PurposeType(AuthNode, PrimaryKeyObjectType):
+    name_fi = graphene.String()
+    name_sv = graphene.String()
+    name_en = graphene.String()
+
     class Meta:
         model = Purpose
-        fields = ("id", "name", "pk")
+        fields = ("id", "pk", "name_fi", "name_sv", "name_en")
```
nicobav, line 99: I think we could use the `get_all_translatable_fields` here

api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -175,10 +188,12 @@ class EquipmentType(AuthNode, PrimaryKeyObjectType):
 
     class Meta:
         model = Equipment
-        fields = ("name", "id")
+        fields = ("id", "name_fi", "name_sv", "name_en")
```
nicobav, line 191: I think we could use the `get_all_translatable_fields` here

api/graphql/resources/resource_types.py
```diff
@@ -21,16 +27,22 @@ class Meta:
         fields = (
             "id",
             "location_type",
-            "name",
-            "description",
+            "name_fi",
```
nicobav, line 30: I think we could use the `get_all_translatable_fields` here

api/graphql/spaces/space_types.py
```diff
@@ -9,6 +9,10 @@
 
 
 class DistrictType(PrimaryKeyObjectType):
+    name_fi = graphene.String()
```
nicobav, line 12: would it be enough to define these fields in the `Meta` with `get_all_translatable_fields` ?

nicobav (changes requested):
Some notes about using the `get_all_translatable_fields`

PR #289, opened 2021-10-07
TILA-809 | Add purposes filter to gql

nicobav:
Purpose filter query to GQL schema.

TILA-809

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_types.py
+++ b/api/graphql/reservation_units/reservation_unit_types.py
@@ -23,6 +23,7 @@ from permissions.api_permissions.drf_permissions import ApplicationRoundPermissi
 from permissions.api_permissions.graphene_permissions import (
     EquipmentCategoryPermission,
     EquipmentPermission,
+    PurposePermission,
     ReservationUnitHaukiUrlPermission,
     ReservationUnitPermission,
 )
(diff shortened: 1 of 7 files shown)


api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -83,9 +84,15 @@ def resolve_keyword_groups(self, info):
...
 
 class PurposeType(AuthNode, PrimaryKeyObjectType):
+    permission_classes = (
+        (PurposePermission,) if not settings.TMP_PERMISSIONS_DISABLED else (AllowAny,)
+    )
+
     class Meta:
         model = Purpose
-        fields = ("id", "name", "pk")
+        fields = ("id", "pk", "name_fi", "name_en", "name_sv")
```
jmp, line 93: I was thinking more about this, and I think it might be okay to leave `name` here for now, because removing it will break the existing API and reservation unit / purpose tests.

We could do a separate PR where we remove the non-language specific fields from all types.

The other option would be to fix the broken reservation unit and purpose tests that already rely on `name` being there.

api/graphql/schema.py
```diff
@@ -181,6 +196,10 @@ def resolve_equipment_category_by_pk(self, info, **kwargs):
         pk = kwargs.get("pk")
         return get_object_or_404(EquipmentCategory, pk=pk)
 
+    def resolve_purpose_by_pk(self, info, **kwargs):
+        pk = kwargs.get("pk")
+        return get_object_or_404(Purpose, pk=pk)
+
```
jmp, line 202: I think we don't need this anymore since we got rid of `purpose_by_pk` in the schema.

api/graphql/reservation_units/reservation_unit_types.py
nicobav: I'll fix the tests and let's change the api since that's the way we want it to be.

PR #292, opened 2021-10-18
TILA-729 | Single reservation gql base

nicobav:


api/graphql/schema.py
```diff
@@ -54,32 +60,53 @@
...
+        )
+        return queryset.filter(
+            Q(reservation_unit__unit__in=get_units_where_can_view_reservations(user))
+            | Q(
+                reservation_unit__unit__service_sectors__in=get_service_sectors_where_can_view_reservations(
+                    user
+                )
+            )
+            | Q(user=user)
+        ).order_by("begin")
```
jmp, line 109: What do you think about doing something like this instead of relying on a specific user called `admin` to exist:
```python
user = info.context.user
viewable_units = get_units_where_can_view_reservations(user)
viewable_service_sectors = get_service_sectors_where_can_view_reservations(user)
if settings.TMP_PERMISSIONS_DISABLED:
    viewable_units = Unit.objects.all()
    viewable_service_sectors = ServiceSector.objects.all()
return queryset.filter(
    Q(reservation_unit__unit__in=viewable_units)
    | Q(reservation_unit__unit__service_sectors__in=viewable_service_sectors)
    | Q(user=user)
).order_by("begin")
```
Maybe an alternative would be to get the the first user with `is_superuser=True`, but I think it would be better not to assume any admin user exists.

api/graphql/tests/snapshots/snap_test_reservation_units.py
```diff
@@ -79,12 +79,12 @@
                             {
                                 'begin': '2021-05-03T00:00:00+00:00',
                                 'end': '2021-05-03T01:00:00+00:00',
-                                'state': 'CREATED'
+                                'state': 'created'
```
jmp, line 82: Just a note, we might have some inconsistency now between how the `state` is given in the input vs. how it appears in the output.

Currently you can filter reservations like this (state name in all caps):
```graphql
reservations(state: ["CREATED", "CONFIRMED"]) {
    begin
    end
    state
}
```
But in the output it will now appear in lowercase, like:
```js
'reservations': [{
    'begin': '2021-05-03T00:00:00+00:00',
    'end': '2021-05-03T01:00:00+00:00',
    'state': 'created',
}]
```
I'm not sure if using lowercase in the filters (`state: ["created", "confirmed"]`) works at the moment, but probably we should use either uppercase or lowercase consistently in both input and output.

siren:
I tried to create reservation with this version of api (with auth disabled) and I am getting the following error, should it work?

```
{
  "errors": [
    {
      "message": "Cannot assign \"<SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x7fc85d8c4400>>\": \"Reservation.user\" must be a \"User\" instance.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createReservation"
      ]
    }
  ],
  "data": {
    "createReservation": null
  }
}

nicobav:
> I tried to create reservation with this version of api (with auth disabled) and I am getting the following error, should it work?
> 
> ```
> {
>   "errors": [
>     {
>       "message": "Cannot assign \"<SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x7fc85d8c4400>>\": \"Reservation.user\" must be a \"User\" instance.",
>       "locations": [
>         {
>           "line": 2,
>           "column": 3
>         }
>       ],
>       "path": [
>         "createReservation"
>       ]
>     }
>   ],
>   "data": {
>     "createReservation": null
>   }
> }
> ```

It fails since in the save part it sets the reservation user from the request - which is anonymous user in your case. Perhaps I'll add some ugliness to that part if we have TMP_PERMISSIONS_DISABLED :thinking:

nicobav:
> > I tried to create reservation with this version of api (with auth disabled) and I am getting the following error, should it work?
> > ```
> > {
> >   "errors": [
> >     {
> >       "message": "Cannot assign \"<SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x7fc85d8c4400>>\": \"Reservation.user\" must be a \"User\" instance.",
> >       "locations": [
> >         {
> >           "line": 2,
> >           "column": 3
> >         }
> >       ],
> >       "path": [
> >         "createReservation"
> >       ]
> >     }
> >   ],
> >   "data": {
> >     "createReservation": null
> >   }
> > }
> > ```
> 
> It fails since in the save part it sets the reservation user from the request - which is anonymous user in your case. Perhaps I'll add some ugliness to that part if we have TMP_PERMISSIONS_DISABLED thinking

And by that I mean to set the user as `None` when `TMP_PERMISSIONS_DISABLED` - any thoughts @jmp @siren ?  I mean that will work the case but it's going to downgrade the readability of the code.

siren:
> > It fails since in the save part it sets the reservation user from the request - which is anonymous user in your case. Perhaps I'll add some ugliness to that part if we have TMP_PERMISSIONS_DISABLED thinking
> 
> And by that I mean to set the user as `None` when `TMP_PERMISSIONS_DISABLED` - any thoughts @jmp @siren ? I mean that will work the case but it's going to downgrade the readability of the code.

Perhaps we should just write instructions on how to configure the local backend to use auth and get rid of this "tmp" permission hack completely?

api/graphql/schema.py
nicobav: I actually need to include the user to get this TMP_PERMISSIONS_DISABLED to work as it supposed for now.

api/graphql/tests/snapshots/snap_test_reservation_units.py
nicobav: > Currently you can filter reservations like this (state name in all caps):

Actually checked that this isn't the case we cannot currently query reservations => I think we can add the filter for state in scope of its own ticket. So I'll focus on changing the mutation inputs to take the lower cases

jmp (approved):
Looks good to me now! 👍

PR #300, opened 2021-10-29
TILA-893 | Make openinghours client timezone aware

nicobav:
Adds new class to represent opening hours with start and end times using datetimes for handier comparing and handling.
Reads timezone information from hauki for opening hours and uses that in the `OpeningHoursClient` class. 

Note that UI is now receiving opening hours as default timezone (currently UTC0)

### DIFF
--- a/opening_hours/decorators.py
+++ b/opening_hours/decorators.py
@@ -1,0 +1,26 @@
+import datetime
+
+from django.utils.timezone import get_default_timezone
+
+TIMEZONE = get_default_timezone()
+
+
+def datetime_args_to_default_timezone(func: callable):
+    def convert_to_utc(*args, **kwargs):
+        converted_args = []
+        for arg in args:
+            if isinstance(arg, datetime.datetime):
+                arg = arg.astimezone(TIMEZONE)
+            elif isinstance(arg, datetime.time):
+                timezone = arg.tzinfo
+                if not timezone:
+                    timezone = TIMEZONE
+                dt = timezone.localize(
+                    datetime.datetime(1977, 1, 1, arg.hour, arg.minute, arg.second)
+                )
+                arg = dt.astimezone(TIMEZONE).time()
+            converted_args.append(arg)
+
+        return func(*converted_args, **kwargs)
+
+    return convert_to_utc
--- a/opening_hours/utils/opening_hours_client.py
+++ b/opening_hours/utils/opening_hours_client.py
@@ -1,14 +1,74 @@
 import datetime
-from typing import Dict, List
+from typing import Dict, List, Union
 
+import pytz
 from django.conf import settings
+from django.utils.timezone import get_default_timezone
 
+from opening_hours.decorators import datetime_args_to_default_timezone
 from opening_hours.hours import (
     Period,
     TimeElement,
     get_opening_hours,
     get_periods_for_resource,
 )
+
+TIMEZONE = get_default_timezone()
+
+
+class OpeningHours:
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    resource_state: str
+    periods: List[int]
+
+    def __init__(
+        self,
+        start_time: datetime.datetime,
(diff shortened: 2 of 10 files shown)


opening_hours/utils/opening_hours_client.py
```diff
@@ -1,15 +1,70 @@
...
+        periods: List[int],
+    ):
+        self.start_time = start_time
+        self.end_time = end_time
+        self.resource_state = resource_state
+        self.periods = periods
+
+    @classmethod
+    def get_opening_hours_class_from_time_element(
+        cls, time_element: TimeElement, date: datetime.date, timezone: pytz.UTC
```
jmp, line 39: Can `timezone` be something different from `pytz.UTC` here?

If `timezone` is always `pytz.UTC`, then we would not need the parameter and can just use the `pytz.UTC` singleton everywhere in the function.

But I suspect we can pass non-UTC timezones here as well. In that case the type hint should be broader.

As far as I can tell, `timezone` can be one of these:

* `pytz.UTC` (singleton)
* an instance of `pytz.tzinfo.DstTzInfo` (common, e.g. "Europe/Helsinki")
* an instance of `pytz.tzinfo.StaticTzInfo` (rare)

Unfortunately all of these implement the `localize` method _independently_ (not overriding a base class), so there is no shared base class we can use. 😭 So the "correct" type hint I would look something like this:

```python
timezone: Union[pytz.UTC, pytz.tzinfo.DstTzInfo, pytz.tzinfo.StaticTzInfo]
```

It's a bit fugly. 😕 But I'm not sure if there's a better solution here because of how `pytz` has designed their class hierarchy.

opening_hours/decorators.py
```diff
@@ -0,0 +1,26 @@
...
+        converted_args = []
+        for arg in args:
+            if isinstance(arg, datetime.datetime):
+                arg = arg.astimezone(TIMEZONE)
+            elif isinstance(arg, datetime.time):
+                timezone = arg.tzinfo
+                if not timezone:
+                    timezone = TIMEZONE
+                dt = timezone.localize(
+                    datetime.datetime(1977, 1, 1, arg.hour, arg.minute)
```
jmp, line 19: Seconds are lost when the time passes through this function.

I think it probably doesn't matter much in our use cases since I think we're not really dealing with one-second granularity. But we could add `arg.second` here for completeness. What do you think?

nicobav: Yep I agree, those seconds would be good to have _just_in_case_

nicobav: Fixed this.

opening_hours/utils/opening_hours_client.py
nicobav: Fixed this.

jmp: Ouch, looks like `pytz.UTC` in the `Union` doesn't work because the class name is shadowed in `pytz`:

```python
class UTC(BaseTzInfo):
    ...

UTC = utc = UTC()  # UTC is a singleton
```

It seems like you have to use `type(pytz.UTC)` instead of `pytz.UTC`, something like this:

```python
timezone: Union[type(pytz.UTC), pytz.tzinfo.DstTzInfo, pytz.tzinfo.StaticTzInfo]
```
😢

nicobav: yeah, type hints are hard when your language so loose :joy: :cry:  I change this to be fetching the UTC type (whic is `_UTC`) as you suggested. After all this is just for IDE or possible future python (pre)compiler. pytz.UTC would kinda work but it is an instance (not a class/type) as you said (singleton instance) and therefore if there would be some timezone inheriting the `_UTC` class it would be causing warning even though it should work.

jmp (approved):
Looks good. Nice work, Nico! 😄

nicobav:
I added one commit about displaying the the time with the time difference info (e.g `12:00:00+00:00`)

PR #301, opened 2021-11-01
TILA-892 | Use custom opening hours end date in ReservationCreateSerializer 

nicobav:
There was a situation that caused the reservation not to be created even it should've been when the reservation would occur after next spring. 
The reason was that the reservation creation validator uses `ReservationUnitReservationScheduler` class which were initially made for fetching next available opening times until next spring (Point when the recurring reservation applications start) and thus it loads opening hours only to that point of time.
This adds new optional date parameter to scheduler class which is used for fetching opening hours until specific date.


jmp (approved):
🚀

PR #302, opened 2021-11-02
TILA-777 | Add check_resolver permission check decorator and use it in resolvers

nicobav:
Decorator that is intended to be used with resolver functions for fields
and in schema for "byPk" fields

TILA-777


jmp (approved):
Looks nice! 💯

PR #304, opened 2021-11-04
TILA-801 | Create resources to hauki

nicobav:
Functionality that sends reservation units to Hauki as resources when reservation are created or updated through api => mutated. Updates the resource if it exists and creates if it does not.


nicobav:
> Why only after they are public? doesn't this mean that there's no way to edit opening hours but after the res unit is made public? isn't it too late?

Good point, it might be too late. Draft requires a name which is a good thing since we're sending the reservation unit name as the resource name. The concern with drafts are that then the name can be somewhat "foggy" at that point. Also the description is sent and it can be empty also. One thing could be that we send it send it even when draft and edit it always when editing. Perhaps we could raise this discussion to slack

nicobav:
Changing the sending logic a bit.

nicobav:
Okay now ready for review again and a now this time it will create or update the resource in reservation unit mutation.

reservation_units/utils/hauki_exporter.py
```diff
@@ -0,0 +1,85 @@
...
+        if not unit:
+            return None
+        unit_resource_id = (
+            f"{unit.hauki_resource_data_source_id}:{unit.hauki_resource_origin_id}"
+        )
+        url = urljoin(settings.HAUKI_API_URL, "/")
+        url = urljoin(
+            url,
+            f"/v1/resource/{unit_resource_id}",
+        )
```
jmp, line 34: I think this should have the same effect as the two `urljoin`s:
```python
url = urljoin(
    settings.HAUKI_API_URL,
    f"/v1/resource/{unit_resource_id}",
)
```

spaces/importers/units.py
```diff
@@ -131,4 +146,63 @@ def create_unit(self, importer_data: dict) -> bool:
...
+                        self.resource_id_map[origin["origin_id"]] = resource["id"]
+            return counter
+
+        data = make_hauki_get_request(url, params)
+        counter = read_response(counter)
+        next = data.get("next", False)
+        while next:
+            data = make_hauki_get_request(next, params)
+            counter = read_response(counter)
+            next = data.get("next", False)
```
jmp, line 198: I think this could be simplified to:
```python
next = url
while next:
    data = make_hauki_get_request(next, params)
    counter = read_response(counter)
    next = data.get("next", False)
```

jmp (approved):
Looks good to me. 🙂👍  Some very minor suggestions for possible simplification etc., but it's good as it is too.

reservation_units/utils/hauki_exporter.py
nicobav: True it should

spaces/importers/units.py
nicobav: This whole paging was a bummer since the `page_size` is limited to 1000... Hoping we could get rid of it in the future

PR #305, opened 2021-11-05
TILA-947 | Allow reservation state change to cancelled in updateMutation

nicobav:
TILA-947


jmp (approved):
Nice! Looking good. 🙂

PR #306, opened 2021-11-05
TILA-920 Reservation unit purpose

jmp:
This PR adds model for reservation unit purposes and GraphQL API for querying the purposes.

The current `Purpose` model contains the values that we need for reservations and recurring reservations, which describes the activity category for a reservation/event. For example:

* Harrastustoiminta, bändi ja musiikki
* Liikuntaseuran harjoitusvuoro
* Aatteellinen, hengellinen tai yhteiskunnallinen tilaisuus
* Senioritoiminta
* etc.

What we don't have is a model for reservation unit purposes, which describes the types of activities that are possible with that reservation unit:
* Järjestä tapahtuma tai näytös
* Sauno ja mökkeile
* Luo uutta tai korjaa vanhaa
* Liiku ja pelaa
* Pidä kokous
* etc.

So I decided to keep the `Purpose` model for the reservation purposes, and create a new `ReservationUnitPurpose` for the reservation unit purposes. This should mean that we don't have to touch the REST APIs related to purposes / recurring reservations.

The most important changes are:
* New `ReservationUnitPurpose` model
* Admin model for managing the `ReservationUnitPurpose` via Django admin
* `ReservationUnit.purpose` is now a foreign key to `ReservationUnitPurpose` instead of `Purpose`
* GraphQL API for getting reservation unit purposes


nicobav (approved):
LGTM :+1:

PR #308, opened 2021-11-09
TILA-920: Reservation purpose changes

jmp:
This PR makes the following changes:

* The `Purpose` model will be used for reservation unit purposes
* New `ReservationPurpose` model will be used for reservation purposes

At the moment, the `Purpose` model is used to mean "reservation purpose", so I've written a migration that converts the current purposes into the new reservation purposes.

There already existed a model called `ReservationPurpose`, which linked the original `Purpose` to a `Reservation`. I got rid of that and replaced it with a `purpose` field (in the `Reservation` model) that refers to a `ReservationPurpose`.

The new reservation purposes can be queried via GraphQL like this:

```graphql
{
  reservationPurposes {
    edges {
      node {
        nameFi
        nameSv
        nameEn
      }
    }
  }
}
```

A purpose can be given with the `purposePk` field when creating a reservation, like this:

```graphql
mutation {
  createReservation(input: {
    name: "Test reservation"
    begin: "2021-01-01T00:00:00Z"
    end: "2021-01-01T01:00:00Z"
    reservationUnitPks: []
    purposePk: 1
  }) {
    errors {
      field
      messages
    }
  }
}
```


jmp:
Updated the PR with new fixtures based on the data model changes.

nicobav (approved):

PR #309, opened 2021-11-09
TILA-907 | Reservation cancellation api when confirmed

nicobav:
Adds cancel reservation gql api to cancel reservations that are confirmed.

Introduces models for predefined cancel reasons and cancellation rules. The rules are to be widen in the future but this initial solution contains only possibility to define the cancellation time.

`cancelReservation` takes `pk` and `cancelReasonPk` as required input and `cancelDetails` as optional.
Returns `state` and `errors` (also the input fields)

TILA-907


jmp (approved):
Looks good to me. 🙂 I was wondering whether we also need a custom cancellation reason (a text field). But it's something that could be checked with the BO and can be added in a separate PR too if needed.

nicobav:
> Looks good to me. slightly_smiling_face I was wondering whether we also need a custom cancellation reason (a text field). But it's something that could be checked with the BO and can be added in a separate PR too if needed.

The custom cancellation reason text is [here](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/309/files#diff-29cb7c2cf4d804ece9ee069fc29e866f35aa7a258c556a31dce8e7f9076d3a19R228) as `cancel_details`

jmp (approved):
Looks good! 👍

PR #311, opened 2021-11-11
TILA-877: Terms of Use

jmp:
This PR adds a new `TermsOfUse` model that can be used for generic payment, cancellation, service-specific or other terms.

Reservation units get new fields:

* Payment terms (FK to `TermsOfUse`)
* Cancellation terms (FK to `TermsOfUse`)
* Service-specific terms (FK to `TermsOfUse`)

Reservations also get one new field:

* `confirmed_at` which gets updated automatically by the `confirmReservation` mutation.

Terms of use field has been removed from `Space` since it's not needed.


nicobav (approved):
LGTM :+1:

PR #320, opened 2021-11-17
TILA-955: Add test for loading fixtures

jmp:
Added a new test that runs the `loaddata` management command with `fixtures/cases.json` as input, to make sure the fixtures get loaded successfully.


nicobav (approved):
:rocket:

PR #321, opened 2021-11-17
TILA-955: Verify that data model visualization is up-to-date

jmp:
This PR makes a few changes:

* The data model visualization image format changed from PNG to SVG. The PNG is a binary that takes some MBs of space, so if we update it regularly, it will keep bloating the repository size quite a lot. By comparison, the SVG only takes a fraction of the space, and is text format.
* Moved the graph model parameters to `settings.py`, so that they can be changed in one place, and the same settings can be used automatically when generating the graph from the command line or test cases.
* Added a new test for verifying that the data model visualization is up-to-date. This generates the graph model as JSON, and compares it to the one stored in the repository. If the graphs are different, the test fails. The order of elements in the JSON vary a little bit between different runs, so to fix that I sorted the items.


nicobav (approved):
Awesome :+1:

PR #322, opened 2021-11-18
TILA-895 | Use state enum in Reservation query 

nicobav:
In `ReservationType` state had been defined as string type even though it is enum. This changes it to be enum instead of string.
In mutations though we are sticking with strings since there's a unsolved bug in gql+drf serializers with using enums and thus this adds some helping text for state field in mutations.

TILA-895


jmp (approved):
Looks good. 🚀

nicobav:
> Looks good. rocket (EDIT: Well, other than the tests not passing laughing)

yep :sweat_smile:  I'll update snapshots =>

PR #324, opened 2021-11-18
TILA-877 | Update cases.json fixture file to not include terms_of_use for spaces

nicobav:
The terms of use was removed in TILA-877 and it didn't have yet fixture load test and now that it is merged it'll fail => this removes
those obsolete terms of use fields from fixtures cases.json.

TILA-877


jmp (approved):
Good catch! 👍

PR #325, opened 2021-11-18
Try to fix broken azure/olpenshift build as suggested

siren:
Tested locally and the build seems to work ok


nicobav (approved):
:crossed_fingers:

nicobav (approved):

PR #326, opened 2021-11-19
TILA-977: Add reservationByPk query to GraphQL API

jmp:
This PR adds `reservationByPk` to the GraphQL API, so a single reservation can be queried, e.g.:

```graphql
{
  reservationByPk(pk: 1) {
    pk
    name
    description
    reserveeFirstName
    reserveeLastName
    reserveePhone
    state
    priority
    user
    purpose {
      nameFi
    }
    ...
  }
}
```


api/graphql/tests/test_reservations.py
```diff
@@ -1114,3 +1114,29 @@ def test_cancel_reservation_fails_when_reservation_in_past(self):
         assert_that(cancel_data.get("errors")).is_not_none()
         reservation.refresh_from_db()
         assert_that(reservation.state).is_equal_to(STATE_CHOICES.CONFIRMED)
+
+
+class ReservationQueryTestCase(ReservationTestCaseBase):
```
nicobav, line 1119: Perhaps add a check that another user cannot see the reservee name.

nicobav (approved):
Looks good, just one test suggestion.

api/graphql/tests/test_reservations.py
jmp: Added a test for this, which revealed that another user could've actually seen the reservee details. So I added new resolvers for each of the reservee fields and now they will return `None` if the user is not allowed to view those fields (e.g. user is different and not a superuser).

PR #328, opened 2021-11-19
TILA-929: Upgrade Graphene-Django to the latest version

jmp:
This PR upgrades Graphene-Django to version 3.0.0b7, which is the latest version. Along with it, Graphene will be upgraded to version 3.0.

3.0.0b7 is a pre-release, but there hasn't been a new release for almost a year, so I'm not sure if there are any other options because the stable releases don't support Graphene 3.0. We need Graphene 3.0 for `null` value support for input fields (see https://github.com/graphql-python/graphene/issues/1362).

The Graphene-Permissions library has a small problem where it imports `ResolveInfo` from the wrong package, causing an `ImportError` with newer versions of Graphene. There is no fix for this yet, so the only way to avoid is to monkey patch it in `settings.py`. I will make a separate PR in the graphene-permissions project to have it fixed in a future version.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -22,7 +22,7 @@ drf-spectacular==0.13.0
 easy-thumbnails==2.7.1
 flake8==3.8.4
 freezegun==0.3.14
-graphene-django==2.13.0
+graphene-django==3.0.0b7
 graphene_permissions==1.1.4
 icalendar==4.0.7
 isort==5.6.4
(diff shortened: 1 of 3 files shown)


nicobav (approved):
very nice :rocket:

jmp:
Related issue opened in graphene-permissions: https://github.com/redzej/graphene-permissions/issues/35 ([PR](https://github.com/redzej/graphene-permissions/pull/36)).

PR #329, opened 2021-11-19
TILA-877: Fix primary key type for terms

jmp:
The PK fields for the different terms was mistakenly defined as integers even though the terms of use PKs are strings.

This PR fixes the reservation unit GraphQL type so that those PKs are the correct type.

TILA-877

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -129,19 +129,19 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         source="cancellation_rule",
         required=False,
     )
-    payment_terms_pk = IntegerPrimaryKeyField(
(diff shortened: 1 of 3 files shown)


nicobav (approved):
I missed the reason for TermsOfUse having id field as charfield?

jmp:
> I missed the reason for TermsOfUse having id field as charfield?

Basically to allow the frontend to hardcode the ID for terms of use instances that are not reservation unit specific, if needed.

If there are multiple `TermsOfUse `instances in the database with the same type (e.g. "General terms of use", or"General terms of use for recurring reservations"), then the UI needs some way to pick the correct one to be shown when the user is shown the reservation terms. A string-based ID is better for this than an integer PK if it needs to be hardcoded in the UI code. It's also the way it currently works in Respa.

The other option would've been to have an integer ID but then that integer ID still needs to be hardcoded in the frontend code. Or, the frontend has to assume that there is exactly one terms of use instance with that type in the database. By having the ID as a string, both options are possible.

PR #331, opened 2021-11-22
TILA-975 | Set FILE_UPLOAD_PERMISSIONS to None

nicobav:
We use azure persistent storage for /media and it does not allow chmod any kind of permission changes for the folders and files. By default storage has this setting on and by setting this to None should skip the chmod  when uploading files.

TILA-975


jmp (approved):
Looks good, let's hope this works. 👍

PR #332, opened 2021-11-22
Graphene compatibility fixes

jmp:
This PR makes a couple of minor changes to the codebase to improve compatibility with future versions of Graphene:

* Fix some tests to use integer PKs instead of string PKs (new Graphene versions seem to be more strict about the types)
* Import `ResolveInfo` from the correct location (from the `graphene` package instead of the private `graphql` package)


nicobav (approved):
awesome :pray:

PR #333, opened 2021-11-22
Use ModelMultipleChoiceFilter instead of ModelInFilter for consistency and Graphene-Django 2.15+ compatibility

jmp:
The `ModelInFilter` from our `common_filters.py` causes problems with Graphene-Django 2.15 and above, because it inherits from both `BaseInFilter` and `ModelChoiceFilter`. Another problem with it is that it requires multiple PKs to be given as a string, like `"1,2,3"` instead of a list like `[1,2,3]`.

I noticed that for GraphQL APIs, we can actually skip `ModelInFilter` altogether and just use `ModelMultipleChoiceFilter` directly, which allows filtering by a single or multiple PKs given as a list. So instead of this:

```graphql
reservationUnits(reservationUnitType: "1,2,3") {
   ...
}
```
...we can do this:
```graphql
reservationUnits(reservationUnitType: [1,2,3]) {
   ...
}
```
And filtering by a single PK still works the same:
```graphql
reservationUnits(reservationUnitType: 1) {
   ...
}
```

This leaves all the filters in `common_filters.py` used only by our REST APIs.

This will also make the filters compatible with the latest version of Graphene-Django (2.15.0 and above), including the pre-releases which support Graphene 3.0.

Note, I also added tests for all of the reservation unit fields which used the `ModelInFilter` to make sure filtering by multiple PKs is possible.

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_filtersets.py
+++ b/api/graphql/reservation_units/reservation_unit_filtersets.py
@@ -5,7 +5,6 @@ import django_filters
 from django.db.models import Q, Sum
 from django_filters import CharFilter
 
-from api.common_filters import ModelInFilter
 from reservation_units.models import (
     KeywordGroup,
     Purpose,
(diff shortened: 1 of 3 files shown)


nicobav (approved):
excellent :pray:

PR #334, opened 2021-11-22
TILA-969 | Make reservation to be not cancellable if no rule

nicobav:
No cancellation rule no cancellation.

TILA-969


jmp (approved):
Looks good. 👍

PR #335, opened 2021-11-23
Fix deprecated test client in GraphQL tests

jmp:
In Graphene 3.x, the test client `_client` in `GraphQLTestCase` has been renamed to `client`.

This PR fixes the related deprecation warnings caused by the Graphene upgrade.


nicobav (approved):
Finally some reason for the naming from graphene :smiley:

PR #336, opened 2021-11-23
Fix order_by filter

nicobav:
Graphene upgrade broke bit the order_by filter. This fixes it and adds
tests for testing order_by filter.

### DIFF
--- a/api/graphql/tests/test_reservation_units.py
+++ b/api/graphql/tests/test_reservation_units.py
@@ -42,7 +42,7 @@ class ReservationUnitTestCase(GrapheneTestCaseBase, snapshottest.TestCase):
     @classmethod
     def setUpTestData(cls):
         super().setUpTestData()
-        cls.type = ReservationUnitTypeFactory(name="Test type")
+        cls.type = ReservationUnitTypeFactory(name="test type", name_fi="test type")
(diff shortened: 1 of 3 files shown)


api/graphql/tests/test_reservation_units.py
```diff
@@ -883,6 +883,182 @@ def test_filtering_by_is_draft_false(self):
...
+        ReservationUnitFactory(
+            unit=unit,
+        )
+        response = self.query(
+            """
+            query {
+                reservationUnits(orderBy: "unit") {
+                    edges {
+                        node {
+                            unit {pk}
```
jmp, line 1029: Let's use `nameFi` or some other constant field instead of `pk` to make the snapshots deterministic. 🙂 Same in `test_order_by_unit_reverse_order` below.

jmp (changes requested):
Looks good! The test execution order seems different in the GitHub pipeline which causes the tests to fail - let's make the tests a bit less dependent on the order.

api/graphql/tests/test_reservation_units.py
nicobav: fixed

jmp (approved):
👍

PR #337, opened 2021-11-26
Added pipeline yml files

Vitals9367:


azure-pipelines-devtest.yml
```diff
@@ -0,0 +1,46 @@
+# Continuous integration (CI) triggers cause a pipeline to run whenever you push 
+# an update to the specified branches or you push specified tags.
+trigger:
+  tags:
```
nicobav, line 4: I think this should be `branches`

nicobav (changes requested):

azure-pipelines-devtest.yml
Vitals9367: Yes, my mistake.

nicobav (approved):

PR #340, opened 2021-12-01
TILA-817: Add pricing fields to reservation units and reservations

jmp:
This PR adds pricing data for reservation units:

* Lowest price
* Highest price
* Price unit
* Price (for fixed price)

And for reservations:

* Price

I looked into making the price-related information a separate model, but the GraphQL API (will be done in a separate PR) started to get a bit too complicated compared to simply having the fields in the reservation unit. So in the end, I decided to keep them in the `ReservationUnit` model. It should be fairly straightforward to migrate to a separate model later if needed.


reservation_units/models.py
```diff
@@ -238,6 +238,47 @@ class ReservationUnit(models.Model):
...
+    PRICE_UNITS = (
+        (PRICE_UNIT_PER_15_MINS, _("per 15 minutes")),
+        (PRICE_UNIT_PER_30_MINS, _("per 30 minutes")),
+        (PRICE_UNIT_PER_HOUR, _("per hour")),
+        (PRICE_UNIT_PER_HALF_DAY, _("per half a day")),
+        (PRICE_UNIT_PER_DAY, _("per day")),
+        (PRICE_UNIT_PER_WEEK, _("per week")),
+        (PRICE_UNIT_FIXED, _("fixed")),
+    )
+    price_unit = models.CharField(
```
nicobav, line 257: Would be nice to have some `help_text` to tell what these prices mean. It would also be rendered into gql api when these fields gets publishe to there, but as well for us good info.

nicobav (changes requested):

reservation_units/models.py
jmp: Good point, fixed. 🙂

nicobav (approved):
Looks nice :rocket:

PR #349, opened 2021-12-02
TILA-818: Add price-related fields to the GraphQL API

jmp:
This PR adds price-related fields for reservation units and reservations to the GraphQL API.

## Reservation units

* `lowestPrice` (`Decimal`†)
  * This is the minimum price for the reservation unit, used when defining the price range in the management UI.
  * Optional, default value is 0.
* `highestPrice` (`Decimal`†)
  * This is the maximum price for the reservation unit, used when defining the price range in the management UI.
  * Optional, default value is 0.
* `priceUnit` (`String`‡)
  * The unit of the price. Possible values are `per_15_mins`, `per_30_mins`, `per_hour`, `per_half_day`, `per_day`, `per_week`, and `fixed`.
  * Optional, default value is `per_hour`.
* `price` (`Decimal`†)
  * The list price of the reservation unit. This is for cases where the price is fixed and this value will be copied to the reservation's price.
  * I'm not sure yet where the value for this will come from  (management UI? Django admin? Calculated based on lowest and highest?) because it's not shown in the UI mockups, but from previous discussions I understood that we need this field since the value must be copied from here to the reservation's `price` field.
  * Optional, default value is 0.

## Reservations

* `price` (`Decimal`†)
  * This is the final price of this particular reservation. It will be either decided by somebody based on the reservation unit's price range, or it will be copied directly from the reservation unit's `price` field (e.g. a reservation unit with a fixed price).
  * This field is *read-only` and will be filled by the backend when the reservation's final price is decided (not implemented yet since this part is still unknown).
  * Default value is 0.

### Notes:

† Graphene-Django [has a bug](https://github.com/graphql-python/graphene-django/blob/dd0d6ef28ffce33d30c4c8908f48a9d44a2b75ce/graphene_django/rest_framework/serializer_converter.py#L113) where it converts `DecimalField` into `Float` in mutations. There's no easy way to get around this. I'm planning to open a PR and hope they will fix it in the near future.

‡ This should actually be an `Enum`, but since we don't have a solution yet for handling enums in the GraphQL API, I have kept it as a `String` for consistency with other existing enums.

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -148,13 +148,30 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         required=False,
     )
     lowest_price = serializers.DecimalField(
-        max_digits=10, decimal_places=2, required=False
+        max_digits=10,
+        decimal_places=2,
+        required=False,
+        help_text="Minimum price of the reservation unit",
     )
     highest_price = serializers.DecimalField(
-        max_digits=10, decimal_places=2, required=False
+        max_digits=10,
+        decimal_places=2,
+        required=False,
+        help_text="Maximum price of the reservation unit",
     )
-    price_unit = serializers.CharField(required=False)
-    price = serializers.DecimalField(max_digits=10, decimal_places=2, required=False)
+    price_unit = serializers.CharField(
+        required=False,
+        help_text=(
+            "Unit of the price. "
+            f"Possible values are {', '.join(value for value, _ in ReservationUnit.PRICE_UNITS)}."
+        ),
+    )
+    price = serializers.DecimalField(
+        max_digits=10,
+        decimal_places=2,
+        required=False,
+        help_text="The current list price for reservation units with a fixed price",
+    )
 
     translation_fields = get_all_translatable_fields(ReservationUnit)
 
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -147,6 +147,14 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         source="service_specific_terms",
         required=False,
     )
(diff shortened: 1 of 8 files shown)


reservation_units/models.py
```diff
@@ -264,7 +264,7 @@ class ReservationUnit(models.Model):
         verbose_name=_("Price unit"),
         choices=PRICE_UNITS,
         default=PRICE_UNIT_PER_HOUR,
-        help_text="Unit of the price",
+        help_text=f"Unit of the price. Possible values are {', '.join(value for value, _ in PRICE_UNITS)}.",
```
nicobav, line 267: Commented above in the serializer about `help_text`

nicobav (approved):

nicobav:
Approved as is. You can decide how you like to handle the `help_text`

kieferbonk:
👍

reservation_units/models.py
nicobav: Actually this won't apply to Mutation documentation:
image

So I think you **must** define the `help_text` in the serializer to get it in the mutation documentation.

nicobav (changes requested):
The help_text needs to be in the serializer as well.

jmp:
I've created [an issue](https://github.com/graphql-python/graphene-django/issues/1276) and a [PR](https://github.com/graphql-python/graphene-django/pull/1277) to the Graphene-Django project for fixing the `DecimalField` to `Float` conversion bug in DRF serializers. Unfortunately, the project doesn't seem very active, but we'll see what happens. 😕

reservation_units/models.py
jmp: I think I'll move all the help texts to the serializers, they seem to be that way for other models. 👍

nicobav: I think the base `help_text` should be in the `model` and if the field gets re-defined/overridden in serializer then we must include the `help_text` to the serializer field.

jmp: I've modified it so that the base help texts are in the models, and if they are redefined in the serializer, the help texts are there as well. 🙂

nicobav (approved):
:sunglasses:

PR #350, opened 2021-12-02
TILA-988 | Search from translated fields in res unit text_search

nicobav:
TILA-988


api/graphql/reservation_units/reservation_unit_filtersets.py
```diff
@@ -62,16 +62,36 @@ def get_text_search(self, qs, property, value: str):
...
+                | Q(purposes__name_sv__icontains=word)
+                | Q(reservation_unit_type__name_fi__icontains=word)
+                | Q(reservation_unit_type__name_en__icontains=word)
+                | Q(reservation_unit_type__name_sv__icontains=word)
+                | Q(equipments__name_fi__icontains=word)
+                | Q(equipments__name_en__icontains=word)
+                | Q(equipments__name_sv__icontains=word)
+                | Q(unit__name_fi__icontains=word)
+                | Q(unit__name_en__icontains=word)
+                | Q(unit__name_sv__icontains=word)
```
jmp, line 94: I wonder if it would be possible to fetch all the translated fields dynamically and then construct the filter in e.g. a `for` loop. 🤔 But the current solution is okay too. 👍

api/graphql/tests/test_reservation_units.py
```diff
@@ -1113,6 +969,349 @@ def test_getting_terms(self):
...
+            query {
+                reservationUnits(textSearch:"Test type fi"){
+                edges {
+                    node {
+                        nameFi
+                        reservationUnitType {
+                            nameFi
+                        }
+                    }
+                }
```
jmp, line 985: Maybe we could indent these lines 2 spaces more to make the indentation consistent. (Also in the other queries below.

jmp (approved):
Looks good. 🙂 Some minor suggestions but it looks fine as-is too.

api/graphql/reservation_units/reservation_unit_filtersets.py
nicobav: yes it would, using `get_translatable_fields` perhaps. I'll check :+1:

api/graphql/tests/test_reservation_units.py
nicobav: Yes, would be more readable :+1:

api/graphql/reservation_units/reservation_unit_filtersets.py
nicobav: well, we could do something dynamic but that would perhaps make nonsense since most of the `Q`'s are focused on fk/m2m relations it would be too complex (imo) to proceed with that. What we currently have is that we can find out the model's translated fields and translated field of given model's specific field. But using those we need to know the models so we would need to specify nine (9) model relations and make'em loop etc... Perhaps you had something else in mind but at least this was my conclusion of this :upside_down_face:

PR #351, opened 2021-12-03
TILA-992: Add reservation start interval enum to the reservation unit model

jmp:
This PR adds a new `reservation_start_interval` field for reservation units.

It determines whether the start time of a reservation can be 15, 30, 60, or 90 minutes.

For example, if the interval is set to 15, then a reservation can start at e.g. 12:00, 12:15, 12:30, 12:45, etc. Or if the interval is set to 60, then a reservation can start at e.g. 12:00, 13:00, 14:00, 15:00, etc.

The field will be added to the GraphQL API in a separate PR.


nicobav (approved):
Looks good to me :heavy_check_mark: 
`ReservationUnit` model is starting to get humongous :mountain:

PR #352, opened 2021-12-03
TILA-999: Add reservation start interval to the GraphQL API

jmp:
Adds the new `reservationStartInterval` field for reservation units in the GraphQL API.

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -172,6 +172,15 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         required=False,
         help_text="The current list price for reservation units with a fixed price",
     )
+    reservation_start_interval = serializers.CharField(
+        required=False,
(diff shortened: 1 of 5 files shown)


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -172,6 +172,15 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         required=False,
         help_text="The current list price for reservation units with a fixed price",
     )
+    reservation_start_interval = serializers.CharField(
+        required=False,
+        help_text=(
+            "Determines the interval for the start time of the reservation. "
+            "For example an interval of 15 minutes means a reservation can "
+            "begin at minutes 0, 15, 30, or 45. Possible values are "
```
nicobav, line 180: Mentioning the possible values is good thing :+1: Perhaps also telling the enum type that is in the `Query` side.

nicobav (approved):
Looks good as is :rocket: 

Left one optional suggestion for `help_text`

PR #355, opened 2021-12-08
TILA-1043: Add tax percentage for reservation units

jmp:
Adds a new `tax_percentage` field for reservation units in the data model and GraphQL API.

The tax percentages can be fetched like this:

Query:
```graphql
{
  taxPercentages {
    edges {
      node {
        pk
        value
      }
    }
  }
}
```
Response:
```json
{
  "data": {
    "taxPercentages": {
      "edges": [
        {
          "node": {
            "pk": 1,
            "value": "0.00"
          }
        },
        {
          "node": {
            "pk": 2,
            "value": "10.00"
          }
        },
        {
          "node": {
            "pk": 3,
            "value": "14.00"
          }
        },
        {
          "node": {
            "pk": 4,
            "value": "24.00"
          }
        }
      ]
    }
  }
}
```

The tax percentage can be set for a reservation unit either in create or update mutation by PK:

```graphql
mutation {
  updateReservationUnit(input:{
    pk: 1
    taxPercentagePk: 1
    isDraft: true
  }){
    pk
  }
}
```

It can also be queried:

```graphql
{
  reservationUnitByPk(pk: 2) {
    pk
    taxPercentage {
      value
    }
  }
}
```
Response:
```json
{
  "data": {
    "reservationUnitByPk": {
      "pk": 2,
      "taxPercentage": {
        "value": "24.00"
      }
    }
  }
}
```


reservation_units/migrations/0034_reservationunit_tax_percentage.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='reservationunit',
+            name='tax_percentage',
+            field=models.DecimalField(choices=[(Decimal('0.00'), '0%'), (Decimal('10.00'), '10%'), (Decimal('14.00'), '14%'), (Decimal('24.00'), '24%')], decimal_places=2, default=Decimal('24.00'), help_text='The tax percentage for the price of the reservation unit', max_digits=5, verbose_name='Tax percentage'),
+        ),
+        migrations.AlterField(
+            model_name='reservationunit',
+            name='reservation_start_interval',
```
nicobav, line 22: I think we have this missing from`main` => perhaps separate PR with separate migration file.

nicobav (changes requested):

reservation_units/migrations/0034_reservationunit_tax_percentage.py
jmp: Nicely spotted. 🙂 Fixed now and added a new PR for that.

reservation_units/models.py
```diff
@@ -280,6 +281,25 @@ class ReservationUnit(models.Model):
         default=0,
         help_text="Maximum price of the reservation unit",
     )
+    TAX_PERCENTAGE_0 = Decimal("0.00")
```
nicobav, line 284: I guess there's been talks that tax percentages can be enums? Possible tax changes needs code changes, but that doesn't happen that often luckily, but for maintainability it would be nice that it would be own model (that can be edited in django admin).

jmp: Good idea -- the current is based on Respa which has it as an enum like this. But like you suggested, having it as a separate model would be better.

jmp: `TaxPercentage` now has a new model that can be edited via Django admin.

Also, the tax percentages have to be given as PKs for reservation units.

nicobav (changes requested):
Looks great :beers:  
There's other PRs merged and causing migrations to conflicting - though it could work (haven't tested) but we have duplicate numbering of migrations in reservation_units.

jmp:
> Looks great 🍻
> There's other PRs merged and causing migrations to conflicting - though it could work (haven't tested) but we have duplicate numbering of migrations in reservation_units.

Fixed the visualization and migration numbers. 👍

nicobav (approved):
Great :rocket:

PR #357, opened 2021-12-09
TILA-1010: Add validation for reservation start interval

jmp:
Adds validation for reservation start interval.

Validations for only 15, 30, and 60 would've been very simple (just check that minutes are 0, 15, or 30), but 90 minutes makes it a bit trickier:

* If the reservation unit opens at 8:00, then valid reservation start times are 8:00, 9:30, 11:00, 12:30, etc.
* If the reservation unit opens at 9:00, then valid reservation start times are 9:00, 10:30, 12:00, 13:30, etc.

So I had to utilize the opening time of the reservation unit for the validation.

TILA-1010


nicobav (approved):
Looks good :+1:

PR #358, opened 2021-12-09
TILA-1044: Add unit price and tax percentage to reservations

jmp:
Adds unit price and tax percentage to reservations, so that the UI (for example) doesn't have to look them up from the reservation unit's fields (which may change).

Reservations will have a new field in the GraphQL API:

Query:
```graphql
{
  reservationByPk(pk: 1) {
    unitPrice
    taxPercentage
  }
}
```
Response:
```json
{
  "data": {
    "reservationByPk": {
      "unitPrice": 0,
      "taxPercentage": "0.00"
    }
  }
}
```

TILA-1044


nicobav (approved):
:rocket:

PR #359, opened 2021-12-10
TILA-970 | Reservation unit image mutations

nicobav:
Introduces graphene file upload library which enables sending multipart request spec to be used in graphene by replacing the default graphqlview with the fileupload view.

Adds mutations for creating reservation unit image objects (with files [images]) 

https://github.com/jaydenseric/graphql-multipart-request-spec

TILA-970


jmp (approved):
This looks good to me. 👍 Nice job. 🙂

PR #362, opened 2021-12-10
Remove beta repos from Dockerfile

jmp:
This should hopefully resolve the build error:

    2021-12-10T12:03:14.9444100Z STEP 10: RUN subscription-manager repos --disable rhel-8-for-x86_64-baseos-beta-rpms
    2021-12-10T12:03:34.8932008Z Error: 'rhel-8-for-x86_64-baseos-beta-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.


nicobav:
Oh my... we had to add those lines few weeks ago since the build was failing otherwise :joy:

nicobav (approved):
:recycle:

PR #364, opened 2021-12-13
Upgrade to Django 3.1.14 for security fixes

jmp:
This bumps Django from 3.1.13 to 3.1.14, which fixes  CVE-2021-44420.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,7 +1,7 @@
 assertpy==1.1
 black==20.8b1
 celery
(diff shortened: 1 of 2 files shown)


nicobav (approved):

nicobav:
we should consider upgrading to django 3.2

nicobav:
> we should consider upgrading to django 3.2

Not by any means within this. Just a reminder for us

PR #365, opened 2021-12-13
TILA-899 | Split ReservationUnit buffer_time_between_reservation field to two fields

nicobav:
Also adds basic api references. Actual copying of the buffer times is done in separate PR.

TILA-899


jmp (approved):
Looks good to me! Nice job. 🙂

PR #366, opened 2021-12-14
TILA-900 | Copy buffer times from resunit to reservation in creation

nicobav:
TILA-900

### DIFF
--- a/api/graphql/reservations/reservation_serializers.py
+++ b/api/graphql/reservations/reservation_serializers.py
@@ -1,4 +1,5 @@
 import datetime
+from typing import List
 
 from django.conf import settings
 from django.utils.timezone import get_default_timezone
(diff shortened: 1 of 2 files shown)


api/graphql/reservations/reservation_serializers.py
```diff
@@ -131,6 +141,18 @@ def validate(self, data):
...
+        self, field: str, reservation_units: List[ReservationUnit]
+    ) -> [datetime.timedelta]:
+        buffer_times = [
+            getattr(res_unit, field)
+            for res_unit in reservation_units
+            if getattr(res_unit, field, None) is not None
+        ]
+        if len(buffer_times) > 0:
+            return max(buffer_times)
+        return None
```
jmp, line 154: Minor thing, but I think we could simplify this to:
```suggestion
        return max(buffer_times, default=None)
```

jmp (approved):
Looks good. 🙂 One very minor suggestion.

PR #367, opened 2021-12-14
TILA-1012 | Add models for publish and reservation begin date times

nicobav:
ReservationUnit should have information when the reservation is possible
(datetime) and when the reservation unit should be visible.

TILA-1012


jmp (approved):
👍

PR #369, opened 2021-12-15
TILA-1041: Metadata fields and sets

jmp:
This PR adds new models for `ReservationMetadataField` and `ReservationMetadataSet`. These are used to create forms for reservations you have to apply for.

New forms (metadata sets) can be created in the Django admin interface. Metadata fields cannot be created, modified or deleted in the admin since each field always require a corresponding field to exist in the `Reservation` model.

It also adds fields to the `Reservation` that are needed for the forms (all optional/nullable by default):
* reservee_type
* reservee_organisation_name
* reservee_address_street
* reservee_address_city
* reservee_address_zip
* reservee_email
* reservee_type
* reservee_id
* reservee_is_unregistered_association
* home_city
* applying_for_free_of_charge
* free_of_charge_reason
* age_group
* billing_first_name
* billing_last_name
* billing_address_street
* billing_address_city
* billing_address_zip
* billing_phone
* billing_email

The implementation is ported from Respa but the fields are different.

Later, `ReservationUnit` will have a foreign key referring to a `ReservationMetadataSet`. This will be added in a separate PR.


reservations/models.py
```diff
@@ -16,6 +18,36 @@
 User = get_user_model()
 
 
+RESERVATION_FORM_FIELDS = (
```
nicobav, line 21: I see duplicate list is used in migration `0019_reservationmetadata.py`. 
I did not find any usage of this from this PR. Where is this list used?

nicobav (commented):
Looks good, one question before approval;

reservations/models.py
jmp: Good catch. This was originally used in the admin model for validating the `form_field` value. But I don't think that validation is needed because the user cannot add/modify/delete values metadata fields in Django admin, so I removed it. This `RESERVATION_FORM_FIELDS` can actually be removed now.

jmp:
PR updated with `reservee_address_city` and `billing_address_city` changed to strings, and unnecessary `RESERVATION_FORM_FIELDS` removed.

nicobav (approved):

PR #372, opened 2021-12-15
TILA-1046: Validate reservation unit's price unit and interval fields

jmp:
Added validation for enum fields:

* `ReservationUnit.price_unit`
* `ReservationUnit.reservation_start_interval`

I was only able to find these two fields with missing validation. The others seemed to be read-only (`Reservation.state`) or had validation already (`Resource.location_type`).


nicobav (approved):
Nice :ok_hand:

PR #374, opened 2021-12-15
TILA-928 | Camelcase field names in errors

nicobav:
By default graphene is camelcasing, but we're missing some custom validation messages to be camelcased

TILA-928

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -1,4 +1,5 @@
 from django.core import validators
+from graphene.utils.str_converters import to_camel_case
 from rest_framework import serializers
 
 from api.graphql.base_serializers import (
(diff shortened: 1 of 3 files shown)


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -1,4 +1,5 @@
 from django.core import validators
+from graphene_django.utils.utils import _camelize_django_str
```
jmp, line 2: Would
```python
from graphene.utils.str_converters import to_camel_case
```
work here? It might be a bit better since it's a public API function.

nicobav: Nice found, quick look and it should be the one to use :+1:

nicobav: changed this :lotus_position_man:

jmp (approved):
Nice 😎

PR #375, opened 2021-12-15
TILA-1052 | Change resunit's surface to decimal fields in gql api

nicobav:
TILA-1052


jmp (approved):
I think this better than having three types. I can only think of two other options:

1. Use `graphene.Float` to have a single type for both mutations and queries, at the cost of possible floating point problems. 😞
2. Somehow integrate [this fix](https://github.com/graphql-python/graphene-django/pull/1277/files#diff-664cb060a42ef59d3d41cb12270525b5f9a03a4cc65e6f13ce79b1e139642e6bR117-R119) to our core until it has been fixed in Graphene-Django.

PR #377, opened 2021-12-16
TILA-1053: Expose cities in the GraphQL API

jmp:
Adds cities to the GraphQL API, so they can be queried like this:

```graphql
{
  cities {
    edges {
      node {
        name
      }
    }
  }
}
```

This is needed so that the frontend can show the city options in dropdown menus.

TILA-1053


nicobav (approved):
:cityscape:

PR #381, opened 2021-12-20
TILA-996 | REST api purposes refactor

nicobav:
Refactor REST api to use `ReservationPurpose` instead of `Purpose`. 
Added m2m relationship to `ReservationUnit` since that were also before with the "wrong" purpose.
Clients which are using the api should not need any changes since the fields/references named as `purpose` published in REST api do not change => this just refactor it to be pointing to correct model (which has same field structure).

TILA-996


jmp (approved):
Looks good. 🙂 (edit: it was supposed to be a smiling face)

jmp (approved):

PR #382, opened 2021-12-21
TILA-1055 | Use highest buffer time in reservation creation validation

nicobav:
When checking buffer times in reservation creation validation we must
use the highest buffer time when checking the reservation overlaps with
buffers included

TILA-1055


jmp (approved):
Looks good to me! 😄

PR #383, opened 2021-12-21
TILA-1057: Add validation for required metadata fields

jmp:
This adds validation for required metadata fields.

When mutating a reservation, we go through each required field in the attached reservation unit's metadata set. If any of the required field is missing/empty, a validation error is raised.

TILA-1057

### DIFF
--- a/api/graphql/reservations/reservation_serializers.py
+++ b/api/graphql/reservations/reservation_serializers.py
@@ -3,6 +3,7 @@ from typing import List
 
 from django.conf import settings
 from django.utils.timezone import get_default_timezone
+from graphene.utils.str_converters import to_camel_case
 from rest_framework import serializers
 
 from api.graphql.base_serializers import (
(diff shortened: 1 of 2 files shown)


api/graphql/reservations/reservation_serializers.py
```diff
@@ -295,6 +296,15 @@ def check_reservation_start_time(self, begin, scheduler):
                 f"Reservation start time does not match the allowed interval of {interval_minutes} minutes."
             )
 
+    def check_metadata_fields(self, data, reservation_unit) -> None:
+        metadata_set = reservation_unit.metadata_set
+        required_fields = metadata_set.required_fields.all() if metadata_set else []
+        for required_field in required_fields:
+            if not data.get(required_field.field_name):
+                raise serializers.ValidationError(
+                    f"Value for required field {required_field} is missing."
```
nicobav, line 305: I assume the `ReservationMetadataField` `field_name` is not camel cased? If so this should  use graphene's `to_camel_case` function.

```
from graphene.utils.str_converters import to_camel_case
.....
f"Value for required field {to_camel_case(required_field)} is missing."
```

nicobav (changes requested):
Looks good, one thing to consider

api/graphql/reservations/reservation_serializers.py
jmp: Good point, also I forgot to use `required_field.field_name` instead of just `required_field`.

nicobav (approved):

PR #385, opened 2021-12-22
TILA-1008: Expose max reservations per user in the GraphQL API

jmp:
Adds `maxReservationsPerUser` for reservation units in the GraphQL API.

This allows the `maxReservationsPerUser` to be queried and also mutated in the create and update mutations.


nicobav (approved):
gr8 :+1:

PR #386, opened 2021-12-22
TILA-1009: Validation for max reservations per user

jmp:
Adds validation for `max_reservations_per_user`.

The validation logic is the same as Respa's:

* We exclude canceled and denied reservations
* We exclude reservations whose end time is in the past
* The remaining reservations for the given user are counted and compared against the max reservations per user for the given reservation unit

### DIFF
--- a/reservations/models.py
+++ b/reservations/models.py
@@ -1,6 +1,7 @@
 from django.contrib.auth import get_user_model
 from django.db import models
 from django.db.models import F, Sum
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from applications.models import (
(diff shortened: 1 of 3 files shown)


reservations/models.py
```diff
@@ -150,6 +151,12 @@ def handling_required(self):
             recurring_reservation=None,
         )
 
+    def current(self):
```
nicobav, line 154: can't figure out the meaning of this filter by its name => the effect of this is quite like the `going_to_occur`

jmp: Good point, this filter is actually the same as `going_to_occur`. I'll remove the duplication.

nicobav (approved):
Looks good :+1:

PR #389, opened 2022-01-03
TILA-1046: Allow uppercase enum values

jmp:
This adds a new `ChoiceCharField` which can be used for enum fields with choices.

It takes the choices as an argument, and includes a validator which makes sure a given value is part of the choices. So no separate validations should be needed in mutation serializers, for example.

This means that:

* In queries, enum strings are returned in `UPPERCASE` (as before)
* In mutation inputs, enum strings can be given in `UPPERCASE` or `lowercase` (previously they had to be given in lowercase)
* In mutation payloads, enum strings are returned in `UPPERCASE` (previously they were returned in lowercase)

Maybe we will get true enums [some day](https://github.com/graphql-python/graphene-django/issues/1280).

### DIFF
--- a/api/graphql/choice_char_field.py
+++ b/api/graphql/choice_char_field.py
@@ -1,0 +1,42 @@
+from django.core.exceptions import ValidationError
+from django.utils.deconstruct import deconstructible
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
+
+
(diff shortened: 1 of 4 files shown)


nicobav (approved):
Cool :sunglasses: 

I like the common class approach :+1:

PR #392, opened 2022-01-05
TILA-1033: Prune inactive reservations

jmp:
This PR adds a the functionality for pruning reservations that have not been confirmed and are older than about 20 minutes.

The implementation has two parts:

1. Function that deletes all inactive reservations that are older than 20 minutes
2. A scheduled task that runs every 10 minutes
   * This means that old reservations will take 20-30 minutes to get pruned.

The scheduled task is done using [Celery's periodic tasks](https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html), but it can be replaced with a management command + cron job if needed.


reservations/models.py
```diff
@@ -340,6 +351,9 @@ class Reservation(models.Model):
         verbose_name=_("Details for this reservation's cancellation"), blank=True
     )
 
+    created_at = models.DateTimeField(
+        verbose_name=_("Created at"), null=True, default=timezone.now
```
nicobav, line 355: `timezone.now` is a function object => it does not call it but rather returns the function. So changing it to be like `default=timezone.now()` would do the trick but I like to use django's `auto_now_add` so instead of giving default value we say: `auto_now_add=True` would be nice.

But that said, I'm thinking that does it use the default timezone :thinking:

nicobav: oh, checking this bit further and it seems that the `default=timezone.now` should work
So django must know it is a callable and calls when necessary.

nicobav (approved):
Looks good :+1: 

I had a bit of a confusion there in the comments but all solid and ready to :rocket:

reservations/models.py
jmp: I'll try to explain my reasoning for choosing `default=timezone.now` here:

I originally used `auto_now_add`, but the problem with it is that the field is not as easily testable. For example, you can't do `ReservationFactory(created_at=...)` because whatever you pass to the factory will always be overwritten by the `auto_now_add` with the current time. Then you have to start mocking the dates with `freezegun` which is not so neat.

With `default=timezone.now` we can just pass whatever time we want to the factory and testing becomes a lot easier. 🙂 So I went for the option that facilitates more convenient testing.

PR #396, opened 2022-01-11
TILA-1057: Fix metadata field validation in confirm mutation

jmp:
Fixes the `confirmReservation` mutation so that metadata fields are validated correctly based on possibly existing values in the reservation instance.

Added a test case to make sure confirming a reservation (without providing the required fields in the mutation input) will work in the future too.

TILA-1057

### DIFF
--- a/api/graphql/reservations/reservation_serializers.py
+++ b/api/graphql/reservations/reservation_serializers.py
@@ -380,7 +380,8 @@ class ReservationUpdateSerializer(
         required_fields = metadata_set.required_fields.all() if metadata_set else []
         for required_field in required_fields:
             internal_field_name = required_field.field_name
-            if not data.get(internal_field_name):
+            existing_value = getattr(self.instance, internal_field_name, None)
+            if not data.get(internal_field_name, existing_value):
                 raise serializers.ValidationError(
                     f"Value for required field {to_camel_case(internal_field_name)} is missing."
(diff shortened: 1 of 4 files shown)


api/graphql/reservations/reservation_serializers.py
```diff
@@ -401,6 +401,9 @@ def validated_data(self):
         validated_data["state"] = STATE_CHOICES.CONFIRMED
         return validated_data
 
+    def check_metadata_fields(self, data, reservation_unit) -> None:
```
nicobav, line 404: What happens if the steps has been like: create (without metadata) => confirm ? Will it pass confirm - even if should not?

nicobav (approved):

api/graphql/reservations/reservation_serializers.py
jmp: Very good point! I didn't think of that scenario. 😖 In that scenario the confirmation would succeed happily, which is not what we want.

With that in mind, I think it's better to do the validation also in `confirmReservation`, but use the existing reservation data as the fallback. I believe that should cover that scenario.

jmp: Fixed this and added a new test case to cover the scenario you described. 🙂

nicobav (approved):

PR #397, opened 2022-01-12
TILA-1061 | Add is_visible filter to reservation unit filterset

nicobav:
This filter either filters in or out the reservation units that has
published (publish_begins & publish_ends)

TILA-1061


jmp (approved):
Looks good. 🙂

PR #398, opened 2022-01-12
TILA-1063 | ReservationUnitImage update mutation

nicobav:
Adds update mutation for `ReservationUnitImage` - the mutation allows updating the reservation unit image's type.

Also add `pk` to `ReservationUnitImageType` to make deletion and update possible.

TILA-1058 TILA-1063


jmp (approved):
Looks good!

PR #400, opened 2022-01-12
TILA-1071 | Reservation Deny reason

nicobav:
Adds model for reason for denying the reservation like the `ReservationCancelReason`. This is used by staff users to fill in predefined reason for some reservations that requires handling.

TILA-1071

changes for handle mutations in separate PRs


reservations/tests/factories.py
```diff
@@ -14,6 +14,11 @@ class Meta:
         model = "reservations.ReservationCancelReason"
 
 
+class ReservationDenyReasonFactory(DjangoModelFactory):
+    class Meta:
+        model = "reservations.ReservationCancelReason"
```
jmp, line 19: Should be `ReservationDenyReason`.

jmp (changes requested):

reservations/tests/factories.py
nicobav: Good catch. I actually fixed this one in the draft PR but it should be here :woozy_face:

nicobav:
fixed those copypaste errors

jmp (approved):
Looks good now! 👍

PR #401, opened 2022-01-12
TILA-1072 | Refactor handle api to deny api

nicobav:
We change the handle api to deny api since approving and denying specs vary that much that it makes sense to separate the apis

TILA-1072


jmp (approved):
Good idea which makes separation of concerns better. 👍 Great!

PR #403, opened 2022-01-13
Add read and write permissions to broker subdirectories

jmp:
Celery queue is not working because of file permissions:

```
PermissionError
[Errno 13] Permission denied: '/broker/queue/3259677_fc2859e1-ea07-4661-8c02-40d6382431d2.celery.msg'
```

The problem seems to be that the Celery worker process can't write files to `/broker/queue`.

Using OpenShift terminal, I checked that adding read/write permissions for the group *should* fix it. But it's a bit difficult to test locally. So I'm not 100% sure if this solution works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,9 +71,9 @@ RUN python manage.py collectstatic --noinput
 RUN chgrp -R 0 /tvp
 RUN chmod g=u -R /tvp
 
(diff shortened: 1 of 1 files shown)


nicobav (approved):
Let's try this :+1:

PR #404, opened 2022-01-14
Upgrade Pillow

jmp:
Bumps the Pillow version which is a dependency of `easy-thumbnails`.

Fixes Dependabot alerts.


nicobav (approved):
Yes please :+1:

PR #405, opened 2022-01-14
TILA-1075 | New reservation state handling required

nicobav:
Adds new reservation state `HANDLING_REQUIRED` which indicates that the reservation is to be handled.
When reservation is confirmed through confirm mutation we check that if it has `metadata_set` then it needs to be handled manually and the state is `HANDLING_REQUIRED`

TILA-1075


jmp (approved):
Looks good! 👍 Always nice to see more lines removed than added.

PR #406, opened 2022-01-14
TILA-1076 | Add reservation approve mutation

nicobav:
The mutation will be used for approving reservations that have state REQUIRES_HANDLING

TILA-1076


jmp (approved):
👍

PR #410, opened 2022-01-18
TILA-1105 | Add filter for getting reservation required handling

nicobav:
We need to show reservations which have been in state REQUIRES_HANDLING
or are => so every "requested" or reservations needed to be handled are
shown with this.

TILA-1105


nicobav:
There are open discussion about naming the filter. It's hard.

jmp (approved):

PR #412, opened 2022-01-20
TILA-1074 | Add custom connection class to Reservation for getting count

nicobav:
TILA-1074


jmp (approved):
Nice and simple! 🙂 I like it.

PR #414, opened 2022-01-24
Allow null for certain reservation unit fields

jmp:
Allows `null` to be passed for fields that may return `null` values in the API.

TILA-1065

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -91,17 +91,17 @@ class PurposeUpdateSerializer(PrimaryKeyUpdateSerializer, PurposeCreateSerialize
 
 
 class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySerializer):
-    terms_of_use_fi = serializers.CharField(required=False)
-    terms_of_use_sv = serializers.CharField(required=False)
-    terms_of_use_en = serializers.CharField(required=False)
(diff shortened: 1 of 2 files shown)


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -91,48 +91,59 @@ class Meta(PurposeCreateSerializer.Meta):
...
+    max_reservation_duration = DurationField(required=False, allow_null=True)
+    min_reservation_duration = DurationField(required=False, allow_null=True)
+    buffer_time_before = DurationField(required=False, allow_null=True)
+    buffer_time_after = DurationField(required=False, allow_null=True)
+    max_persons = serializers.IntegerField(required=False, allow_null=True)
     space_pks = serializers.ListField(
         child=IntegerPrimaryKeyField(queryset=Space.objects.all()),
         source="spaces",
         required=False,
+        allow_null=False,
```
nicobav, line 109: Should we allow nulls to here? and also to `resource_pks`, `purpose_pks`, `equipment_pks`,  `service_pks`

jmp: Good question. I double checked this -- we don't want to allow `null` here.

Since the value is returned as an empty list (not `null`) in queries, we also don't need to allow setting it to `null`.

However, the explicit `allow_null=False,` was left here by mistake. 😳 I removed it from this and the other fields.

nicobav (approved):

nicobav:
Looks good :+1: :shipit:

PR #416, opened 2022-01-26
TILA-1136: Add SKU to the reservation unit and reservation data models

jmp:
Adds a new `sku` field for reservation units and reservations.

The SKU ("tiliöintitiedot") will be modifiable only in Django admin. It's a string that looks something like `XXXXXX__YYYYYYYYYY___ZZ_WWWWWWWW` and will be used with payments.

When a new reservation is made, the SKU should be copied from the reservation unit to the reservation instance. This part will be done in a separate PR.

TILA-1136


nicobav (approved):
Alrighty

PR #417, opened 2022-01-26
TILA-1137: Copy SKU from reservation unit to reservation

jmp:
When a new reservation is created, we want to copy the SKU from the reservation unit to the reservation instance.

The SKU will be used later related to payments.

TILA-1137

### DIFF
--- a/api/graphql/reservations/reservation_serializers.py
+++ b/api/graphql/reservations/reservation_serializers.py
@@ -160,6 +160,7 @@ class ReservationCreateSerializer(PrimaryKeySerializer):
         if hasattr(reservation_units, "all"):
             reservation_units = reservation_units.all()
 
+        sku = None
         for reservation_unit in reservation_units:
             if (
                 reservation_unit.reservation_begins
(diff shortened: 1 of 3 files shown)


api/graphql/reservations/reservation_serializers.py
```diff
@@ -215,6 +215,7 @@ def validate(self, data):
             self.check_max_reservations_per_user(
                 self.context.get("request").user, reservation_unit
             )
+            data["sku"] = reservation_unit.sku
```
nicobav, line 218: Perhaps we should add a check that there aren't more `sku`s than one in the reservation's reservation units

nicobav (changes requested):

api/graphql/reservations/reservation_serializers.py
jmp: Good idea. I was trying to think what would happen if there is more than one reservation unit, but couldn't decide on the right approach. Maybe raising a validation error is the way to go here.

jmp: I added validation for this, and also a test case that makes sure it works. This required some changes to the mock opening hours to allow for multiple reservation units.

nicobav (approved):
Awesome :sunglasses:

PR #418, opened 2022-01-27
TILA-1129 | Copy handling details on approva and deny to working memo

nicobav:
This is the way.
For now.

TILA-1129


jmp (approved):
Looks good. 👍

PR #419, opened 2022-01-27
TILA-1131 | Add mutation for changing reservation status to requires handling

nicobav:
This is for the reservations which has status of CONFIRMED or DENIED

TILA-1131


nicobav:
oopsies, wrong refs

PR #420, opened 2022-01-27
TILA-1128 | Add mutation for changing reservation status to requires handling

nicobav:
This is for the reservations which has status of CONFIRMED or DENIED

TILA-1128


jmp (approved):
Looks good to me. 🙂

PR #421, opened 2022-01-27
TILA-1147 | Add field for require_reservation_handling to res unit

nicobav:
Based on this field we can determine is the reservation of this reservation unit something that needs handling (separate PR)

TILA-1147


jmp (approved):
🚀

PR #423, opened 2022-01-28
Refresh RedHat subscriptions in Dockerfile

jmp:
An attempt to fix the `403 Unauthorized` errors with some RedHat repos.

Based on various RedHat discussions where refreshing the subscriptions has worked.


nicobav (approved):
:crystal_ball:

PR #424, opened 2022-01-28
TILA-1148 | require_reservation_handling usage in apis

nicobav:
Adds functionality to add/change/read the `require_reservation_handling` boolean value through the api.
Change confirm serializer to check the boolean value instead of the metadata_set too determine is the reservation state to be REQUIRES_HANDLING (or not)

TILA-1148


jmp (approved):
👍

PR #426, opened 2022-01-31
TILA-1154: Add merchant ID to service sectors

jmp:
Adds a new "Merchant ID" field to service sectors.

The merchant ID is the `namespace` (a UUID) that identifies a merchant in the _Kaupunkiyhteinen verkkokauppa_.

The field can be edited via Django Admin.

TILA-1154


jmp:
Making this draft while waiting for confirmation that service sectors are the right place for this.

PR #427, opened 2022-01-31
TILA-1138: Add `authentication` field to reservation unit data model

jmp:
Adds the new `authentication` field in the `ReservationUnit` model.

Possible values:

* weak
* strong

I will create a separate PR for adding the field to the GraphQL API.

TILA-1138


jmp:
Removed the `none` option from the choices, since no actual need was identified for it.

nicobav (approved):
Nice :+1:

PR #431, opened 2022-02-03
TILA-1164 | Change SpaceType surfaceArea to be float

nicobav:
TILA-1164

### DIFF
--- a/api/graphql/tests/test_spaces.py
+++ b/api/graphql/tests/test_spaces.py
@@ -1,9 +1,11 @@
 import json
 
+import snapshottest
 from assertpy import assert_that
 from django.contrib.auth import get_user_model
 from graphene_django.utils import GraphQLTestCase
 
+from api.graphql.tests.base import GrapheneTestCaseBase
 from applications.tests.factories import ApplicationRoundFactory
 from permissions.models import GeneralRole, GeneralRoleChoice
 from reservation_units.tests.factories import ReservationUnitFactory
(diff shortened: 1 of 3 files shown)


api/graphql/tests/test_spaces.py
```diff
@@ -219,3 +232,34 @@ def test_regular_user_cannot_update(self):
         assert_that(content.get("errors")).is_not_none()
         self.space.refresh_from_db()
         assert_that(self.space.name_fi).is_equal_to("Space1")
+
+
+class SpacesQueryTestCase(GrapheneTestCaseBase, snapshottest.TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+        cls.space = SpaceFactory(surface_area=40.5)
```
jmp, line 241: Let's add some constant `name` for the space so that it will stay the same through separate test runs. 🙂

jmp (changes requested):

api/graphql/tests/test_spaces.py
nicobav: that's fixed now

api/graphql/tests/test_spaces.py
```diff
@@ -219,3 +232,34 @@ def test_regular_user_cannot_update(self):
...
+            """
+            query {
+                spaces {
+                    edges {
+                        node {
+                            nameFi
+                            surfaceArea
+                            code
+                            maxPersons
+                            code
```
jmp, line 256: We can probably remove this extra `code`. 🙂

jmp (approved):
Looks good, just one minor comment.

PR #435, opened 2022-02-10
TILA-1199: Remove `elastic-apm` dependency

jmp:
This PR removes the `elastic-apm` dependency from the project.

TILA-1199


nicobav (approved):
by apm :wave:

PR #436, opened 2022-02-11
TILA-1247 | Space permissions for Unit and Service sector

nicobav:
Adds some permissions for spaces in Unit and Service sector level

TILA-1247


jmp (approved):
Looks good. 👍 Nice and clean.

PR #437, opened 2022-02-11
TILA-972 | Send reservation confirmation email

nicobav:
Adds reservation email notification sending ability. 
New django app: `email_notification` which contains models for email templates and handling of sending emails

TILA-972


api/graphql/reservations/reservation_serializers.py
```diff
@@ -29,6 +31,14 @@
 
 DEFAULT_TIMEZONE = get_default_timezone()
 
+RESERVATION_STATE_EMAIL_TYPE_MAP = {
+    STATE_CHOICES.CONFIRMED: EmailType.RESERVATION_CONFIRMED,
+    STATE_CHOICES.REQUIRES_HANDLING: EmailType.HANDLING_REQUIRED_RESERVATION,
+    STATE_CHOICES.CANCELLED: EmailType.RESERVATION_CANCELLED,
+    STATE_CHOICES.DENIED: EmailType.RESERVATION_REJECTED,
+    "APPROVED": EmailType.RESERVATION_HANDLED_AND_CONFIRMED,
```
jmp, line 39: The asymmetry here is a little bit bugging me. But maybe it is okay. 🙂

The only alternative I can think of is removing ´"APPROVED"` from this dictionary and just using `EmailType.RESERVATION_HANDLED_AND_CONFIRMED` directly in `ReservationApproveSerializer.save`. But I'm not sure if it's any better. 🤷‍♂️

jmp (approved):
It looks good. 🙂 You'll have to rebase though thanks to my merge... 🙇‍♂️

api/graphql/reservations/reservation_serializers.py
nicobav: yup, I'm not certainly liking this either. I could remove the "APPROVED" and make the necessary changes in approve serializer but then `RESERVATION_STATE_EMAIL_TYPE_MAP` would not be fulfilling all cases :woozy_face: But should it have to?  Well...

PR #439, opened 2022-02-11
TILA-1260 | Django admin role searches and filters

nicobav:
Adds user to be searchable from roles
Adds user to be using select2 widget when creating a role (for user)

TILA-1260


jmp (approved):
Looks good. 🙂

PR #440, opened 2022-02-16
TILA-1257 | Unit importer to django admin

nicobav:
Add unit importer to django admin to UnitAdmin actions => so user can update unit information from tprek (to those where tprek_id is known)

TILA-1257

### DIFF
--- a/spaces/importers/units.py
+++ b/spaces/importers/units.py
@@ -80,11 +80,11 @@ class UnitImporter:
             self.field_map = field_map
 
         self.imported_unit_ids = []
+        self.creation_counter = 0
+        self.update_counter = 0
 
     @transaction.atomic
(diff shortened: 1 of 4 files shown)


spaces/importers/units.py
```diff
@@ -115,6 +115,10 @@ def _update_counters(self, created: bool):
 
         self.update_counter += 1
 
+    def _reset_counters(self):
```
Niilis, line 118: This function might be useful, but is not currently used anywhere. I would not add any unnecessary code that is not used anywhere just because it might be useful.

nicobav: Very true, I'll get rid of this

Niilis (approved):

PR #443, opened 2022-02-17
Tila-1248 | Organisation validation

Niilis:
Fix bug where empty email field on organisation causes failed save operation.

TILA-1248


nicobav (approved):
Very noice :+1:

PR #445, opened 2022-02-25
TILA-1095 | CSV export of applications

Niilis:
Add Django management command to export all applications into CSV file for manual handling.

TILA-1095

### DIFF
--- a/applications/management/commands/export_applications.py
+++ b/applications/management/commands/export_applications.py
@@ -1,0 +1,23 @@
+from typing import Any, Optional
+
+from django.core.management.base import BaseCommand
+
+from ...utils.export_data import ApplicationDataExporter
+
+
+class Command(BaseCommand):
+    help = "Exports application data from the database"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "application_round",
+            nargs="+",
+            type=int,
+            help="an integer or several representing the ID of the application round(s) to export",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> Optional[str]:
+        application_rounds = options.get("application_round")
+
+        for application_round in application_rounds:
+            ApplicationDataExporter.export_application_data(application_round)
--- a/applications/tests/test_application_export.py
+++ b/applications/tests/test_application_export.py
@@ -1,0 +1,401 @@
+import csv
+from datetime import datetime, time, timedelta
+from pathlib import Path
+
+import pytest
+from assertpy import assert_that
+from django.conf import settings
+from django.core.management import call_command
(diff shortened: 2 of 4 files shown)


applications/utils/export_data.py
```diff
@@ -0,0 +1,167 @@
...
+    @classmethod
+    def export_application_data(cls) -> None:
+        now = timezone.now()
+        root = Path(settings.BASE_DIR)
+        file_name = f"application_data_{now.strftime('%d-%m-%Y')}.csv"
+        path = root / "exports"
+        path.mkdir(parents=True, exist_ok=True)
+
+        application_events: List[ApplicationEvent] = list(
+            ApplicationEvent.objects.all()
```
nicobav, line 21: I think we need to get application events per application round. So we'll need to parametrize this `ApplicationDataExporter.export_application_data` OR set this be stateful and create an instance of this and set the application_round there and then call this `export_application_data` (and change it to instance method). 
I'm perhaps more towards the parameterized classmethod call  but you can decide.

applications/management/commands/export_applications.py
```diff
@@ -0,0 +1,12 @@
...
+
+from django.core.management.base import BaseCommand
+
+from ...utils.export_data import ApplicationDataExporter
+
+
+class Command(BaseCommand):
+    help = "Exports application data from the database"
+
+    def handle(self, *args: Any, **options: Any) -> Optional[str]:
```
nicobav, line 11: This also should have parametrization by application round

nicobav (changes requested):
Looking nice :+1: 
some tweaking with the querying the data to do :wrench:

applications/tests/test_application_export.py
```diff
@@ -0,0 +1,347 @@
...
+
+from ..utils.export_data import ApplicationDataExporter
+from .factories import (
+    ApplicationEventFactory,
+    ApplicationEventScheduleFactory,
+    EventReservationUnitFactory,
+)
+
+
+class ApplicationExport(TestCase):
```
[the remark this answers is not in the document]
Niilis: Class renamed to the second suggested name.

applications/utils/export_data.py
Niilis: I elected for parameterization as I believe this is the better option as well :smiley:

applications/management/commands/export_applications.py
Niilis: Parameterization implemeted to the command.

nicobav (approved):
Nice :+1: :rocket:

PR #446, opened 2022-02-25
TILA-1284 | Application additional information

nicobav:


applications/models.py
```diff
@@ -643,6 +643,14 @@ class Application(APPLICANT_TYPE_CONST, models.Model):
         null=True,
     )
 
+    additional_information = models.TextField(
+        verbose_name=_("Additional information"),
+        null=False,
+        blank=False,
```
Niilis, line 649: If this defaults to blank, but does not allow blank, can this cause problems in the future if some model is saved without changing the empty string for this field? This also means that this would be a required field, is this intentional?

Niilis (approved):
Looking good! :100:

PR #447, opened 2022-02-28
TILA-1311 | Application event schedule updates

Niilis:
* Default Application Event Schedule models' priority to `PRIORITY_HIGH`.
* Migrate all existing Application Event Schedule models' priority to `PRIORITY_HIGH`
* Use random priority in PriorityEventScheduleFactory

TILA-1311


Niilis:
It actually came to my mind (since I recently merged the application export) that the priority should probably be represented in the export as well? What do you think @nicobav ?

nicobav (approved):
:arrow_double_up:

PR #448, opened 2022-03-01
Upgrade pillow

nicobav:


nicobav:
Let's not bump the easythumbnails afterall since reportlab

nicobav:
Bump just pillow

Niilis (approved):
Header of this PR is a little outdated, but this is not the most important thing about this PR. :arrow_up_small:

PR #449, opened 2022-03-02
TILA-1315 | Application additional information

nicobav:
So here's the one with renamed branch => fixed migration etc.

Adds additional_information field to `Application` model as a optional field.

refs TILA-1315


Niilis (approved):
:arrow_double_up:

PR #451, opened 2022-03-03
TILA-1332 | Add unit filter to application rest end point

nicobav:
TILA-1332


Niilis (commented):
Code looks good but test seems to be failing. :thinking:

nicobav:
> Code looks good but test seems to be failing. thinking

Yes, there's some stupid mistake by me within in the test :disappointed:  I'll fixit

nicobav:
Fixed :magic_wand:

Niilis (approved):
All good now! :100:

PR #452, opened 2022-03-04
TILA-1325 | Improvements to CSV export of applications

Niilis:
Changes:

* New fields added to export:
* Multiple CSVs are now created, one for each priority (if customers has added entries with the priority)

TILA-1325


nicobav (approved):
Looks good :+1:

PR #453, opened 2022-03-07
TILA-1344 | UnitRole unit and unit_groups to m2m relationship

nicobav:
Makes `UnitRole`s unit and unit_group to be many to many relationship rather than foreign key relationship.

TILA-1344


permissions/helpers.py
```diff
@@ -72,14 +73,13 @@ def can_manage_service_sector_roles(user: User, service_sector: ServiceSector) -
...
+def can_manage_unit_roles(user: User, units: List[Unit]) -> bool:
     permission = "can_manage_unit_roles"
+    service_sectors = ServiceSector.objects.filter(units__in=units)
     return (
         is_superuser(user)
-        or has_service_sector_permission(
-            user, list(unit.service_sectors.all()), permission
-        )
-        or has_unit_permission(user, [unit], permission)
+        or has_service_sector_permission(user, list(service_sectors.all()), permission)
```
Niilis, line 81: This does not exactly concern this Pull request, but `has_service_sector` helper function can be optimized to perform 1 query less by having it accept a queryset for the second parameter instead of a list.

For example, instead of `list(service_sectors.all())` this would be `service_sectors.all().values_list("id", flat=True)`.

For cases where a single service sector is given, it would become `ServiceSector.objects.filter(id=service_sector.id).values_list("id", flat=True)`.

Don't know how many queries this will save, but piling up multiple calls to these permissions, I think it makes a difference.

permissions/helpers.py
```diff
@@ -252,12 +252,10 @@ def get_units_where_can_view_reservations(user: User) -> list:
     if has_general_permission(user, permission) or is_superuser(user):
         return list(Unit.objects.all())
 
-    return list(
-        map(
-            lambda role: role.unit,
-            user.unit_roles.filter(role__permissions__permission=permission),
-        )
-    )
+    units = user.unit_roles.filter(
```
Niilis, line 255: This is a great change! :+1: 

But, if we would have a type checker, this would not pass.

Function states that it will return a list, but this is a queryset. Can you update this to execute the query or update the return type to a union (`Union[list, queryset]` or whatever queryset should be annotated as :smile: )?

permissions/migrations/0015_unit_role_m2m_unit_and_groups.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 3.2.12 on 2022-03-07 10:09
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
Niilis, line 6: I am slightly concerned about this migration. This is a destructive operation and will remove all currently existing references to units and unit groups. Is this a problem?

If we want to preserve the existing units and unit groups, this migration must be written manually and I think it actually becomes a little tricky. Specially the reverse migration.

I can elaborate more if you do not catch what I am trying to say, but let's first establish if we do want to preserve data.

Niilis (commented):
Generally the changes are good, but the migration is slightly concerning. Let's discuss that first before I can approve the changes :smiley:

permissions/migrations/0015_unit_role_m2m_unit_and_groups.py
nicobav: Yup, good point - I'll write a custom migration

nicobav: And for backward reversible this is not. If this needs to be reversed we could do some manual handling for those.

Niilis: Yeah, to make it reversible it does require a decision of how to decide which unit/unit group should be preserved. Let's hope we do not have to reverse this migration. :sweat_smile:

nicobav: It actually could be done by creating multiple UnitRoles but still if we would unapply these it would be done manually or within some other migration that changes that back so its only logical to handle those cases then when know better.

permissions/helpers.py
nicobav: Good observation :mag:  Since this is not in scope and we've not yet encountered any performance issues, I would leave it as is. Also to notice that we have not yet been developing the permission edit UI nor the API (there is stub api existing for that but should be developed to gql side as new api)

nicobav:
Added custom db migration code to retain the units and unit_groups in UnitRole and modified the one return type to have QuerySet

permissions/migrations/0015_unit_role_m2m_unit_and_groups.py
Niilis: Yeah, you are right!

Changes good changes here. As this is not code that is constantly run in functions, I guess performance is not necessary. But, one thing to keep in mind. Since the `UnitRole` and `UnitGroup` is not used for grabbing data, I am fairly certain we do not actually have to fetch the objects from the database.

It would be enough to do this:

```pyton
role = UnitRole(id=role_id)
group = UnitGroup(id=unit_group_id)
role.unit_group.add(group)
```

Only the group update will hit the database. This is not necesary to change now.

Niilis (approved):
Changes are good to GO :100:

permissions/migrations/0015_unit_role_m2m_unit_and_groups.py
nicobav: Didn't know that you could add the m2m relations like that (with instantiating only id to model).
Actually I had whole objects in memory at first but when tested it failed and needed to those instances to be refetched. But yeah have check that way out later :+1:

PR #457, opened 2022-03-09
TILA-1352 | Application csv export period field

Niilis:
Correct application CSV export field data

Refs TILA-1352

### DIFF
--- a/applications/tests/test_application_export.py
+++ b/applications/tests/test_application_export.py
@@ -88,7 +88,7 @@ class ApplicationDataExporterTestCase(TestCase):
         if self.export_dir.is_dir():
             shutil.rmtree(self.export_dir)
 
-    def test_application_export__basic_case(self):
+    def test_basic_case(self):
         call_command("export_applications", self.application_round_id)
 
         event: ApplicationEvent = self.application_event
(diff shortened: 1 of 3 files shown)


applications/tests/test_application_export.py
```diff
@@ -797,6 +836,169 @@ def test_application_export__empty_purpose(self):
         assert_that(not_existing_file_medium.is_file()).is_false()
         assert_that(not_existing_file_low.is_file()).is_false()
 
+    def test_application_export__empty_event_begin_time(self):
```
nicobav, line 839: just noticed that most of these test methods has double underscore in name. No biggie since these are tests and so on, just had a need to say this :joy:

nicobav (approved):
Looks good :rocket:

applications/tests/test_application_export.py
Niilis: It is actually intentional and now I realize they are unnecessary. In previous projects, sometimes the double underscores are there to name tests in a recognizable name. This is especially useful when you have tests for different functionality in the same file.

I am not that fond of keeping tests of different functionalities in a single file, and we have not actually done this in this project. At least not in the tests I have written.

I can remove the double underscores and rename the tests without the "functionality" prefix for tests.

Niilis: Functionality prefix removed and double underscores at the same time :smile: .

PR #459, opened 2022-03-11
TILA-1296 | Change default time zone to "Europe/Helsinki"

nicobav:
Changes the default time zone to europe/helsinki

TILA-1296

### DIFF
--- a/api/graphql/opening_hours/opening_hours_types.py
+++ b/api/graphql/opening_hours/opening_hours_types.py
@@ -1,10 +1,25 @@
+import datetime
+
 import graphene
+import pytz
 from django.conf import settings
 from django.utils.timezone import get_default_timezone
 
 from opening_hours.utils.opening_hours_client import OpeningHoursClient
 
 DEFAULT_TIMEZONE = get_default_timezone()
+
+
+def get_time_as_utc(time: datetime.time, date: datetime.date, tz_info: pytz.timezone):
+    """Gets utc time from given time, date and timezone information.
+    Graphene seems to render time with offset info (e.g 10:00:00+00:00) for localized to UTC only.
+    So with e.g time zone being 'Europe/Helsinki' just don't work for some reason not yet known.
+    So for now we render all as UTC using this little helper.
+    """
+    start_dt = datetime.datetime.combine(date, time)
+    start_dt = tz_info.localize(start_dt)
+    start_dt = start_dt.astimezone(pytz.UTC)
+    return start_dt.timetz()
 
 
 class TimeSpanType(graphene.ObjectType):
@@ -24,16 +39,14 @@ class TimeSpanType(graphene.ObjectType):
         if not self.start_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        start = tzinfo.localize(self.start_time)
-
+        start = get_time_as_utc(self.start_time, datetime.date.today(), tzinfo)
         return start
 
     def resolve_end_time(self, info):
         if not self.end_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        end = tzinfo.localize(self.end_time)
-
+        end = get_time_as_utc(self.end_time, datetime.date.today(), tzinfo)
         return end
 
 
@@ -65,16 +78,14 @@ class OpeningTimesType(graphene.ObjectType):
         if not self.start_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        start = tzinfo.localize(self.start_time)
-
+        start = get_time_as_utc(self.start_time, self.date, tzinfo)
         return start
 
     def resolve_end_time(self, info):
         if not self.end_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        end = tzinfo.localize(self.end_time)
-
+        end = get_time_as_utc(self.end_time, self.date, tzinfo)
         return end
 
     def resolve_periods(self, info, **kwargs):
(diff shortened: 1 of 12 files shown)


api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -24,17 +27,19 @@ def resolve_start_time(self, info):
         if not self.start_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        start = tzinfo.localize(self.start_time)
-
-        return start
+        start_dt = datetime.datetime.combine(datetime.date.today(), self.start_time)
```
Niilis, line 30: I am having a hard time understanding what these three lines do. I understand they do timezone conversion, but for quickly figuring out what happens, some commenting would help.

It also seems that these three lines are duplicated in many places. If these lines would be extracted into a helper function, this would aid in understanding code, and further cases.

api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -24,17 +27,19 @@ def resolve_start_time(self, info):
...
+        return start_dt.timetz()
 
     def resolve_end_time(self, info):
         if not self.end_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        end = tzinfo.localize(self.end_time)
-
-        return end
+        end_dt = datetime.datetime.combine(datetime.date.today(), self.end_time)
```
Niilis, line 39: Duplicated logic here, see previous comment.

api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -65,17 +70,19 @@ def resolve_start_time(self, info):
         if not self.start_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        start = tzinfo.localize(self.start_time)
-
-        return start
+        start_dt = datetime.datetime.combine(self.date, self.start_time)
```
Niilis, line 73: Duplicated logic here, see previous comment.

api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -65,17 +70,19 @@ def resolve_start_time(self, info):
...
+        return start_dt.timetz()
 
     def resolve_end_time(self, info):
         if not self.end_time:
             return None
         tzinfo = self.start_time.tzinfo or DEFAULT_TIMEZONE
-        end = tzinfo.localize(self.end_time)
-
-        return end
+        end_dt = datetime.datetime.combine(self.date, self.end_time)
```
Niilis, line 82: Duplicated logic here, see previous comment.

api/graphql/reservations/reservation_serializers.py
```diff
@@ -402,8 +402,8 @@ def validate(self, data):
     def validated_data(self):
         validated_data = super().validated_data
         validated_data["user"] = self.instance.user  # Do not change the user.
-        validated_data["confirmed_at"] = datetime.datetime.now(
-            tz=get_default_timezone()
+        validated_data["confirmed_at"] = datetime.datetime.now().astimezone(
```
Niilis, line 405: Time zone helpers might be more readable here, up to preference :smile: .

[now()](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.timezone.now)
[localtime()](https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.timezone.localtime)

So: `localtime(now())`

api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -1,4 +1,7 @@
+import datetime
+
 import graphene
+import pytz
```
Niilis, line 4: I don't know if there is a lot we can do here, but I believe this is a good thing to be aware of.

Django will deprecate pytz in Django [v4 and remove support in v5](https://docs.djangoproject.com/en/4.0/releases/4.0/#zoneinfo-default-timezone-implementation), but we are still using python v3.8. The new standard will be available in python 3.9. So, we cannot start using the new [zoneinfo](https://docs.python.org/3/library/zoneinfo.html#module-zoneinfo) yet.

api/graphql/tests/test_reservation_units.py
```diff
@@ -96,8 +96,8 @@ def setUpTestData(cls):
             price_unit=ReservationUnit.PRICE_UNIT_PER_HOUR,
             is_draft=False,
             reservation_start_interval=ReservationUnit.RESERVATION_START_INTERVAL_30_MINUTES,
-            reservation_begins=datetime.datetime.now(),
-            reservation_ends=datetime.datetime.now(),
+            reservation_begins=datetime.datetime.now(tz=get_default_timezone()),
```
Niilis, line 99: Django timezone helpers might be a little more clean here. This is not important to change, but up to preference.

[now()](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.timezone.now)
[localtime()](https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.timezone.localtime)

So: `localtime(now())`.

Niilis (commented):
No big comments overall. Some code duplication and "up to preference" changes :smile: .

api/graphql/opening_hours/opening_hours_types.py
Niilis: For some reason this was in "another review" so had to post multiple reviews :thinking:

nicobav: yes, this was tricky and took me some time to fiddle with this part of the code since this will actually change the gql api to return always UTC opening hours since for some reason the timezone information was not rendered to the time value when it was set to `Europe/Helsinki` but only when it was pytz.UTC. Also why converting the time to datetime was that the localize method does not work for new timezone :exploding_head: 

Util function would be a good thing to be used here :+1:  I'll do the changes.

Niilis (approved):
Good to go :100:

PR #460, opened 2022-03-11
TILA-1363 | Display unit in reservation unit rest serializer

nicobav:
TILA-1363


Niilis (approved):
Good changes here :shipit:

PR #461, opened 2022-03-14
TILA-1355 | Fix CSV faulty write

Niilis:
Quote all data and change writing to excel style. Some data is difficult for default style and columns are not correct. This commit fixes the issues.

Refs TILA-1355


nicobav (approved):

nicobav:
Looks like not black enough

Niilis:
I must have an outdated black config locally since save should re-format my code.

Also, `black reservation_units/utils/` does not re-format code :thinking:

Niilis:
Ah, it's isort. Never mind, fixed.

PR #463, opened 2022-03-16
TILA-1250 | Allow post code to be longer

Niilis:
Post code should allow longer values as some countries use more characters
than Finland. These should be supported as well.

TILA-1250


nicobav (approved):
:rocket:

PR #465, opened 2022-03-18
TILA-1307 | Login session updates

Niilis:
- Shorter cookie age
- Secure cookie to only set cookie for https sessions

Refs TILA-1307


nicobav (approved):
:+1:

PR #466, opened 2022-03-21
TILA-1408 | Add user filter to Applications REST endpoint

nicobav:
Adds user filter to applications REST api for filtering applications by a user 

TILA-1408


Niilis (approved):
Changes good :+1:

PR #468, opened 2022-03-23
Set reservation begin dates to test_filter_reservation_state_accepts_multiple_values

nicobav:
The test was failing randomly due to the begin date being created to a random point of time.
Fix random fail by adding static begin times to reservations in the test


Niilis (approved):
Super! :100:

PR #469, opened 2022-03-23
TILA-1390 | User data anonymization capability to django admin

nicobav:
Adds anonymization functions to user application and reservation data and possibility to anonymize the data from django admin.

TILA-1390

### DIFF
--- a/tilavarauspalvelu/tests/test_anonymization.py
+++ b/tilavarauspalvelu/tests/test_anonymization.py
@@ -1,0 +1,197 @@
+from assertpy import assert_that
+from auditlog.models import LogEntry
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+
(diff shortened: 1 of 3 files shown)


Niilis (approved):
Very clean and readable code, well done!

If we do not have any other logs that needs cleaning, this branch is good to go. I mean, some access logs, session logs or other textfiles where emails I.E. are written? I could at least not find any `logger` that would log sensitive data.

nicobav:
> Very clean and readable code, well done!
> 
> If we do not have any other logs that needs cleaning, this branch is good to go. I mean, some access logs, session logs or other textfiles where emails I.E. are written? I could at least not find any `logger` that would log sensitive data.

Only what I can think of is sentry where there might be at least user email, but that data isn't kept very long period of time afaik.
But inside the app these are the places where the data is.

PR #470, opened 2022-03-25
TILA-1410 | Add application_round filter to res_unit gql query

nicobav:
TILA-1410


Niilis (approved):
Interesting that bandit complains about the random usage now but did not do it earlier when the code was merged.

Anyway, these changes are good to go :+1:

nicobav:
> Interesting that bandit complains about the random usage now but did not do it earlier when the code was merged.
> 
> Anyway, these changes are good to go +1

When I looked at it I think the github builds weren't run :open_mouth:  I'll fix those random stuff in separate PR

PR #471, opened 2022-03-28
Skipt bandit check for getting first and last name in anonymisation.

nicobav:
Also get rid off unused function for creating random text


Niilis:
Bandit still complain about some lines :thinking:

nicobav:
> Bandit still complain about some lines thinking
> 
> Is there some "skip this line" comment available for bandit?

I forgot the actual skip line comment :joy:

Niilis (approved):
Now everything is OK :shipit:

PR #475, opened 2022-04-01
TILA-1421 | Add GDPR api for deleting and fetching user data

nicobav:
Deleting here means anonymization

TILA-1421


api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+                    res.free_of_charge_reason,
+                    res.cancel_details,
+                ]
+            )
+        return reservations
+
+    @property
+    def applications(self) -> List:
+        applications = []
+        for app in self.user.application_set.all():
```
Niilis, line 61: Black does not enforce spacing, but in a few places here, I believe spacing makes code more readable. I'll comment on each line that I believe padding will help.

Here for example, an empty line before the `for` loop makes it more clear that a new section of code starts.

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+            applications.append(app.additional_information)
+            events = []
+            for e in app.application_events.all():
```
Niilis, line 64: Here again, an empty line before the for loop will help readability.

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+                events.append(e.name)
+                events.append(e.name_fi)
+                events.append(e.name_en)
+                events.append(e.name_sv)
+            applications.append({"events": events})
```
Niilis, line 69: Indentation in Python helps, but I believe this could have padding of empty line before and after this line.

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+                            app.organisation.name,
+                            app.organisation.identifier,
+                            app.organisation.email,
+                            app.organisation.core_business,
+                            app.organisation.core_business_fi,
+                            app.organisation.core_business_en,
+                            app.organisation.core_business_sv,
+                        ]
+                    }
+                )
```
Niilis, line 95: Empty line before next `if` could help here as well.

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+                            app.organisation.address.street_address_fi,
+                            app.organisation.address.street_address_en,
+                            app.organisation.address.street_address_sv,
+                            app.organisation.address.city,
+                            app.organisation.address.city_fi,
+                            app.organisation.address.city_en,
+                            app.organisation.address.city_sv,
+                        ]
+                    }
+                )
```
Niilis, line 111: Empty line before next `if` here as well could help.

These are very much up to preference and some might hate it that there is much whitespace in code.

However, even though there are a few bytes more in the file, it does make readability more enjoyable.

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return applications
+
+    @property
+    def email(self):
+        return self.user.email
+
+
+class TilavarauspalveluGDPRAPIView(GDPRAPIView):
```
Niilis, line 136: I wouldn't mind a little strategic padding in the methods within this class :pleading_face:

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+    def get_object(self) -> User:
+        user = get_object_or_404(User, uuid=self.kwargs["uuid"])
+        profile = ProfileUser(user)
+        # Permissio check???
```
Niilis, line 140: Not sure if this comment is necessary here. What is it's function?

api/gdpr.py
```diff
@@ -0,0 +1,165 @@
...
+        return profile
+
+    def _delete(self):
+        profile_user = self.get_object()
+        if not self._check_user_can_be_anonymized(profile_user.user):
+            raise DeletionNotAllowed()
+        anonymize_user_data(profile_user.user)
+
+    def _check_user_can_be_anonymized(self, user):
+        # TODO: Something to check with applications, are there any applications that needs to be handled still.
```
Niilis, line 151: Is this method complete? Will this be implemented at a later stage? :thinking:

api/tests/test_gdpr/gdpr_key.py
```diff
@@ -0,0 +1,68 @@
...
+    for name in ["n", "e"]:
+        value = key.public_key_jwk[name]
+        if isinstance(value, bytes):
+            key.public_key_jwk[name] = value.decode("utf-8")
+
+    return key
+
+
+rsa_key = _build_key(
+    """-----BEGIN PRIVATE KEY-----
```
Niilis, line 31: Ahaa! This is a test file. Good that we are not leaking any sensitive data.

api/tests/test_gdpr/test_gdpr_api.py
```diff
@@ -0,0 +1,159 @@
...
+            first_name="joe",
+            last_name="regular",
+            email="joe.regular@foo.com",
+        )
+        ReservationFactory(user=cls.user)
+        billing_address = AddressFactory()
+        ApplicationFactory(user=cls.user, billing_address=billing_address)
+        cls.url = reverse("gdpr_v1", kwargs={"uuid": cls.user.uuid})
+
+    def setUp(self) -> None:
```
Niilis, line 46: Ahaa! This resets the user data that might change in the tests?

Can you add a clarifying comment to reduce confusion in the future :pleading_face:

api/tests/test_gdpr/test_gdpr_api.py
```diff
@@ -0,0 +1,159 @@
...
+    def test_delete_user_returns_forbidden_when_using_wrong_scope(self, req_mock):
+        auth_header = self.get_auth_header(
+            self.user, ["testprefix.wrong_scope"], req_mock
+        )
+        self.client.credentials(HTTP_AUTHORIZATION=auth_header)
+        response = self.client.delete(self.url)
+        assert_that(response.status_code, 403)
+        self.user.refresh_from_db()
+        assert_that(self.user.username).does_not_contain("anonymized")
+
```
Niilis, line 144: Should a test be added that tries to anonymize a user, that is not currently logged in? So that users cannot start anonymizing everyone they want destroying functionality.

api/urls.py
```diff
@@ -125,4 +126,9 @@
 
 urlpatterns = [
     path("graphql/", csrf_exempt(FileUploadGraphQLView.as_view(graphiql=True))),
+    path(
+        "gdpr/v1/user/<str:uuid>/",
```
Niilis, line 130: :thinking: 

This is not super urgent but in the `tilavaraus/urls.py` this is imported as `from api.urls import urlpatterns as graphql` which is outdated after this update.

Niilis (approved):
Minor clarifications and some very much "up to preference" changes.

What would you say, we are starting to write more and more typed python. Should we include a typechecker I.E. `mypy` into the project? :thinking:

api/gdpr.py
nicobav: yes, it is reminder for later if we would like to do some checks that if the user data is not anonymizable => not yet specsed

api/tests/test_gdpr/gdpr_key.py
nicobav: Yes, it's a copy from [helsinki-profile-gdpr-api](https://github.com/City-of-Helsinki/helsinki-profile-gdpr-api/blob/main/tests/keys.py) tests. Parking permits seems to use the same as well.

api/tests/test_gdpr/test_gdpr_api.py
nicobav: Yes :+1:  It could also create the whole data here as well but the class method (setUpTestData) is better usually. 
`setUp` method is useful for "setupping" each test since `setUpTestData` is a class method which is called when the `TestCase` class is created => data changes between tests are possible => `setUp` is run per test.

api/urls.py
nicobav: oh good catch, I'll change the importing

api/gdpr.py
nicobav: it is dirt I'll clean it up

api/gdpr.py
nicobav: I partly agree with some spacing we could improve readability. Though I would see it too airy to put it like:
```
applications = []

for app in self.user.application_set.all():
        applications.append(app.additional_information)
```
since the `applications = []` is a variable associated with the for loop and thus make sense (to me) to wrap it firmly (:heart: ) with for loop since it makes "logical whole" (again _to me_) 

For those other spacings I'll answer below.

api/gdpr.py
nicobav: So referring to the other answer I partly agree here the spacing. I'll suggest following:
```
applications.append(app.additional_information)

events = []
for e in app.application_events.all():
    events.append(e.name)
```

since the `applications.append....` belongs to the above for loop and the `events = []` here is the sublogical whole with this for loop. I don't know if you can get into my brains with this thinking :brain:

api/gdpr.py
Niilis: Yeah, this is stuff that coders could argue about till the end of times!

I see your point here and will not argue that padding is a must here.

I see `applications` as a "global" variable within the function, since it is also the variable that is returned at the end of the function. The for loop is then the "builder" that assembles the output to the function. That is why I would like it stand-alone at the top of the function followed by padding before the for loop.

But, I see your point. This is getting more philosophical than necessary :smile:

api/gdpr.py
Niilis: Ah, I checked it out locally and this makes total sense! If you add a padding line after `applications.append` it is clear the events variable, the for loop and the append to applications is related.

PR #478, opened 2022-04-04
TILA-1219 | application rounds gql

Niilis:


api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -262,7 +262,7 @@ def resolve_category(self, info):
         return self.category
 
 
-class ApplicationRoundType(AuthNode, PrimaryKeyObjectType):
+class ReservationUnitApplicationRoundType(AuthNode, PrimaryKeyObjectType):
```
nicobav, line 265: I think we can get rid of this and use the new application round type.

api/graphql/application_rounds/application_round_types.py
```diff
@@ -0,0 +1,207 @@
...
+
+    approved_by = graphene.String()
+
+    applications_sent = graphene.Boolean()
+
+    class Meta:
+        model = ApplicationRound
+        fields = (
+            "id",
+            "name_fi",
```
nicobav, line 110: we have helper method for adding all translatable fields: `get_all_translatable_fields`
Example usage [here](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/478/files#diff-fda3743b409d6f4efdcea4942c95df1c6f607363ccb91ad90ef806a1ddffc44fR251).

nicobav (changes requested):
Looks good, some more of a cleaning up type of things to do :+1:

api/graphql/application_rounds/application_round_types.py
```diff
@@ -0,0 +1,207 @@
...
+    application_round_baskets = graphene.List(ApplicationRoundBasketType)
+    reservation_units = graphene.List(ReservationUnitType)
+    purposes = graphene.List(ReservationPurposeType)
+    service_sector = graphene.Field(ServiceSectorType)
+    status = graphene.Field(
+        graphene.Enum("status", ApplicationRoundStatus.STATUS_CHOICES)
+    )
+    status_timestamp = graphene.DateTime()
+
+    is_admin = graphene.Boolean()
```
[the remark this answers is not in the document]
Niilis: Removed as value is unnecessary.

api/graphql/application_rounds/application_round_types.py
Niilis: Used helper instead of manually added fields.

nicobav: oh, the link does not work but here's an example:
```    
class Meta:
  model = ModelOfTheType
  fields = [
      "field_1",
      "field_2",
  ] + get_all_translatable_fields(model)

File: api/graphql/reservation_units/reservation_unit_types.py (hunk 1 above)
Comment (Niilis) [reply api/graphql/reservation_units/reservation_unit_types.py]: Removed and functionality ported to new Type. (That functionality that did not exist yet)

**Review by nicobav — APPROVED** [review]
Great :rocket:

**Comment by nicobav** [comment]
Some snapshot fails

File: api/graphql/tests/test_application_rounds.py [hunk 4]
```diff
@@ -0,0 +1,108 @@
...
+                applicationRounds {
+                    totalCount
+                    edges {
+                        node {
+                            id
+                            nameFi
+                            nameEn
+                            nameSv
+                            reservationUnits {
+                                id
```
nicobav, line 65: Oh, missed these; Id fields tends to fail while using snapshots since the id field vary. Perhaps use some other field e.g name

api/graphql/tests/test_application_rounds.py
```diff
@@ -0,0 +1,108 @@
...
+                            }
+                            purposes {
+                                id
+                            }
+                            serviceSector {
+                                id
+                            }
+                            status
+                            statusTimestamp
```
nicobav, line 74: this will also fail unless its some statically typed value in tests.

api/graphql/tests/test_application_rounds.py
Niilis: Replaced with name when applicable.

api/graphql/tests/test_application_rounds.py
Niilis: This seems to be the same always and has not failed at all.

nicobav: its ok then :+1:

PR #479, opened 2022-04-05
TILA-1407 | Show applications to unit admins

nicobav:
Shows applications to unit admins for those who have permissions
"can_validate_applications" even though application would have events to
other units => right to see application is when there is at least one event
to the unit admin's unit. (ReservationUnit that belongs to the unit)

TILA-1407


Niilis (approved):
Tests and implementation look great, this is good to go :100:

PR #481, opened 2022-04-07
Bugfix | Adjust end-dates to never be same as start date

Niilis:
Tests can randomly fail when end date is same as start date. This commit fixes the issue to always have end date at least 1 day after start date.


nicobav (approved):
LGTM :+1:

PR #482, opened 2022-04-07
TILA-1472 | graphql filter reservation units by ids

Niilis:
Reservation units should be filterable by pk.

Refs TILA-1472

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_filtersets.py
+++ b/api/graphql/reservation_units/reservation_unit_filtersets.py
@@ -18,6 +18,9 @@ from spaces.models import Unit
 
 
 class ReservationUnitsFilterSet(django_filters.FilterSet):
+    pk = django_filters.ModelMultipleChoiceFilter(
+        field_name="pk", method="filter_by_pk", queryset=ReservationUnit.objects.all()
+    )
     unit = django_filters.ModelMultipleChoiceFilter(
         field_name="unit", queryset=Unit.objects.all()
     )
@@ -65,7 +68,7 @@ class ReservationUnitsFilterSet(django_filters.FilterSet):
 
     class Meta:
         model = ReservationUnit
-        fields = ["unit", "keyword_groups"]
+        fields = ["pk", "unit", "keyword_groups"]
 
     def get_text_search(self, qs, property, value: str):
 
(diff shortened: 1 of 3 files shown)


api/graphql/reservation_units/reservation_unit_filtersets.py
```diff
@@ -18,6 +19,7 @@
 
 
 class ReservationUnitsFilterSet(django_filters.FilterSet):
+    pk = NumberInFilter(field_name="pk", lookup_expr="in")
```
nicobav, line 22: Sadly the django-graphene has still some tails to fix with decimal fields so this will require (and accept) strings instead of  integer values which we'd like... Using `django_filters.ModelMultipleChoiceFilter` should work with proper `method` defined should accept list of integers though. So I suggest to change this to `ModelMultipleChoiceFilter`

nicobav (changes requested):
~Something off with formatting but after reformatting~  <-- browser cache :woozy_face:

api/graphql/reservation_units/reservation_unit_filtersets.py
Niilis: Documentation was incredibly poor for this type of filtering. But, if what you suggest works, that is definitely better! I'll have a look.

Niilis: The `ModelMultipleChoiceFilter` requires a queryset and is designed for related models. I did not get it to work by simply using that filter, adding a `field_name` and `method` kwarg.

How is this supposed to work since we filter on the model's own ID?

nicobav: in the `method` filter the `qs` by the value (which is queryset of reservationunit objects) should work

Niilis: I was perhaps not clear enough. The `ModelMultipleChoiceFilter` is throwing an error when it does not get the `queryset` kwarg given to it. There is no queryset to be given to that argument as we want to filter the model, not a related model.

nicobav: Let's discuss this in slack

Niilis: I found a solution but it feels wrong :thinking:

nicobav: Yes, its the friday wild code feeling :tm: . It's a bit of a hack but afaik _it is the way_  for getting integers to the list. 

Would need to dig deeper into graphene and get the decimals work in serializers etc. Actually we would need graphene to update their versions to newer. There's some PRs added for that but no releases yet

nicobav (approved):
:rocket:

PR #483, opened 2022-04-11
TILA-1407 | Show applications to unit group admins

nicobav:
Shows applications to unit group  admins for those who have permissions
"can_validate_applications" even though application would have events to
other units => right to see application is when there is at least one event
to the unit group admin's unit. (ReservationUnit that belongs to the
unit that belongs to the group)

TILA-1407

### DIFF
--- a/api/tests/test_application_api.py
+++ b/api/tests/test_application_api.py
@@ -752,10 +752,32 @@ def test_unit_admin_can_view_application(
 
 
 @pytest.mark.django_db
-def test_unit_admin_cant_view_application(
+def test_unit_group_admin_can_view_application(
+    unit_group_admin_api_client, application, event_reservation_unit
+):
+    response = unit_group_admin_api_client.get(
+        reverse("application-detail", kwargs={"pk": application.id}),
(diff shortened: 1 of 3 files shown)


api/tests/test_application_api.py
```diff
@@ -762,6 +773,17 @@ def test_unit_admin_cant_view_application(
     assert response.status_code == 404
 
 
+@pytest.mark.django_db
+def test_unit_group_admin_cant_view_application(
```
Niilis, line 777: To make this more clear, should this be renamed `test_unit_group_admin_cant_view_other_group_application`?

Now, one test says unit group admin should be able to view application and this test says unit group admin can't view application. This makes it more clear why the admin should not be able to see the application.

Niilis (approved):
Minor test rename suggestion but otherwise code looks good :+1:

api/tests/test_application_api.py
nicobav: changed this and other test naming a bit. added `_not_for_<unit/unitgroup)>`

PR #487, opened 2022-04-20
TILA-1488 | Application api endpoint slowness

Niilis:
* Permission and permission helper query performance updates
* `/v1/application/` endpoint query performance updates

Refs TILA-1488

### DIFF
--- a/permissions/api_permissions/drf_permissions.py
+++ b/permissions/api_permissions/drf_permissions.py
@@ -129,9 +129,9 @@ class GeneralRolePermission(permissions.BasePermission):
 class UnitRolePermission(permissions.BasePermission):
     def has_object_permission(self, request, view, unit_role):
         if unit_role.unit_group:
-            return can_manage_unit_group_roles(request.user, unit_role.unit_group)
+            return can_manage_unit_group_roles(request.user, [unit_role.unit_group.id])
         if unit_role.unit:
-            return can_manage_unit_roles(request.user, unit_role.unit)
+            return can_manage_unit_roles(request.user, [unit_role.unit.id])
         return False
 
     def has_permission(self, request, view):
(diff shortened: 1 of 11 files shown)


permissions/api_permissions/drf_permissions.py
```diff
@@ -254,33 +252,36 @@ def check_permissions_for_single(self, request, status, application_event_id):
...
-            service_sector = ServiceSector.objects.get(
-                applicationround=ApplicationRound.objects.get(
-                    applications=Application.objects.get(
+            service_sector = ServiceSector.objects.filter(
+                applicationround=ApplicationRound.objects.filter(
+                    applications=Application.objects.filter(
                         application_events=application_event_id
-                    )
-                )
+                    )[:1]
```
nicobav, line 259: What's the reason for slicing here? 
I see we're using `.first()` in lower section of the code for this QS, shouldn't that be fine without need to slice :thinking: 

Also seeing slicing used in other parts of the code too, so extra interested :nerd_face:

nicobav (approved):
Looks good :+1: 

Slicing raised a question, not a change request

permissions/api_permissions/drf_permissions.py
Niilis: Slicing is for the filter queries, that they will always return only 1 (even though we filter by IDs to foolproof the query).

This optimization to use `.filter()` instead of `.get()` is to keep these queries lazy as they will not necessarily be used therefore saving queries to the database.

Niilis: So, slicing is for the "subqueries" `.first()` is for the final call to database for a result.

nicobav: Yup, the filter is good way to go. The slicing might be unnecessary now that I think of it more but its fine

PR #488, opened 2022-04-20
TILA-1471 | Increase header block size

Niilis:
It seems block size is not large enough with the default. This commit double the header block size.

This can be because some requests having longer urls than others and only narrowly passing the threshold. We do not want to increase this too much [as it could be a security risk](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#buffer-size), but only slightly to suit our current needs.

Refs TILA-1471


nicobav (approved):
Let's try

PR #499, opened 2022-04-26
TILA-1446 | Add base Connection class which contains total count

nicobav:
TILA-1446


Niilis (approved):
Super nice with base class for this purpose :100:

PR #502, opened 2022-04-27
TILA-1485 | CSV export improvements

Niilis:
Minor "nice to have" fixes for CSV export.

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -53,6 +53,18 @@ def year_not_in_future(year: Optional[int]):
         raise ValidationError(format_lazy("{year} {msg}", year=year, msg=msg))
 
 
+class StatusMixin:
+    STATUS_CHOICES = ()
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


applications/models.py
```diff
@@ -53,6 +53,18 @@ def year_not_in_future(year: Optional[int]):
         raise ValidationError(format_lazy("{year} {msg}", year=year, msg=msg))
 
 
+class StatusBaseClass:
```
nicobav, line 56: I would leave `Class` from the name => `StatusBase` or would call it a mixin class => `StatusMixin` :thinking:

nicobav (changes requested):

applications/models.py
Niilis: That's a good point! I'll change it to `StatusMixin` as I feel that is more descriptive than Base.

nicobav (approved):
LGTM :rocket:

PR #504, opened 2022-04-29
Test improvements

Niilis:
Make tests more robust by hardening test cases and adding additional
checks.


nicobav (approved):
Awesome :beers:

PR #505, opened 2022-05-03
TILA-1517 | Add reservation unit and application counts to ApplicationRoundType for gql api

nicobav:
TILA-1517


Niilis (changes requested):
Code changes look good, but I think it would be good to add a test case that explicitly test that applications in `in_review` status are not counted.

Niilis (approved):
Ready to go :rocket:

PR #508, opened 2022-05-04
TILA-1558 | ReservationUnit ordering by rank

nicobav:
Adds to `ReservationUnit` model a custom order field called `rank` to be used to order graphql results in wanted order.

TILA-1558


api/graphql/tests/test_reservation_units.py
```diff
@@ -1389,6 +1413,31 @@ def test_order_by_name_and_unit_name_reversed(self):
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
 
+    def test_order_by_rank(self):
+        ReservationUnit.objects.exclude(id=self.reservation_unit.id).delete()
+        ReservationUnitFactory(rank=1)
```
Niilis, line 1418: Should these be crated in a "random" order. I.E. not in order as done now?

This is to make sure ordering is not correct because they were created in the correct order, rather they are ordered correctly because we asked for the order.

Niilis (approved):
One minor comment for testing, but changes look good to me :+1:

api/graphql/tests/test_reservation_units.py
nicobav: Good point, it certainly makes the test more error proof for future changes :+1: , I'll shuffle the creation order a bit.

But to proof, there is already one reservation unit which does not have rank at all the and it gets [last](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/508/files#diff-6ed23f26f26cfdcba0d88627d44f787980c5d75843ff4bc4ff419024d6d9fca6R910).

PR #510, opened 2022-05-10
TILA-1569 | Applications fetch in graphql

Niilis:
Applications endpoint was translated to GraphQL in this branch for better specification in data fetches. This will improve performance when less information is queried and gathered in each request when only the necessary data has to be fetched instead of all possible values.

* Translate `/v1/applications/` endpoint from REST to GraphQL.
* Move performance investigation tooling to documentation instead of ready-made settings.
* Improve code coverage pytest settings.

Refs TILA-1569

### DIFF
--- a/permissions/api_permissions/graphene_permissions.py
+++ b/permissions/api_permissions/graphene_permissions.py
@@ -4,6 +4,7 @@ from django.shortcuts import get_object_or_404
 from graphene import ResolveInfo
 from graphene_permissions.permissions import BasePermission
 
+from applications.models import Application
 from permissions.helpers import (
     can_create_reservation,
     can_handle_reservation,
@@ -19,6 +20,7 @@ from permissions.helpers import (
     can_manage_units_reservation_units,
     can_manage_units_spaces,
     can_modify_reservation,
+    can_read_application,
     can_view_recurring_reservation,
 )
 from reservation_units.models import ReservationUnit, ReservationUnitImage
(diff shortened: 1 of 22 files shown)


permissions/api_permissions/graphene_permissions.py
```diff
@@ -36,6 +38,56 @@ def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> b
...
+        application = Application.objects.filter(id=id)
+
+        if application:
+            return user.is_authenticated and can_read_application(user, application)
+
+        return False
+
+    @classmethod
+    def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> bool:
+        # request = info.context.request
```
nicobav, line 58: We could clean these commented lines and introduce them when needed.

tilavarauspalvelu/settings.py
```diff
@@ -518,3 +519,23 @@ def get_git_revision_hash() -> str:
     # NOTE: If you run this in docker,
     # here is additional information for set-up: https://django-debug-toolbar.readthedocs.io/en/latest/installation.html
     # right before "Troubleshooting"
+
+# GraphQL debug toolbar and django debug toolbar cannot be enabled at the same time
+USE_GRAPHQL_DEBUG_TOOLBAR = env("USE_GRAPHQL_DEBUG_TOOLBAR")
```
nicobav, line 524: I think these should be defined in local_settings and then this information how to conf these would be in README

permissions/api_permissions/graphene_permissions.py
```diff
@@ -36,6 +38,56 @@ def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> b
         return False
 
 
+class ApplicationPermission(BasePermission):
+    @classmethod
+    def has_permission(cls, info: ResolveInfo) -> bool:
+        return info.context.user.is_authenticated
```
nicobav, line 44: This is quite a ponder actually... I would say only staff users with proper permissions would be able to view applications. At least some of the details. But can't tell for certain now....

nicobav (changes requested):
Looks good!
Some small things to debate

tilavarauspalvelu/settings.py
Niilis: Hmm, I think you have a good point here! These affect performance significantly, and should only be used in development, so, I agree it would be a good idea to move configuration into the documentation. I'll remove the ENV variables and move this to documentation.

Is it too much of a hassle to install packages so that the debug packages would not be included in the requirements list?

permissions/api_permissions/graphene_permissions.py
Niilis: Hmm... The permission is modeled after the REST API permission. Also, the "main" filtering of which applications are included in the "applications" query, is defined in the `ApplicationsFilter` `resolve_queryset` method.

But, I do think it would be good to "doublecheck" that a user has permission to access some application! So, the permission can be improved to include more elaborate checking.

nicobav: yes, there might be some information that does not concern the public viewer.

tilavarauspalvelu/settings.py
nicobav: I don't think it's too much of a hassle. In future we could have separate requirements for dev as we discussed irl while back.

permissions/api_permissions/graphene_permissions.py
Niilis: I don't  know if this is what you meant, but I made the permission tests more robust. Unit-, unit group- and service sector permissions now create a new type of application that the tested user should not have access to. This should make tests more robust that users do not have access to applications they should not see.

I also added "regular user" test to the permissions, even though this is kind of tested in the feature tests. But, just to make sure that permissions/ filtering is working.

nicobav (approved):
GR8! :+1:

PR #511, opened 2022-05-13
TILA-1501 | Optimize REST API endpoints

Niilis:
Used database query optimization techniques in all REST endpoints to reduce amount of queries to the database. This should bring performance improvements for more snappy fetches.

Refs TILA-1501

### DIFF
--- a/api/application_round_api.py
+++ b/api/application_round_api.py
@@ -1,10 +1,14 @@
 import logging
 
 from django.conf import settings
-from django.db.models import Sum
+from django.db.models import Case, Exists, OuterRef, Prefetch, Subquery, Sum
+from django.db.models import Value as V
+from django.db.models import When
+from django.db.models.functions import Coalesce, Concat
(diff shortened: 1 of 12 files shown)


api/application_round_api.py
```diff
@@ -1,10 +1,14 @@
 import logging
 
 from django.conf import settings
-from django.db.models import Sum
+from django.db.models import Case, Exists, OuterRef, Prefetch, Subquery, Sum
+from django.db.models import Value as V
```
nicobav, line 5: Why `V`? I think `Value` is good as is.

nicobav (approved):

api/application_round_api.py
Niilis: This is what the Django documentation recommends. Have not putten much thought to why, but I would guess that it clashes with some very common library or function. For example in the `Concat` database function: https://docs.djangoproject.com/en/4.0/ref/models/database-functions/#concat

PR #513, opened 2022-05-18
TILA-1411 |  Equipment category rank

nicobav:
TILA-1411


Niilis (approved):
Code looks good and tests are sound. Well done! :ship:

PR #514, opened 2022-05-18
TILA-1610 | ReservationUnitType rank

nicobav:
to be used for sorting most popular types.

TILA-1610


Niilis (approved):
I was wondering why these show similar changes to the previous commits you have made. But I did realize that these depend on each other.

Anyway, these changes are good to go as well :100:

PR #516, opened 2022-05-19
TILA-1609 | Add order by to Applications

Niilis:
* Order by filter to applications GraphQL query
* Tests for the different ordering types

Refs TILA-1609

### DIFF
--- a/api/graphql/applications/application_filtersets.py
+++ b/api/graphql/applications/application_filtersets.py
@@ -1,3 +1,7 @@
+from django.db.models import Case, CharField, F
+from django.db.models import Value as V
+from django.db.models import When
+from django.db.models.functions import Concat
 from django_filters import rest_framework as filters
 
 from applications.models import Application, ApplicationRound, User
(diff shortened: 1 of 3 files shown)


Niilis:
Question @nicobav , As you can see in the filtering, I annotated the "applicant" field for filtering with a 3 part database Case clause. My question is, the current "applicant_*" fields uses the "user" field but this ordering is more elaborate. Should the "applicant_*" field be updated to be as elaborate as this order by?

What I mean is to select the applicant_* values with the same logic.

nicobav:
> Question @nicobav , As you can see in the filtering, I annotated the "applicant" field for filtering with a 3 part database Case clause. My question is, the current "applicant__" fields uses the "user" field but this ordering is more elaborate. Should the "applicant__" field be updated to be as elaborate as this order by?
> 
> What I mean is to select the applicant_* values with the same logic.

This looks awesome :+1:  
I would say that we leave the output data as is.

nicobav (approved):

PR #517, opened 2022-05-20
TILA-1621 | ReservationUnit min_persons

nicobav:
TILA-1621


saaste (approved):
Good stuff! 👏 

What is the proper way to update `tilavarauspalvelu_visualized.json`? ☺️

nicobav:
> What is the proper way to update `tilavarauspalvelu_visualized.json`? relaxed

Good question; there's a file called `fixture_graph_commands.txt` in root of the project. It contains two commands including the visualized.json update (the graph_models it is). This probably should be moved to `CONTRIBUTING.md` or `README.md` => the whole README should be re-organized more sane way => now the graph seems to be just there lying around without a context.

PR #518, opened 2022-05-20
TILA-1619 | ReservationUnit's reservation kind

nicobav:
the reservation_kind describes what kind of reservations is this  ReservationUnit used for; season or direct or both

TILA-1619


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -335,6 +336,18 @@ def validate_for_publish(self, data):
                 "Not draft reservation unit must have a reservation unit type."
             )
 
+        max_persons = data.get(
+            "max_persons", getattr(self.instance, "max_persons", None)
```
saaste, line 341: Django n00b question: could you clarify what happens here? What is the difference between `data.get("max_persons")` and `getattr(self.instance, "max_persons", None)`?

saaste (approved):
Looks good to me! Approved with a question 👍

api/graphql/reservation_units/reservation_unit_serializers.py
nicobav: The `data` contains the data that is coming from the request and the `self.instance` is the possible already existing (if it is an update) db instance of the `ReservationUnit` 
=> the `data` is a dictionary so it can be accessed using its `get` method (also this works: data["max_persons"], but it could throw a `KeyValueError` and should be handled) . and the second argument is the default value if the value does not exist at all.
=> lastly the `getattr` is for getting any object's attributes => getting the instance's `max_persons` value and defaults to None (if the instance would be None). 
Using this this way imo its more cleaner looking rather than using `if` `else` clauses. 

So shortly, first we try to get the input value of the request and then default to db instance value if it does not exist.

saaste: Got it! Now it makes sense. Thanks for the explanation! ☺️

PR #519, opened 2022-05-23
Add min_persons and reservation_kind to res_unit_by_pk type

nicobav:
I missed these fields from `ReservationUnitByPk` type.


nicobav:
Really should get ridoff this type in the future

saaste (approved):

PR #520, opened 2022-05-23
TILA-1620 | ReservationUnit apply for free of charge

nicobav:
Introduce field `can_apply_free_of_charge` to ReservationUnit to indicate whether it can be applied free of charge (if otherwise chargeable).

TILA-1620


saaste (approved):
The amount of lines changed scared me in the beginning. Turns out most of it comes from the generated stuff 😅 

Looks good to me! 👏

PR #521, opened 2022-05-24
TILA-1606 | Standardized graphql query optimization

Niilis:
Standardized utils function to optimize GraphQL queries dynamically.

Applied query optimizer for applications field as an example and testing field.

### DIFF
--- a/utils/query_performance.py
+++ b/utils/query_performance.py
@@ -1,0 +1,314 @@
+from typing import Any, Dict, List, Union
+
+import graphene
+from django.db.models import Prefetch, QuerySet
+
+
+class QueryPerformanceOptimizerMixin:
+    """Mixin for GraphQL database query optimization
+
+    Inherit from this class and add "QueryOptimization" to class.
+
+    QueryOptimization must include the following fields:
+        * field_name (str) = The name of the query field
+        * query_optimization (dict) = dict of field names mapped to optimizations
+
+    query_optimizations values must be 2-tuples with "optimization type" as first value
+    and "model field name" or a dict as second value.
+    Dict as second value is only supported for "prefetch_related" fields as it will generate a `Prefetch()` object.
+
+    Dict as optimization value must include following keys:
+        * field_name (str) = name of model field name
+        * base_queryset (QuerySet) = Django model queryset for the field_name
+        * child_optimizations (dict) = dict with deeper optimizations
+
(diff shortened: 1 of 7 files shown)


utils/query_performance.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # TODO: Currently, this optimization only supports queries with one node
+        # If multiple nodes are added to query, this implementation does not
+        # know how to optimize the n:th query!
+
+        # Start parsing GraphQL query structure
+        for selection in operation.selection_set.selections:
+            if selection.name.value == field_name:
+                query_selections = selection.selection_set.selections
+
```
nicobav, line 57: I'd get rid off the air here

utils/query_performance.py
```diff
@@ -0,0 +1,316 @@
...
+                break
+
+        # Query selections found, find edges
+        if query_selections:
+            for selection in query_selections:
+                if selection.name.value == "edges":
+                    edge_selections = selection.selection_set.selections
+
```
nicobav, line 65: I'd get rid off the air here

nicobav, line 65: I'd get rid off the air here

nicobav (changes requested):
Looks great :+1: 

Only some line breaks to get rid off :shrug:

nicobav (approved):
amazing!

saaste (approved):
Took me a while to understand all the bells and whistles and I still don't quite comprehend everything - mostly because of my tiny Django experience. But the code is easy to read and follow and in general it makes sense.

So, LGTM!

I wonder if something like this has been already done and provided as a package. This feels such a common problem that one would think that we can't been the only one struggling with query optimization 😄

Niilis:
> Took me a while to understand all the bells and whistles and I still don't quite comprehend everything - mostly because of my tiny Django experience. But the code is easy to read and follow and in general it makes sense.
> 
> So, LGTM!
> 
> I wonder if something like this has been already done and provided as a package. This feels such a common problem that one would think that we can't been the only one struggling with query optimization smile

@saaste there is this one [Graphene Django query optimizer](https://github.com/tfoxy/graphene-django-optimizer), but I did not get it to work on deeper levels than 1-2. Within this project, some structures are many levels more deep and will definitely clog servers with the right kind of a query. That is why this PR was created as this does recognize deeper levels and includes the necessary defined optimizations.

PR #523, opened 2022-05-25
TILA-1622 | ReservationUnit reservations max and min days

nicobav:
Adds fields to `ReservationUnit` which indicate the "time window" when reservations are possible. For e.g reservations must be done at least one (1) day before the reservation date => reservations_min_days_before.

Also adds a validation logic to reservation creation.

TILA-1622


saaste (approved):
Looks great to me! 👍

PR #524, opened 2022-05-25
TILA-1635: Allow reservations without opening hours

saaste:
The PR adds a new boolean flag to `ReservationUnit`. If the flag is set true, reservations can be done without opening hours.

I'm still lacking a lot of context so it is quite likely that not everything I did might make sense 😅  Take extra care when reviewing this PR and feel free to add as many comments as needed to address all the issues. I'll check them out on Monday ☺️


api/graphql/reservations/reservation_serializers.py
```diff
@@ -157,6 +157,31 @@ def validate_reservee_type(self, value):
             )
         return value
 
+    def validate_reservation(self, reservation_unit, scheduler, begin, end):
+        if reservation_unit.allow_reservations_without_opening_hours:
```
nicobav, line 161: returning here will mean we don't check `reservation_begins` dates and not check the overlap.

So I would remove this and add a condition in the `is_reservation_unit_open` like:
```
is_reservation_unit_open = reservation_unit.allow_reservations_without_opening_hours or scheduler.is_reservation_unit_open(begin, end)
```

But this thing should perhaps have its own method where no other than opening hour related would be checked (refs the above comment)

api/graphql/reservation_units/reservation_unit_serializers.py.orig
```diff
@@ -0,0 +1,429 @@
+from django.core import validators
```
nicobav, line 1: This file is probably some leftover from gitting and should get deleted

api/graphql/reservation_units/reservation_unit_types.py.orig
```diff
@@ -0,0 +1,526 @@
+import datetime
```
nicobav, line 1: This file is probably some leftover from gitting and should get deleted

api/graphql/tests/test_reservation_units.py.orig
```diff
@@ -0,0 +1,3388 @@
+import datetime
```
nicobav, line 1: This file is probably some leftover from gitting and should get deleted

reservation_units/models.py.orig
```diff
@@ -0,0 +1,715 @@
+import datetime
```
nicobav, line 1: This file is probably some leftover from gitting and should get deleted

tilavarauspalvelu_visualized.json.orig
```diff
@@ -0,0 +1,5 @@
+<<<<<<< HEAD
```
nicobav, line 1: This file is probably some leftover from gitting and should get deleted

api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -217,7 +217,7 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
     allow_reservations_without_opening_hours = serializers.BooleanField(
         required=False,
         default=False,
-        help_text="Allow booking without opening hours. Used for testing.",
+        help_text="Allow reservations without opening hours. Used for testing.",
```
nicobav, line 220: This rename commit should be rebased into the original commit where this field was introduced to avoid code churn

nicobav (changes requested):
Looks good overall, some cleanup things mostly :+1: 

You also need to rebase with main and resolve conflicts and  also re-create the migration file since in the main there is now already migration file in reservation_units with number 0056 in it.

saaste:
I close this one. #526 has much cleaner commit history.

PR #530, opened 2022-06-02
TILA-1694: Allow filtering units by name and service sector

saaste:
Adds name and service sector filtering to `UnitType`


nicobav (approved):
:shipit:

PR #531, opened 2022-06-07
TILA-1613 | Add capability for querying application events to gql

nicobav:
First part of the application event gql api - mutations coming after this. 

TILA-1613

### DIFF
--- a/api/graphql/schema.py
+++ b/api/graphql/schema.py
@@ -8,8 +8,15 @@ from graphene_permissions.mixins import AuthFilter
 from graphene_permissions.permissions import AllowAny, AllowAuthenticated
 from rest_framework.generics import get_object_or_404
 
-from api.graphql.applications.application_filtersets import ApplicationFilterSet
-from api.graphql.applications.application_types import ApplicationType, CityType
+from api.graphql.applications.application_filtersets import (
+    ApplicationEventFilterSet,
+    ApplicationFilterSet,
+)
+from api.graphql.applications.application_types import (
+    ApplicationEventType,
+    ApplicationType,
+    CityType,
+)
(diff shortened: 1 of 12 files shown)


saaste (commented):
This looks good to me! It also has some nice, more complex query examplesfor me 😄 
I think this is good to go as soon as tests are green ✅

api/graphql/schema.py
```diff
@@ -169,6 +177,47 @@ def resolve_queryset(
...
+        ).values_list("unit", flat=True)
+        group_ids = user.unit_roles.filter(
+            role__permissions__permission="can_validate_applications"
+        ).values_list("unit_group", flat=True)
+        units = Unit.objects.filter(
+            Q(id__in=unit_ids) | Q(unit_groups__in=group_ids)
+        ).values_list("id", flat=True)
+
+        return queryset.filter(
+            Q(
```
saaste, line 211: I have to somehow remember these cryptic one-letter functions like `Q`, `V` and `F` Looks useful but it probably takes a while until I remember to use them 😅

saaste (approved):
🚢  it!

api/graphql/schema.py
nicobav: The `Q` here is a class for making db queries and here it is used because want to have `OR` condition => so the `Q`s can be used in AND/OR operations and then can be passed to `.filter` like:
`condition = (Q(field_one__gte=else) & Q(field_one__lte=other) | Q(field_too=False)`
`Model.objects.filter(condition)`

nicobav:
@kieferbonk @siren This renames `ageGroupDisplay` field to `ageGroup` from `ApplicationEventType`

PR #537, opened 2022-06-14
TILA-1725: Add ReservationUnit state property

saaste:
State is a computed field based on the following properties:
- is_draft
- is_archived
- publish_beings / publish_ends
- reservation_begins / reservation_ends

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_types.py
+++ b/api/graphql/reservation_units/reservation_unit_types.py
@@ -303,6 +303,7 @@ class ReservationUnitType(AuthNode, PrimaryKeyObjectType):
     buffer_time_before = Duration()
     buffer_time_after = Duration()
     metadata_set = graphene.Field(ReservationMetadataSetType)
+    state = graphene.String()
 
     permission_classes = (
         (ReservationUnitPermission,)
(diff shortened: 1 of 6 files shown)


api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -303,6 +303,7 @@ class ReservationUnitType(AuthNode, PrimaryKeyObjectType):
     buffer_time_before = Duration()
     buffer_time_after = Duration()
     metadata_set = graphene.Field(ReservationMetadataSetType)
+    state = graphene.String()
```
nicobav, line 306: This would be nice to get to the api as enum so that the values can be found in doc. But we don't have the "standardized" way of doing that yet so its fine as is.
It could be graphene.Enum type of field which *should* work with python enums (I have tested without luck :( ) Then again when have the field in the model as charfield and it has choices defined it will work _out-of-the-box_

saaste: I'll merge this now and let's figure out a good way to handle those enums. I personally don't like to use "magic strings" so if there is a good way to use enums, I'd like to use that 😄

PR #534, opened 2022-06-09
TILA-1538: Update ApplicationRound, Application and ApplicationEvent statuses

saaste:
This PR removes unused statuses and adds new ones to the following classes:
- `ApplicationRoundStatus`
- `ApplicationStatus`
- `ApplicationEventStatus`

All tests are updated to match the changes and I also wrote new ones to cover the new cases.

I created a separate utility module for `ApplicationRound` status changes. It deals with cases where status of a child entity is updated when `ApplicationRound` status changes.

This was a bit challenging task because of the lack of knowledge of the previous status handling 😅  So take your time to review it. There might be stuff I missed and understood wrong.

PR #538, opened 2022-06-15
TILA-1538: ApplicationRound/Application/ApplicationEvent status data migration

saaste:
In the [previous PR](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/534) we updated those statuses in the code but forgot to migrate the data in the database. This lead to a bug where code couldn't handle the old values coming from the database.

Querying those models lead to the following error: `Expected a value of type 'applicationEventStatus' but received: 'allocated'`. This PR fixes the issue by adding the necessary data migrations.

⚠️ This is a migration that **CANNOT** be rolled back! ⚠️


applications/migrations/0060_change_application_status_values.py
```diff
@@ -0,0 +1,24 @@
...
+    pass
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("applications", "0059_change_application_round_status_values"),
+    ]
+
+    operations = [migrations.RunPython(change_status_to_new, change_status_to_old)]
```
nicobav, line 24: Nice that there's a reverse code :+1:  
Tip wednesday: instead of giving none doing function, we can give/say `migrations.RynPython.noop`:  `operations = migrations.RunPython(change_status_to_new, migrations.RunPython.noop)]` so it does nothing but can be "reverted"

nicobav (approved):
These migrations are great as is :+1:  

Another thing is the previous migration file which removed old and introduced new application statuses. I believe reverting it will cause same like data issues as well. But case we end up reverting to it might not happen and we can manage it by hand then so totally optional to edit that migration file to have its own revert function.

applications/migrations/0060_change_application_status_values.py
saaste: Lovely! I cleaned the scripts 🧹

PR #539, opened 2022-06-15
TILA-1741: Hide archived reservation units by default

saaste:
The only way to interact with archived reservation units is the Django Admin, I also removed `is_archived` filter since there is no use for it.

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_filtersets.py
+++ b/api/graphql/reservation_units/reservation_unit_filtersets.py
@@ -46,8 +46,6 @@ class ReservationUnitsFilterSet(django_filters.FilterSet):
 
     is_draft = django_filters.BooleanFilter(field_name="is_draft")
 
-    is_archived = django_filters.BooleanFilter(field_name="is_archived")
-
     is_visible = django_filters.BooleanFilter(method="get_is_visible")
 
     application_round = django_filters.ModelMultipleChoiceFilter(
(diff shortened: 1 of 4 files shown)


api/graphql/reservation_units/reservation_unit_filtersets.py
```diff
@@ -101,6 +99,11 @@ class Meta:
         model = ReservationUnit
         fields = ["pk", "unit", "keyword_groups"]
 
+    # Hide archived reservation units
+    def filter_queryset(self, queryset):
```
nicobav, line 103: I'd add this functionality to `ReservationUnitsFilter`'s `resolve_queryset` method in `schema.py` 

Its clear it works from here also, but there in the schema file we've done already some filtering to the types

nicobav (approved):
Basic logic seems good, but placement could be different

nicobav (changes requested):
wrong button before :joy:

api/graphql/reservation_units/reservation_unit_filtersets.py
saaste: Took me a while to figure out where to put this code so that it affects only GQL and not Django Admin. This was the first location that seemed to work 😄 

I'll move it! 👍

nicobav (approved):
Beautiful!

PR #540, opened 2022-06-15
TILA-1165: Remove is_draft from resource

saaste:
Previously, validations happened only for non-draft resources. After this change, it is not possible to save draft resources so validations happen every time. I also removed tests that were written for draft resources.


nicobav (approved):
Goodbye draft resources :wave: :ship:

PR #541, opened 2022-06-15
TILA-1649 | Application round statistics command

nicobav:
Django command for exportin csv file including reservation unit data per application round

TILA-1649


nicobav:
This is to be later enchanced perhaps, but better not leave this lying in the repo as own branch.

saaste (approved):
Looks great! 👏 
I did not know there is a way to write custom commands ☺️

PR #542, opened 2022-06-16
TILA-1749: Remove description and translation requirements from resource

saaste:
Description is not in use anymore and it should be possible to save resources without translations.


nicobav (approved):
Looks good :rocket: 

Slightly pondered about saving the description field in model for future needs but yolo lets do this :sweat_smile:

PR #543, opened 2022-06-17
TILA-1730: Filtering reservation units by state

saaste:
`state` is a calculated field.  Because the value depends on the current time, we can't really store it to the database. For these reasons I made a separate utility class that handles the state calculation and filter generation. I couldn't figure out an easier way to do this, but if you know how to improve this, please let me know ☺️ 

Oh I also switched strings to enums ☺️

### DIFF
--- a/reservation_units/models.py
+++ b/reservation_units/models.py
@@ -4,7 +4,6 @@ import uuid as uuid
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import models
-from django.utils.timezone import get_default_timezone
 from django.utils.translation import gettext_lazy as _
 from easy_thumbnails.fields import ThumbnailerImageField
 
@@ -574,21 +573,11 @@ class ReservationUnit(models.Model):
 
     @property
     def state(self) -> ReservationUnitState:
-        now = datetime.datetime.now(tz=get_default_timezone())
-        if self.is_archived:
-            return ReservationUnitState.ARCHIVED
-        elif self.is_draft:
-            return ReservationUnitState.DRAFT
-        elif (self.publish_begins and now < self.publish_begins) or (
-            self.publish_ends and now >= self.publish_ends
-        ):
-            return ReservationUnitState.SCHEDULED_PUBLISHING
-        elif (self.reservation_begins and now < self.reservation_begins) or (
-            self.reservation_ends and now >= self.reservation_ends
-        ):
-            return ReservationUnitState.SCHEDULED_RESERVATION
-        else:
-            return ReservationUnitState.PUBLISHED
+        from reservation_units.utils.reservation_unit_state_helper import (
+            ReservationUnitStateHelper,
+        )
+
+        return ReservationUnitStateHelper.get_state(self)
 
 
 class ReservationUnitImage(models.Model):
--- a/reservation_units/models.py
+++ b/reservation_units/models.py
@@ -573,7 +573,7 @@ class ReservationUnit(models.Model):
         return settings.HAUKI_ORIGIN_ID
 
     @property
-    def state(self) -> str:
+    def state(self) -> ReservationUnitState:
         now = datetime.datetime.now(tz=get_default_timezone())
         if self.is_archived:
             return ReservationUnitState.ARCHIVED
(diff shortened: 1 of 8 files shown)


reservation_units/utils/tests/test_reservation_unit_state_helper.py
```diff
@@ -0,0 +1,80 @@
...
+from assertpy import assert_that
+from django.test.testcases import TestCase
+
+from reservation_units.enums import ReservationUnitState
+from reservation_units.models import ReservationUnit
+from reservation_units.tests.factories import ReservationUnitFactory
+from reservation_units.utils.reservation_unit_state_helper import get_state
+
+
+class ReservationUnitStateHelperTestCase(TestCase):
```
nicobav, line 12: Adding a test for raising the `ValueError` would be nice.

reservation_units/utils/reservation_unit_state_helper.py
```diff
@@ -0,0 +1,124 @@
...
+        return ReservationUnitState.SCHEDULED_RESERVATION
+    if is_published(reservation_unit):
+        return ReservationUnitState.PUBLISHED
+
+    raise ValueError(
+        f"Unable to detect the state of the reservation unit {reservation_unit.id}"
+    )
+
+
+def get_state_query(state: str) -> Q:
```
nicobav, line 110: This whole logic could be wrapped into a class instead of separate functions :shrug: At least there's a some functions smelling like private methods for a class :thinking:

reservation_units/models.py
```diff
@@ -573,22 +572,10 @@ def hauki_resource_data_source_id(self):
...
-        elif (self.reservation_begins and now < self.reservation_begins) or (
-            self.reservation_ends and now >= self.reservation_ends
-        ):
-            return ReservationUnitState.SCHEDULED_RESERVATION
-        else:
-            return ReservationUnitState.PUBLISHED
+    def state(self) -> ReservationUnitState:
+        from reservation_units.utils.reservation_unit_state_helper import get_state
+
+        return get_state(self)
```
nicobav, line 578: this can raise a `ValueErrror` should it be handled?

nicobav (changes requested):
Looks good overall, only some minor things to ponder about

reservation_units/models.py
saaste: The problem is that currently it is not possible to get this exception. `is_published` is basically an `else` branch but I didn't want to call it such because it is more complex in `get_is_published_query()`. So I don't really know how this could be tested 🤔 

I added the `throw` there for the future cases where we end up changing the definitions and somehow forget to cover some corner case. In that case it will throw an exception instead of just swallowing the missing case with `else`.

What are your thoughts on this? 😄

reservation_units/utils/reservation_unit_state_helper.py
saaste: Changed to class and made most of the functions private methods. Feel free to check if my approach to "classification" follows the best practices 🥸

reservation_units/models.py
nicobav: Well this is getting kind of a test coverage issue then, but now we have code block that isn't tested. I like the idea that it will throw an error in future if those states are edited / add so removing it is something that we don't want.
Some test for a new type of enum value could be therefore invented if possible.

saaste: Actually, now that I think of it, my reasoning is not very solid for two reasons:

1) We are using enums. If we ever add a new value, it is very easy to find the code where enum is used and update it
2) We are expecting tests for all enum values. If we ever add a new value, we need to write a test for it. That requires that we also update the implementation.

As a bonus reason, my idea hugs the magical premature optimization territory. I'm doing stuff based on my own expectations of the future. Not gud!

So I think the idea of raising "safety exception" is kinda pointless because the style the code is implemented and our testing requirements already handle everything.

I'll change the implementation to use if-else statement. It'll get rid of this nasty untested code issue ☺️

reservation_units/utils/tests/test_reservation_unit_state_helper.py
saaste: `ValueError` is now gone ☺️

nicobav (approved):
Noice :guitar:

PR #546, opened 2022-06-20
TILA-1758 | ApplicationEventScheduleResult GraphQl API

nicobav:
Adds `ApplicationEventScheduleResult` to GraphQl. 
This can be queried through `ApplicationEventSchedule` in `ApplicationEventType`

Adds mutations to create and edit `ApplicationEventScheduleResult`s 

TILA-1758

### DIFF
--- a/api/graphql/applications/application_types.py
+++ b/api/graphql/applications/application_types.py
@@ -31,9 +31,10 @@ from permissions.api_permissions.graphene_field_decorators import (
 )
 from permissions.api_permissions.graphene_permissions import (
     AddressPermission,
+    ApplicationEventScheduleResultPermission,
     ApplicationPermission,
     CityPermission,
-    OrganisationPermission, ApplicationEventScheduleResultPermission,
+    OrganisationPermission,
 )
 from spaces.models import Space
 from utils.query_performance import QueryPerformanceOptimizerMixin
--- a/api/graphql/applications/application_types.py
+++ b/api/graphql/applications/application_types.py
@@ -33,7 +33,7 @@ from permissions.api_permissions.graphene_permissions import (
     AddressPermission,
     ApplicationPermission,
     CityPermission,
-    OrganisationPermission,
+    OrganisationPermission, ApplicationEventScheduleResultPermission,
 )
 from spaces.models import Space
 from utils.query_performance import QueryPerformanceOptimizerMixin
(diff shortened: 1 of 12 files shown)


api/graphql/applications/application_types.py
```diff
@@ -107,16 +112,58 @@ class ApplicationEventAggregatedDataType(graphene.ObjectType):
...
+        graphene.Enum("allocatedDay", [(v.upper(), k) for k, v in DAY_CHOICES])
+    )
+
+    class Meta:
+        model = ApplicationEventScheduleResult
+        fields = (
+            "id",
+            "pk",
+            "accepted",
+            "declined",
```
saaste, line 131: What is the reason we have separate fields for `accepted` and `declined`? I know this comes from `ApplicationEventScheduleResult` but do you know/remember the reason? 🤔

saaste (approved):
Looks noice to me! Just one generic question.
The PR is still in draft state, are you planning to add mutations?

api/graphql/applications/application_types.py
nicobav: declineds are for the weekly amount reduction logic for the algorithm

nicobav:
> Looks noice to me! Just one generic question. The PR is still in draft state, are you planning to add mutations?

Yes, draft state it was, but now added the mutations and ready for :mag:

permissions/api_permissions/graphene_permissions.py
```diff
@@ -91,6 +92,46 @@ def has_permission(cls, info: ResolveInfo) -> bool:
...
+        )
+
+    @classmethod
+    def has_permission(cls, info: ResolveInfo) -> bool:
+        user = info.context.user
+
+        if user and user.is_anonymous:
+            return False
+
+        service_sector_roles = user.service_sector_roles.all()
```
saaste, line 120: Is there a need to check if `user` is `None`?
Just thinking if it is possible for this code to crash because `None` type does not have `service_sector_roles` property?

saaste (commented):
Looks good to me! Just one question regarding None type checks ☺️

permissions/api_permissions/graphene_permissions.py
nicobav: good point, I guess we can make this `or`

nicobav: OR not! :smile: I'll refactor anyhow

saaste (approved):
Yey! Let's 🚢  it!

PR #547, opened 2022-06-22
TILA-1734 | Add applicantType filter to application and applicationEvents

nicobav:
GraphQl filters for application event and application queries

TILA-1734


saaste (approved):
Awesome! 👌

PR #548, opened 2022-06-22
TILA-1735 | Add application to ApplicationEventType

nicobav:
So that one can navigate to application from application event if want
so.

TILA-1735


saaste (approved):
Sweet! :shipit:

PR #549, opened 2022-06-22
TILA-1736 | Remove obsolete id references and change necessary to pk

nicobav:
In application graphql types there was some *_id fields declared which
we don't need in graphql since we can just add the corresponding
relative type.
This removes the id fields and introduces relatative types if needed and
changes one id to _pk which is what we prefer.

TILA-1736


nicobav:
Added @siren and @kieferbonk also as reviewer so you could check UI things before we merge and :boom:

saaste (approved):
Looks great! Nice cleanup 🧹

PR #550, opened 2022-06-23
TILA-1743: Add reservation user filter

saaste:
This also addresses TILA-1491 by moving permission checks behind filter.


api/graphql/reservations/reservation_filtersets.py
```diff
@@ -22,6 +29,12 @@ class ReservationFilterSet(django_filters.FilterSet):
     # Filter for displaying reservations which requires or had required handling.
     requested = django_filters.BooleanFilter(method="get_requested")
 
+    only_with_permission = django_filters.BooleanFilter(
+        method="get_only_with_permission"
+    )
+
+    user = django_filters.NumberFilter(method="get_user")
```
nicobav, line 36: We could use `ModelChoiceFilter` and then we don't need the get_user  method

api/graphql/tests/test_reservations/test_reservation_queries.py
```diff
@@ -296,6 +340,149 @@ def test_reservation_total_count(self):
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
 
+    def test_hide_fields_with_personal_information(self):
+        self.create_reservation_by_admin()
+        self.client.force_login(self.regular_joe)
+        response = self.query(
```
nicobav, line 346: Lots of duplicate query string, perhaps add a class method for getting it?
Also could add begin and end fields at least to confirm that reservation time things are visible.

nicobav (changes requested):
Looks nice :+1: Only some friendly suggestions to ponder about

api/graphql/reservations/reservation_filtersets.py
saaste: Oh, true! Replaced 🧹

api/graphql/tests/test_reservations/test_reservation_queries.py
saaste: Great ideas! Implemented both 🧹

nicobav (approved):
Magnificent :shipit:

PR #551, opened 2022-06-27
TILA-1749: Fix resource update validation

saaste:
- Finnish name translation is required
- English and Swedish can be `null`

Tests are also updated to match the change.


saaste:
This should fix the issue you discovered, @siren 🧹

nicobav (approved):
Perfect :pinched_fingers:

PR #552, opened 2022-06-28
TILA-1716: Update maxPersons and minPersons filters

saaste:
MaxPersons filters were updated. MinPersons filters are new.


nicobav (approved):
oh yes :rocket: :metal:

PR #553, opened 2022-06-28
TILA-1551 | Application and event decline mutations

nicobav:
TILA-1551

### DIFF
--- a/api/graphql/applications/application_decline_serializers.py
+++ b/api/graphql/applications/application_decline_serializers.py
@@ -1,0 +1,84 @@
+from rest_framework import serializers
+
+from api.graphql.base_serializers import PrimaryKeySerializer
+from applications.models import (
+    Application,
+    ApplicationEvent,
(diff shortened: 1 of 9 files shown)


api/graphql/applications/application_decline_serializers.py
```diff
@@ -0,0 +1,84 @@
...
+                f"Only applications with status as {', '.join(self.CAN_DECLINE_STATUSES)} can be approved."
+            )
+        data = super().validate(data)
+
+        return data
+
+    def save(self, **kwargs):
+        for event in self.instance.application_events.all():
+            event.set_status(ApplicationEventStatus.DECLINED)
+        return self.instance
```
saaste, line 42: First I thought that you are missing `self.instance.set_status(ApplicationStatus.DECLINED)`, but then I remembered that this is not how `ApplicationStatus` works 😄

saaste (approved):
Beautiful! 👸

PR #554, opened 2022-06-29
Add missing migrations for reservation unit

nicobav:
Pricing type choices had changed and same for reservation kind.


saaste (approved):
I think that at least pricing type was my mistake 🙈 
Thanks for fixing this so quickly! 🙇‍♂️

PR #557, opened 2022-06-30
TILA-1773: Add pricing fields to Reservation Unit mutation.

saaste:
Originally, I thought that the pricing fields are managed via Django Admin, but that assumption turned out to be wrong. So, now they are part of the mutations.

Refs TILA-1773

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_serializers.py
+++ b/api/graphql/reservation_units/reservation_unit_serializers.py
@@ -18,6 +18,7 @@ from api.reservation_units_api import (
 from reservation_units.models import (
     Equipment,
     EquipmentCategory,
+    PricingType,
     Purpose,
     ReservationKind,
     ReservationUnit,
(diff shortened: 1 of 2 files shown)


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -152,6 +153,12 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         required=False,
         allow_null=True,
     )
+    pricing_terms_pk = serializers.PrimaryKeyRelatedField(
```
nicobav, line 156: Using `IntegerPrimaryKeyField` would be prefferred since this field from drf will interpret itself as string field in the api.

saaste: I think we have the same problem with all terms fields. I could fix all of them if that is the preferred way 🤔

saaste: `TermsOfUse` ID which is the primary key is `CharField`: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/terms_of_use/models.py#L25

I have a feeling `IntegerPrimaryKeyField` is not really the way to go here ☺️

nicobav: Oh, I stand corrected. Didn't lookup inside of the model itself. Its alright then :+1: 

(Just wonder why the id field was defined as charfield :thinking: )

nicobav (approved):
LGTM :rocket:

PR #558, opened 2022-06-30
Set reservation to REQUIRES_HANDLING on subsidy

saaste:
When reservation has applying_for_free_of_charge on, reservation is moved to REQUIRES_HANDLING state on confirm mutate call.

Refs TILA-1767


nicobav (approved):
Nice :ship:

PR #559, opened 2022-07-01
TILA-1714 | Lower ReservationUnit reservation_kind values

nicobav:
They were mistakenly initiated as upper case instead of preferred lower case
This caused some unwanted situtation in django admin etc.

TILA-1714


saaste (approved):
🚢  it!

PR #561, opened 2022-07-04
TILA-1774 - Reservation filters

saaste:


api/graphql/reservations/reservation_filtersets.py
```diff
@@ -80,3 +110,61 @@ def get_only_with_permission(self, qs, property, value: bool):
...
+                When(
+                    reservee_type=CUSTOMER_TYPES.CUSTOMER_TYPE_INDIVIDUAL,
+                    then=Concat("reservee_first_name", V(" "), "reservee_last_name"),
+                ),
+                default=V(""),
+                output_field=CharField(),
+            )
+        )
+
+        # Match name
```
nicobav, line 167: Does it need to match? Since the code below will apply to partial matches. So should it be exact or is the contains what is wanted?

api/graphql/reservations/reservation_filtersets.py
```diff
@@ -37,6 +47,26 @@ class ReservationFilterSet(django_filters.FilterSet):
...
+
+    unit = django_filters.ModelMultipleChoiceFilter(
+        method="get_unit", queryset=Unit.objects.all()
+    )
+
+    reservation_unit_type = django_filters.ModelMultipleChoiceFilter(
+        method="get_reservation_unit_type", queryset=ReservationUnitType.objects.all()
+    )
+
+    text_search = django_filters.CharFilter(method="get_text_search")
```
nicobav, line 68: I'm pondering this `text_search` naming. Looking inside the `get_text_search` function I see that there is only name related search done there.

nicobav (approved):
Looks good to me! :beers: 

Left one or two question but no change request

api/graphql/reservations/reservation_filtersets.py
```diff
@@ -1,19 +1,29 @@
+import operator
+from functools import reduce
+
 import django_filters
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.db.models import Q
+from django.db.models import Case, CharField, F, Q
+from django.db.models import Value as V
+from django.db.models import When
```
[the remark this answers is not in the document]
saaste: I was wondering the same thing! 🤯 🤷

api/graphql/reservations/reservation_filtersets.py
saaste: Oops! This comments shouldn't even be here. `icontains` is the one that is expected ☺️ 
I removed the pointless comment.

api/graphql/reservations/reservation_filtersets.py
saaste: Yeah, I see your point. I guess we wanted this because we already had these "super searches" in other entities 🤔

PR #562, opened 2022-07-06
TILA-1639 | Reservations can be made to as long as reservation unit allows

nicobav:
... with a limitation of two years. 
Previously reservations could be made only to next spring because of the seasonal reservation applications etc. but since from that spec we have been introduced few fields in the ReservationUnit which we can use to determine these times.

Also fixes one buggy in the scheduler where scrolling time over daylight saving using timedelta and the timezone
diff did not get noticed and thus it resulted those cases the winter diff (+02.00 instead of +03:00

TILA-1639


saaste (approved):
Great stuff! 👏

PR #563, opened 2022-07-06
Bump django to latest

nicobav:


saaste (approved):

saaste:
> [Trunc()](https://docs.djangoproject.com/en/4.0/ref/models/database-functions/#django.db.models.functions.Trunc) and [Extract()](https://docs.djangoproject.com/en/4.0/ref/models/database-functions/#django.db.models.functions.Extract) database functions were subject to SQL injection if untrusted data was used as a kind/lookup_name value.

Classic SQL injection 🕺 💃

PR #564, opened 2022-07-07
TILA-1775: Add new reservation ordering fields

saaste:
- reservation unit name (all languages)
- unit name (all languages)
- reservee name

Refs TILA-1775

### DIFF
--- a/api/graphql/tests/test_reservations/test_reservation_queries.py
+++ b/api/graphql/tests/test_reservations/test_reservation_queries.py
@@ -836,6 +836,136 @@ class ReservationQueryTestCase(ReservationTestCaseBase):
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
 
+    def test_order_by_reservation_unit_name(self):
+        resunitA = ReservationUnitFactory(
+            name_fi="a Unit", name_en="d Unit", name_sv="g unit"
+        )
+        resunitB = ReservationUnitFactory(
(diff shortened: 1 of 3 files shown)


api/graphql/tests/test_reservations/test_reservation_queries.py
```diff
@@ -836,6 +836,136 @@ def test_filter_by_text_search_individual_reservee_name(self):
...
+        ReservationFactory(name="this should be 2nd", reservation_unit=[resunitB])
+        ReservationFactory(name="this should be 3rd", reservation_unit=[resunitC])
+
+        self.client.force_login(self.general_admin)
+        test_data = ["fi", "en", "sv"]
+        for lang in test_data:
+            response = self.query(
+                """
+                query {
+                    reservations(orderBy:"reservation_unit_name_%s") {
```
nicobav, line 860: There should be some magic in the graphene django code and this filter name should work as camel case (as we camelcase the api). So we could/should have tests in camelCases since its the way that the frontend is using this.
`reservations(orderBy: "reservationUnitNameFi"` etc.

nicobav (changes requested):
Looks great :+1: 

Only some miney minor test stuff.

api/graphql/tests/test_reservations/test_reservation_queries.py
saaste: Learned something new today! 🧠 
I changed the tests so that they now use cameCase 🧹

nicobav (approved):
very noice :sunglasses:

PR #565, opened 2022-07-07
TILA-1592 | Application mutations to GraphQL as is in REST

nicobav:
Ports application create and update operations from REST api so behavior should  be the same in both.

Some refactors etc also done

TILA-1592


nicobav:
We reached above 1000 unit tests :beers:

api/graphql/tests/test_applications/test_application_create.py
```diff
@@ -0,0 +1,255 @@
...
+                createApplication(input: $input){
+                    pk
+                    errors {
+                        messages field
+                    }
+                }
+            }
+        """
+
+    def get_application_data(self, create_events=False):
```
saaste, line 50: Looks like `create_events` parameter is not used 🤔

api/graphql/tests/test_applications/test_application_create.py
```diff
@@ -0,0 +1,255 @@
...
+        application = Application.objects.get(id=app_data.get("pk"))
+        assert_that(application).is_not_none()
+        assert_that(application.status).is_equal_to(data["status"])
+
+    def test_application_create_organization_identifier_not_required(self):
+        assert_that(Application.objects.count()).is_equal_to(1)
+        self.client.force_login(self.regular_joe)
+
+        data = self.get_application_data()
+        _ = data["organisation"].pop("identifier")
```
saaste, line 127: Is there a reason to assign the popped value to `_`?

api/graphql/tests/test_applications/test_application_create.py
```diff
@@ -0,0 +1,255 @@
...
+        self.client.force_login(self.regular_joe)
+        data = self.get_application_data()
+        data["organisation"].update({"email": None})
+
+        response = self.query(self.get_create_query(), input_data=data)
+        assert_that(response.status_code).is_equal_to(200)
+
+        content = json.loads(response.content)
+        app_data = content.get("data").get("createApplication")
+        assert_that(app_data.get("errors")).is_not_none()
```
saaste, line 219: Most failure tests check that the data was not updated to the database. Perhaps it could be useful here as well:
```
assert_that(Application.objects.count()).is_equal_to(1)
```

api/graphql/applications/application_serializers.py
```diff
@@ -0,0 +1,247 @@
...
+        self, organisation_data: Dict[Any, Any]
+    ) -> Union[Organisation, None]:
+        if organisation_data is not None:
+            # CASE: Create new organisation
+            if "pk" not in organisation_data or organisation_data["pk"] is None:
+                return OrganisationSerializer(data=organisation_data).create(
+                    validated_data=organisation_data
+                )
+
+            # CASE: Update exisitng organisation
```
saaste, line 172:
```suggestion
            # CASE: Update existing organisation
```

saaste (changes requested):
Awesome! Soooo many new tests! 👏 
I left some teeny-tiny comments and nitpicks, but other than that looks great to me 🥇

api/graphql/tests/test_applications/test_application_create.py
nicobav: yup, true. Also not `get_event_data` - I started to having the create and update tests in same file so these are leftovers

 - I'll perhaps get rid-off these or make a test which has the events as well in the creation (if we actually support it :sweat_smile: )

api/graphql/tests/test_applications/test_application_create.py
nicobav: absolutely no reason - _ is a throwaway variable. Using it in popping is useless but for e.g when getting something which is tuple it is a convention used for _not used variables like:
`a, _, c, _ = some_func_returns_four_var_tuple()` or `for k, _ in dictionary.items():`

I'll remove it here

api/graphql/applications/application_serializers.py
nicobav: Nice found! Didn't see that even my IDE seems to nag about it :joy:

api/graphql/tests/test_applications/test_application_create.py
nicobav: added this

api/graphql/tests/test_applications/test_application_create.py
nicobav: removed.

api/graphql/tests/test_applications/test_application_create.py
nicobav: I added test for adding events in create - I had refactor the code to actually support it. Mut was a miney change in serializer.

api/graphql/applications/application_serializers.py
nicobav: fix'd

saaste (approved):
Very nice

PR #574, opened 2022-08-12
TILA-1805: Add reservation staff_event field

saaste:
- Visible only with specific permissions
- Mutations allowed only with specific permissions


nicobav (approved):
NOICE!

PR #575, opened 2022-08-12
TILA-1804: Revert JWT audience default value change

saaste:
The previous change I made to JWT audience value was just a lucky guess that happened to fix the issue back then. But it is not the right solution so... reverting it 😄


nicobav (approved):
:broom:

kieferbonk (commented):
`.env.example:163` could be touched too.

saaste:
> .env.example:163 could be touched too.

Good catch! Fixed!

PR #577, opened 2022-08-12
TILA-1819: Add reservation type

saaste:
- Requires special permission to query/mutate
- Similar to staff_event field


reservations/models.py
```diff
@@ -22,6 +22,11 @@
 User = get_user_model()
 
 
+class ReservationType(models.TextChoices):
+    TYPE_NORMAL = "normal"
```
nicobav, line 26: Could these be called as `NORMAL` and `BLOCKED`?

nicobav (changes requested):

reservations/models.py
saaste: Sure! Fixed 🛠️

nicobav (approved):

PR #579, opened 2022-08-16
TILA-1550 | Flagging application events

nicobav:
Adds `flagged` field to `ApplicationEvent` and two mutations for changing it through the gql api:
 - For single `ApplicationEvent`
 - For all of the events in an `Application` 

TILA-1550


saaste (changes requested):
Tests are failing but otherwise looks great! 👏

nicobav:
> Tests are failing but otherwise looks great! clap

Yes, the visualizations... something is wonky there since it fails after every update :exploding_head:

saaste (approved):

PR #580, opened 2022-08-17
Tila 1816 - Add error codes to reservation mutation validation errors

saaste:
This will change all reservation mutation validation errors to include a separate `error_code` and `field` fields. `field` value is `nonFieldError` unless error is specific to a certain field. I did not change the actual error messages.

In same cases different error messages might have the same code. I tried to use common sense when combining them. For example:
```
{
    "message": "Reservation start time is earlier than N days before.",
    "error_code": "RESERVATION_NOT_WITHIN_ALLOWED_TIME_RANGE"
}

{
    "message": "Reservation start time is less than N days before.",
    "error_code": "RESERVATION_NOT_WITHIN_ALLOWED_TIME_RANGE"
}
```

All error codes are defined in one file. I tried to use quite long and verbose names so that we don't need to have a separate documentation for the codes. Also, when it is implemented like this, it is easy to see where each code is used.


api/graphql/application_errors.py
```diff
@@ -5,7 +5,34 @@
 
 
 class ValidationErrorCodes(Enum):
-    pass
+    RESERVATION_UNIT_NOT_RESERVABLE = "RESERVATION_UNIT_NOT_RESERVABLE"
```
saaste, line 8: @kieferbonk @siren What do you think, are these error codes clear enough so that no extra documentation is not needed?

nicobav (approved):
Awesome :clap:

PR #581, opened 2022-08-18
TILA-1829: Add reservationUnit filter

saaste:
Add a new filter to Reservations GQL query: reservationUnit. Filter supports arrays so multiple Reservation Unit PKs can be used as a filter.


nicobav (approved):
LGTM :ok_hand:

PR #582, opened 2022-08-23
Add migration containing correct reservation types

nicobav:


saaste (approved):
Oops! Nice catch and thanks for fixing it! 👍

PR #583, opened 2022-08-23
TILA-1823 | Reservee language

nicobav:
Adds `reservee_language` field to `Reservation` model and gql api update and create functions.

TILA-1823


saaste (approved):
Grrrrreat! 🐯

PR #587, opened 2022-08-25
TILA-1820: Add current user query and mutation to GQL

saaste:
When I started working on this task, it was just a matter of adding a new field to `User` model. Initially, I did these changes to the existing REST API and only after I was done, I remembered that all new stuff should be done on the GraphQL API.

So, I ended up copying the existing API functionality to GQL and also added the new field. I kept the commits where I changed the old API. What do you think, should I remove them or is it OK to keep them?


api/graphql/users/user_filtersets.py
```diff
@@ -0,0 +1,10 @@
+import django_filters
+
+
+class UsersFilterSet(django_filters.FilterSet):
+    def filter_queryset(self, queryset):
```
nicobav, line 5: Would be in line with existing implementations to move this logic into schema.py's `UsersFilter` class in the method `resolve_queryset`

nicobav (changes requested):
Looks good, some minor tweaks. 

Also raised a discussion in slack about the `is_staff` property.

api/graphql/tests/test_users.py
```diff
@@ -0,0 +1,293 @@
...
+    UnitRole,
+    UnitRoleChoice,
+)
+from spaces.tests.factories import ServiceSectorFactory, UnitFactory, UnitGroupFactory
+from users.models import ReservationNotification, User
+
+
+class UserQueryTestCaseBase(GraphQLTestCase, snapshottest.TestCase):
+    @classmethod
+    def setUpTestData(cls):
```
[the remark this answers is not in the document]
saaste: Since both Query and Update tests share the same setup, I renamed this base class to `UserTestCaseBase` and made both test classes to inherit this. That way I got rid of some extra code and don't have to repeat the same setup ☺️

api/graphql/tests/test_users.py
```diff
@@ -0,0 +1,293 @@
...
+        )
+
+        assert_that(response.status_code).is_equal_to(200)
+        content = json.loads(response.content)
+        self.assertMatchSnapshot(content)
+
+
+class UserMutateTestCaseBase(GraphQLTestCase, snapshottest.TestCase):
+    @classmethod
+    def setUpTestData(cls):
```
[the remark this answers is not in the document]
saaste: See my previous comment ⬆️

api/graphql/users/user_filtersets.py
saaste: I managed to do this, but when I removed `UsersFilterSet` class from the code, I started to get errors about missing `Meta.fields` from `UsersFilterSet` class. I didn't really understand where it comes from. I'm assuming it is raising from Django base class or something like that because I got this error even when I removed `UsersFilterSet` class.

I solved the issue by leaving the class but making it empty with `pass`. Let me know if you know a better way to handle this ☺️

saaste:
> Also raised a discussion in slack about the is_staff property.

Logic is now updated. Instead of `is_staff`, user must be a superuser or have at least one role. Updating other users is still not allowed - even for superuser.

api/graphql/users/user_filtersets.py
nicobav: perhaps we need some follow up ticket to enhance this to something like `user is in at least one of these roles` type of check but that needs refinement so this is fine for now.

api/graphql/tests/test_users.py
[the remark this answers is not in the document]
nicobav: oh, didn't notice that :+1:

nicobav:
> > Also raised a discussion in slack about the is_staff property.
> 
> Logic is now updated. Instead of `is_staff`, user must be a superuser or have at least one role. Updating other users is still not allowed - even for superuser.



perhaps we need some follow up ticket to enhance this to something like user is in at least one of these roles type of check but that needs refinement so this is fine for now.

api/graphql/users/user_filtersets.py
saaste: Managed to get this fixed with your help! 🥳

nicobav: Yup, slacked.

api/graphql/schema.py
```diff
@@ -292,6 +295,26 @@ class UnitsFilter(AuthFilter, django_filters.FilterSet):
...
+    @classmethod
+    def resolve_queryset(
+        cls, connection, iterable, info, args, filtering_args, filterset_class
+    ):
+        queryset = super().resolve_queryset(
+            connection, iterable, info, args, filtering_args, filterset_class
+        )
+
+        if not info.context.user:
+            return None
```
nicobav, line 312: None might work as good and at least does not return any users and  `return queryset.none()` would do almost same but in difference will return an empty queryset

saaste: Fix'd!

nicobav (approved):

PR #588, opened 2022-08-29
TILA-1843: Update reservation time validation

saaste:
Reservations_min_days_before is not checked against current time
but the beginning of the current day. Previously, at 12:00 you could
only book a unit for the next day at 12:00, but now it is possible to
book it for 00:00 next day.

Similar logic is already done in the frontend.


nicobav (approved):
Alrighty then :rocket:

PR #590, opened 2022-09-01
TILA-1821: Send reservation notifications

saaste:
There is a new helper that tries to figure out all the reservation notification recipients based on unit permissions and user-defined notification settings.

I changed the email sending function so that it takes recipients as an argument. If `None` is given (default) then the existing logic is used. I did it this way just to avoid breaking existing email sending logic.

Finally, I added a new ENV settings for configuring the maximum number of recipients just to avoid crazy email spam. The default is 100. If there are more than that, an exception is thrown and no emails are sent. This would be one of those exceptions I'd like to be part of monitoring since it should never happen 🤔 

Handler emails are sent only in specific cases that are described in the [ticket](https://helsinkisolutionoffice.atlassian.net/browse/TILA-1821).


api/graphql/reservations/reservation_serializers.py
```diff
@@ -677,10 +679,33 @@ def validated_data(self):
...
     def save(self, **kwargs):
         instance = super().save(**kwargs)
+
         if instance.state in RESERVATION_STATE_EMAIL_TYPE_MAP.keys():
             send_reservation_email_task.delay(
                 instance.id, RESERVATION_STATE_EMAIL_TYPE_MAP[instance.state]
             )
+
+        if instance.state == STATE_CHOICES.REQUIRES_HANDLING:
+            recipients = get_notification_recipients(
```
nicobav, line 689: I would add this whole block to be run by a task. In `get_notification_recipients` we're looping all users and doing checks to each of them so it might take too long for the client to wait for it.

email_notification/helpers.py
```diff
@@ -0,0 +1,38 @@
...
+def get_notification_recipients(
+    reservation: Reservation, notification_settings: List[ReservationNotification]
+) -> List[str]:
+    notification_recipients: List[User] = []
+    reservation_units = reservation.reservation_unit.all()
+    units = Unit.objects.filter(reservationunit__in=reservation_units)
+
+    # Get all users and collect the ones that can manage relevant units,
+    # have matching notification setting are not the reservation creator
+    users = User.objects.all()
```
nicobav, line 18: we could restrict the amount of user objects by filtering them by having some roles. Unit roles, service sector roles etc. and then we can go through "relevant" user instances.

email_notification/helpers.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_notification_recipients(
+    reservation: Reservation, notification_settings: List[ReservationNotification]
+) -> List[str]:
+    notification_recipients: List[User] = []
+    reservation_units = reservation.reservation_unit.all()
+    units = Unit.objects.filter(reservationunit__in=reservation_units)
+
+    # Get all users and collect the ones that can manage relevant units,
```
nicobav, line 16: We could make this comment a docstring `""" get all users.... """` and put this as a first line of the function.

email_notification/helpers.py
```diff
@@ -0,0 +1,38 @@
+from typing import List
+
+from permissions.helpers import has_unit_permission
+from reservations.models import Reservation
+from spaces.models import Unit
+from users.models import ReservationNotification, User
+
+
+def get_notification_recipients(
```
nicobav, line 9: Would  `get_staff_notification_recipients` be more descriptive name?

nicobav (changes requested):

api/graphql/reservations/reservation_serializers.py
saaste: Good idea! I created a new task for fetching recipients and sending emails 👍

email_notification/helpers.py
saaste: I was thinking about optimizing this, but because I'm such a mold head and didn't add a TODO comment, I forgot to do  it 🤦 

But now it is there! I didn't try to optimize the DB query to the max, but it should return much less users. Final checks are done in the code just like before.

email_notification/helpers.py
saaste: Changed 👍

email_notification/helpers.py
saaste: It would indeed! Renamed 👍

email_notification/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Get users with unit roles and notifications enabled, collect the ones that can manage relevant units,
+    have matching notification setting are not the reservation creator
+    """
+    notification_recipients: List[User] = []
+    reservation_units = reservation.reservation_unit.all()
+    units = Unit.objects.filter(reservationunit__in=reservation_units)
+
+    users = User.objects.filter(
+        Q(unit_roles__isnull=False) & ~Q(reservation_notification="NONE")
```
nicobav, line 23: service_sector_roles perhaps needs also? Or is that something that is "too high" in hierarchy and don't want mails?

email_notification/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    for user in users:
+        if (
+            has_unit_permission(user, units, "can_manage_reservations")
```
nicobav, line 27: And my above comment made me think about this one. This also restricts the service_sector_role users.

nicobav (approved):
I'll add an approval for convenience reasons. Please be aware of the comments below if they'll need some actions.

email_notification/helpers.py
saaste: It was specified so that notifications are sent only to people who have unit-level roles. From the ticket:
> jos oikeus on has_service_sector_permission kautta eli käyttäjällä on oikeudet toimipisteeseen sen toimialan kautta yleisesti, niin tällöin ilmoituksia ei siitä myöskään lähetetä (ellei lisäksi ole unit-tason permissio).

email_notification/helpers.py
saaste: Yep, same here. This is intentional ☺️

saaste:
I updated your existing language-specific email test to use BCC field instead of TO field. The new sender function uses BCC field every time, even when there is just one recipient. If you can see a problem with that, let's tune it ☺️

PR #592, opened 2022-09-01
TILA-1824 | EmailTemplate translation support 

nicobav:
Adds content and subject as translatable fields in `EmailTemplate` model.
If no content translations then we default to fi in email templates when compiling content.

TILA-1824


saaste (approved):
Looks smooooth 👍

nicobav:
> Looks smooooth +1

silky smooth?

PR #593, opened 2022-09-02
TILA-1820: Change currentUser query to return a single object

saaste:
Previously it returned an array which was a bit confusing. I think this makes more sense since it will always return max 1 user.


nicobav (approved):
niccce

PR #594, opened 2022-09-02
TILA-1859: Many-to-Many payment types

saaste:
Currently, reservation unit can have only one payment type. This PR changes that.

Existing `payment_type` field in the database is kept, but GQL does not use it anymore. Both create and update mutations support providing multiple payment types. Payment types do not use incremental IDs so that we can easily reference them in the code. Also, they are not editable in Django Admin.

The migration script takes the existing payment type and creates a new relation accordingly.

I have a feeling there might be some pieces of code that could be written more elegantly, so feel free to bombard me with comments 😄


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -258,12 +269,15 @@ class ReservationUnitCreateSerializer(ReservationUnitSerializer, PrimaryKeySeria
         ),
     )
 
-    payment_type = ChoiceCharField(
+    # Using ListField here because MultipleChoiceField is such
+    # a pain in the a** and I couldn't make it work
+    payment_types = serializers.ListField(
```
nicobav, line 274: Would it work to do this like: 
```
serializers.ListField(
        child=serializers.PrimaryKeyRelatedField(queryset=ReservationUnitPaymentType.objects.all()),
        source="payment_types",
        .....     
    )
```
Since now the payment type is a model and its primary key is `code`

api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -365,6 +379,15 @@ def validate(self, data):
                     "nameFi is required for draft reservation units."
                 )
 
+        if data.get("payment_types"):
```
nicobav, line 382: If this is still relevant I would make it as own method.

nicobav (changes requested):

api/graphql/reservation_units/reservation_unit_serializers.py
saaste: Seems to work! 👏

api/graphql/reservation_units/reservation_unit_serializers.py
saaste: It is not relevant from validation point of view so I removed it. However, I just realized how awful to provided error is:
```
{
  "data": {
    "createReservationUnit": {
      "errors": [
        {
          "messages": [
            "0"
          ]
        }
      ],
      "pk": null
    }
  }
}
```

It basically says the index of array where the invalid value is and nothing more. I think this is not good. I could restore the custom validation and make it as own method.

saaste: Actually, the build-in validation kicks in before `validate()` method is called 😢
Any wild ideas?

nicobav: Wild idea would be make own field inheriting the `ListField` and do some overwriting there :thinking:

saaste: It works! 🥳

api/graphql/validating_list_field.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def run_child_validation(self, data):
+        if self.child and self.child.choices:
+            allowed_values = list(self.child.get_choices().values())
+            for value in data:
+                if value not in allowed_values:
+                    raise ValidationError(
+                        f"{value} is not a valid value. Allowed values: {', '.join(allowed_values)}."
```
nicobav, line 21: Will this possibly print bunch of `pk`s that are existing and could be used? If so I think it might be a bit too much :thinking: 
We could check if the child is PrimaryKeyRelated field or some  other pk field and give different validation error message

saaste: I didn't find any clever way to make this happen so I chose to follow the dumb way. Since the number of available values can be any number, I just check if it is less than five. It it is, we show the options, otherwise we don't.

What do you think? 😄

nicobav: Sounds reasonable :ok_hand:

nicobav (approved):
Awesomeness :rocket:

PR #596, opened 2022-09-02
TILA-1836 | Unit rank

nicobav:
Adds rank field to `Unit` and also capability to sort by it to gql api.
Adds django-admin-sortable2 for making use of the rank field in reservation unit and unit in django admin.

TILA-1836


spaces/models.py
```diff
@@ -108,6 +108,12 @@ class Unit(models.Model):
     hauki_resource_id = models.CharField(
         verbose_name=_("Hauki resource id"), max_length=255, blank=True, null=True
     )
+    rank = models.PositiveIntegerField(
+        blank=True,
+        null=True,
```
saaste, line 113: Just wondering if it would make sense to make this non-nullable with default value of 0. It looks like `null` does not have any special meaning here and might just complicate front-end code if they have to deal with nulls.

From performance point-of-view it probably does not make a difference, unless if we use old version of PostgreSQL 🤪 

I leave it up to you ☺️

saaste (approved):
\*click\*...noice!

spaces/models.py
nicobav: I was thinking that if we order results with this `rank` then 1st 2st would be 1,2 so zeros 0 would be on top right? Those were my fri-yay thoughts :smile:   

But tuesday thoughts would be that  setting a value for those is needed if we end up using e.g `django-admin-sortable2` since it needs some values to those. But then they could be just the `pk`

saaste (approved):
Looks mighty fiiiine. Just fix formatting issue and this is good to go 👏

PR #600, opened 2022-09-07
TILA-1882 | Add filter for units to filter only with_permission

nicobav:
With this filter api results only the units which the has some permission through service sector roles or unit roles.

Refs TILA-1882


saaste (approved):
Good stuff!

PR #601, opened 2022-09-08
TILA-1883 | Only with unit permission resrvation unit filter

nicobav:
And also fixes some issues in the unit filter as well.

TILA-1882 TILA-1883


saaste (approved):
Very noice 👍

PR #602, opened 2022-09-08
TILA-1887 | Remove ReservationPurpose fk ref from ServiceSector and ReservationUnit

nicobav:
Seems that they're not needed anymore.

Refs TILA-1887


saaste (approved):
Cleaning day! 🧹

PR #603, opened 2022-09-09
TILA-1859: Document payment type codes

saaste:
For some reason the `help_text` argument in the custom `ValidatingListField` does not work. I tried to check it with a debugger. The value is clearly set, but it does not show in the GraphiQL.

As a quick fix, I moved the documentation under `PaymentTypeType`. There it works with `description` parameter.


nicobav (approved):
I think this is a good option :+1:

PR #604, opened 2022-09-15
TILA-1900 - Add createdAt to ReservationType

saaste:
The field is shown in the UI.


api/graphql/reservations/reservation_types.py
```diff
@@ -219,6 +221,11 @@ class Input:
 
     calendar_url = graphene.String()
 
+    def resolve_created_at(self, info: ResolveInfo) -> str:
+        if self is None:
+            return ""
+        return self.created_at.strftime("%Y-%m-%dT%H:%M:%S%z")
```
saaste, line 227: I don't know if this is really necessary. I just noticed that when `reservationByPk` is called, we will get in the typical `"%Y-%m-%dT%H:%M:%S%z"` format. However, when `reservations` is called, it includes milliseconds too.

I have no idea why this happens, so just to keep things consistent, I forced the formatting. If you know a better solution for this, please let me know ☺️

nicobav: Yup, its a bummer that these are to be handled this way :disappointed: I think this is the way. :+1:

nicobav (approved):

PR #606, opened 2022-09-15
TILA-1826 | Reservation statistics

nicobav:
Adds reservation statistic model and it to be created OR updated in every Reservation instance's save.

TILA-1826


nicobav:
I was having some weird problems with using signal and also for some reason the db visualization didn't update (except for the json)

saaste (approved):
Looks good! 👏

PR #607, opened 2022-09-16
TILA-1861: Add support for future and past pricings

saaste:
Because of the new requirements, this PR also separates pricing information from `ReservationUnit` models. Existing data is still there. I will remove them separately after we are confident this is the approach we want to go with.

The big challenge here is the fact that we wanted to modify pricing in the same mutation with `ReservationUnit`. Because units in draft state have very little requirements, this because surprisingly complicated to handled. Have to admit I'm not super proud of this because it adds a lot of manual validation we have to maintain. And I feel the code is confusing AF 💩 

Basically, this is how it works:
- Units in draft state don't have to define pricings. If they do, there can be only one ACTIVE and FUTURE pricings.
- Active pricings `begins` must be in the past or today
- Future pricings `begins` must be in the future
- Units in the non-draft state must have one ACTIVE pricing. FUTURE pricing is optional.
- When unit is updated, existing pricings can be modified and new ones added. Same rules apply.


api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -419,17 +490,131 @@ def validate_for_publish(self, data):
...
+                future_count += 1
+                if pricing.get("begins") <= date.today():
+                    raise serializers.ValidationError(
+                        "FUTURE pricing must be in the future"
+                    )
+            else:
+                other_count += 1
+
+        is_draft = data.get("is_draft", getattr(self.instance, "is_draft", False))
+        if (not is_draft and active_count != 1) or (is_draft and active_count > 1):
```
nicobav, line 524: Would it be sufficient to just check that the active_count is not zero? Can't figure out why we need multiple active pricings even when in draft state

So disregard the `is_draft` here and just iffing => `if active_count < 1: <kaboom>`

api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -419,17 +490,131 @@ def validate_for_publish(self, data):
...
             )
 
+        if not data.get("pricings", []):
+            raise serializers.ValidationError(
+                "pricings is required and must have one ACTIVE and one optional FUTURE pricing"
+            )
+
+    def validate_pricing_fields(self, data):
+        pricings = data.get("pricings", [])
+        if not pricings:
```
nicobav, line 500: Should we check that the pricings exists in some particular case?

api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -419,17 +490,131 @@ def validate_for_publish(self, data):
...
+            )
+        if other_count > 0:
+            raise serializers.ValidationError(
+                "only ACTIVE and FUTURE pricings can be mutated"
+            )
+
+    @staticmethod
+    def handle_pricings(pricings: List[Dict[Any, Any]], reservation_unit):
+        for pricing in pricings:
+            new_pricing = ReservationUnitPricing(
```
nicobav, line 540: You could just `ReservationUnitPricing.objects.create(**pricing, reservation_unit=reservation_unit)`

reservation_units/migrations/0072_separate_pricing_model.py
```diff
@@ -0,0 +1,54 @@
...
+                ('highest_price', models.DecimalField(decimal_places=2, default=0, help_text='Maximum price of the reservation unit', max_digits=10, verbose_name='Highest price')),
+                ('status', models.CharField(choices=[('past', 'past'), ('active', 'active'), ('future', 'future')], help_text='Status of the pricing', max_length=20, verbose_name='Status')),
+                ('reservation_unit', models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='pricings', to='reservation_units.reservationunit', verbose_name='Reservation unit')),
+                ('tax_percentage', models.ForeignKey(default=reservation_units.models.get_default_tax_percentage, help_text='The percentage of tax included in the price', on_delete=django.db.models.deletion.PROTECT, related_name='reservation_unit_pricings', to='reservation_units.taxpercentage', verbose_name='Tax percentage')),
+            ],
+        ),
+        migrations.RunPython(migrate_active_pricing, migrations.RunPython.noop),
+        migrations.RunSQL(migrations.RunSQL.noop, reverse_sql='SET CONSTRAINTS ALL IMMEDIATE'),
```
nicobav, line 53: Why this line?

reservation_units/models.py
```diff
@@ -634,6 +634,100 @@ def state(self) -> ReservationUnitState:
...
+    pricing_type = models.CharField(
+        max_length=20,
+        verbose_name=_("Pricing type"),
+        choices=PricingType.choices,
+        blank=True,
+        null=True,
+        help_text="What kind of pricing types are available with this reservation unit.",
+    )
+
+    PRICE_UNIT_PER_15_MINS = "per_15_mins"
```
nicobav, line 655: we could move all of these as separate TextChoices or enums

reservation_units/models.py
```diff
@@ -634,6 +634,100 @@ def state(self) -> ReservationUnitState:
...
+    tax_percentage = models.ForeignKey(
+        TaxPercentage,
+        verbose_name=_("Tax percentage"),
+        related_name="reservation_unit_pricings",
+        on_delete=models.PROTECT,
+        default=get_default_tax_percentage,
+        help_text="The percentage of tax included in the price",
+    )
+
+    PRICING_STATUS_PAST = "past"
```
nicobav, line 703: Samey, could move this as TextChoices or enum

reservation_units/tests/factories.py
```diff
@@ -109,6 +112,27 @@ class Meta:
...
+    lowest_price = 5
+    highest_price = 10
+    tax_percentage = SubFactory(TaxPercentageFactory, value=10.0)
+    status = "active"
+
+    class Meta:
+        model = "reservation_units.ReservationUnitPricing"
+
+    @post_generation
+    def reservation_unit(self, create, reservation_unit, **kwargs):
```
nicobav, line 129: No need for this since this `@post_generation` annotated is usually for adding m2m relationships.

api/graphql/reservation_units/reservation_unit_serializers.py
saaste: We can't do that because pricings are not required in the draft state. So when reservation unit is draft, `active_count` can be zero.

api/graphql/reservation_units/reservation_unit_serializers.py
saaste: It is checked in `validate_for_publish`, but I'll do some refactoring and move the validation code under `reservation_unit_pricing_helper`. Flake8 started to complain about method being too complex 🤦

api/graphql/reservation_units/reservation_unit_serializers.py
saaste: Ah, of course! 🤦  Fixed 👍

reservation_units/migrations/0072_separate_pricing_model.py
saaste: Ooops from copy-pasting code 😄 
I needed that in one of the previous change to allow smooth rollback but it is not needed here. Removed 👍

reservation_units/models.py
saaste: Done! I did this refactoring in a separate PR. Otherwise things started to look too confusing for my taste.

reservation_units/models.py
saaste: Done! I did this refactoring in a separate PR. Otherwise things started to look too confusing for my taste.

reservation_units/tests/factories.py
saaste: Removed 👍

nicobav (approved):

PR #608, opened 2022-09-16
TILA-1837 | Add new filters to Units query

nicobav:
Adds filters for querying units with published reservation units and units with own reservation (and the possibility to order them by the count)

TILA-1837


saaste (approved):
Veeeery nice 👍

PR #609, opened 2022-09-20
TILA-1864: Task for scheduled pricing update

saaste:
Adds a task which we can run once a day to update reservation unit pricings. Currently based on #607 so I'll update the PR after #607 is merged.


nicobav (approved):
superb

PR #612, opened 2022-09-21
TILA-1927 | Remove ReservationPermission check from ReservationUnit type's reservations

nicobav:
In ReservationType there is field based field checks that checks should the field be visible or nulled so the reservation check can and should be removed since reservations are to be viewable without logged in

Refs TILA-1927


saaste (approved):
Great to see that hiding sensitive fields works without any extra work 🥳

PR #615, opened 2022-09-22
TILA-1894 | Make email notification builder to respect timezones

nicobav:
Previously the email notification builder was getting times and dates as is withouth taking into consideration the timezone. This changes that so it takes the timezone into account.

Refs TILA-1894


saaste (approved):
Noice!

PR #617, opened 2022-09-23
TILA-1841: Add reservation unit purpose image

saaste:
Uploading is done via Django Admin so there is no mutations for images.


api/graphql/tests/test_purposes.py
```diff
@@ -111,7 +114,21 @@ def test_normal_user_cannot_update(self):
 class PurposeQueryTestCase(GrapheneTestCaseBase, snapshottest.TestCase):
     @classmethod
     def setUpTestData(cls):
-        cls.purpose = PurposeFactory(name_fi="fi", name_en="en", name_sv="sv")
+        bytes = bytearray()
```
saaste, line 117: Do you happen to know any better way to test image uploads and thumbnail generation?

Playing with real files and doing file cleanup in the `tearDown` feels a bit wrong and dangerous 😛

nicobav: If you really want you could try to use Mock. 
The image upload itself is handled by django and we can trust it to work - same basically is with the thumbnail generation. But what we might would like to test is that the image urls are formed as should but that's almost 99% handled by django/graphene/easythumbnailer so.... :cowboy_hat_face:

saaste: Yeah that is a valid point. This tests the libraries we are using and not our code, really. I think I just undo my test changes. It does not add any real value but adds headaches. That is a bad combination 🤠

nicobav (approved):
Noice :+1:

PR #619, opened 2022-09-27
TILA-1154 & TILA-1920: Payment merchant model and API client

saaste:
This PR adds two things:

### Merchant Experience API client
There is now client code for [Merchant Experience API](https://checkout-test-api.test.hel.ninja/v1/merchant/docs/swagger-ui/). It supports fetching, creating and updating merchants.

~~There are some issues in the Merchant API. I added TODO comments for the changes we need to do after API has been updated. Related ticket: [KYV-747](https://helsinkisolutionoffice.atlassian.net/browse/KYV-747)~~

### Payment Merchant Model
The only things we store on our end are ID and name. Everything else is stored in the Merchant Experience API. However, we want to modify the data via Django Admin so I created a separate admin form that handles API calls.

Also
Refs TILA-1920


verkkokauppa/tests/test_merchant_requests.py
```diff
@@ -0,0 +1,216 @@
...
+from ..merchants.requests import (
+    create_merchant,
+    get_merchant,
+    get_merchants,
+    update_merchant,
+)
+from ..merchants.types import Merchant, MerchantInfo
+from .mocks import mock_get, mock_post
+
+# Create merchant
```
nicobav, line 24: If we used test classes we could make CreateMerchantTestCase and wouldn't need to comment :wink:

verkkokauppa/tests/test_merchant_requests.py
```diff
@@ -0,0 +1,216 @@
...
+def test_create_merchant_raises_exception_if_status_code_is_not_201(
+    create_merchant_params, mutation_merchant_response
+):
+    post = mock_post(mutation_merchant_response, status_code=500)
+    assert_that(create_merchant).raises(CreateMerchantError).when_called_with(
+        create_merchant_params, post
+    )
+
+
+# Update merchant
```
nicobav, line 81: similarly here :monocle_face:

nicobav (approved):
Very nice :ok_hand: 

Tests could be written as TestCases but it is surely fine like this also :+1:

verkkokauppa/tests/test_merchant_requests.py
saaste: I changed tests to TestClasses 👍

PR #620, opened 2022-09-28
TILA-1518: Bump ortools to latest version

saaste:
Older ortools has a protobuf dependency which has a [known vulnerability](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/security/dependabot/9).


nicobav (approved):
Yes please :pray:

PR #611, opened 2022-09-21
TILA-1840: Add reservation unit purpose rank

saaste:
Default ordering stays the same. This just adds a possibility to sort the purposes in the admin UI and in the query.

PR #621, opened 2022-09-29
TILA-1840: Add missing db migration

saaste:
I just realized, that I forgot to create DB migrations in #611 and `django migrate` was complaining.
So, fixing my own 💩


nicobav (approved):
:broom:

PR #622, opened 2022-09-29
TILA-1921: Add merchant to Unit and Reservation Unit

saaste:
In the GQL API merchant is visible only if user has manage permissions to Unit/Reservation Unit. In case of Reservation Unit, merchant is returned from the Reservation Unit, and if that is not available merchant from Unit is used.


reservation_units/models.py
```diff
@@ -541,6 +542,15 @@ class ReservationUnit(models.Model):
         db_index=True,
     )
 
+    payment_merchant = models.ForeignKey(
+        PaymentMerchant,
+        verbose_name=_("Payment merchant"),
+        related_name="reservation_units",
+        on_delete=models.PROTECT,
```
nicobav, line 549: We cannot null the payment merchant in reservation unit when its deleted?

nicobav (approved):
Solid :rock:

reservation_units/models.py
saaste: That is a good question. To be honest, I don't know.

Deleting Merchants is kinda tricky because the API does not have a delete endpoint. So, even if we delete it from our end, it won't be deleted from the API. I thought it is safer to at least prevent the deletion if merchant is attached to Reservation Unit or Unit just to avoid weird orphans. Have to check if there is some reasons to prevent deletion of merchants even more. Something like purchase history or stuff like that .

I'll check this from Juhis.

PR #623, opened 2022-09-30
Add reservations to reservationUnitByPk

nicobav:
this was missing from there.

(someday I hope we refactor away of these ByPks)


saaste (changes requested):
Tests are failing and this needs some rebasing, but other than that this looks great! 👍

saaste:
I fixed the test by removing ID from the response because that'll ruin snap tests. This is needed by frontend guys so I'll merge this with my own permission 😄

saaste (approved):

PR #624, opened 2022-09-30
TILA-1909: Add PaymentProduct (and ref to Reservation Unit)

saaste:
PaymentProduct is also one of those models that are actually created in [Product Experience API](https://checkout-test-api.test.hel.ninja/v1/product/docs/swagger-ui/#/). For these, there is no Admin UI. The data will be created via API when Reservation Unit is mutation. This will be implemented as a separate ticket so the current code is pretty much just a preparation for it.


merchants/models.py
```diff
@@ -26,3 +26,24 @@ def save(
 
     def __str__(self) -> str:
         return self.name
+
+
+class PaymentProduct(models.Model):
```
nicobav, line 31: I just realized, when I started to think what this `PaymentProduct` means, that perhaps we should refactor a bit and move this `merchant` app to `verkkokauppa` app. What do you think? Or we can get ridoff the `verkkokauppa` app :thinking: but having the two is getting perhaps a bit confusing.

In the `verkkokauppa` we have data model of `Product` (which is clearly outdated) and it sounds pretty much what this represents => obviously without being a django model => written to database but rather its read from the source.

saaste: Yeah, my naming for the app wasn't very good. I have to check how easy it is to change the name of the app in Django ☺️ 

What we currently have under `verkkokauppa` is not Django app at all. Basically, it is just a package with files related to Verkkokauppa API client. So basically `PaymentProduct` is our Django model and `Product` is Verkkokauppa API model. They are related but two separate things.

That being said, I think these two could be combined. All verkkokauppa related could be under one app and then we could have `client` subfolder for Verkkokauppa API related stuff.

I'll investigate what kind of mess renaming does. Never done that before 😄

saaste: Uff, I tried to figure out an easy way to rename an app in Django. Turns out it is such as pain in the 🍑 if you already have migrations with relations. So, perhaps we just have to live with my bad decisions and keep the `merchants` app.

Instead of renaming the app, I moved `verkkokauppa` folder to `merchants` so that everything related to verkkokauppa is in one place. We still have two similarly named classes:
- `PaymentMerchant` and `PaymentProduct`: our own Django models with fewer fields
- `Merchant` and `Product`: Verkkokauppa API models

nicobav (approved):
:shipit:

PR #626, opened 2022-10-04
TILA-1921: Fix merchant save

saaste:
Oops. I didn't do a proper check for merchant creation. Apparently instance is always there.


saaste:
#627 replace this one.

PR #627, opened 2022-10-05
TILA-1937: Add merchant shop and fix saving issues

saaste:
Merchant Experience API added a new required field `merchantShopId`. Added that to our model.

Also, I fixed issues we had in saving a merchant. Now reservation unit does not require merchant field to be filled. This will help frontend devs to continue their work even there is something wrong with the verkkokauppa integration part.


nicobav (approved):
:rocket:

PR #629, opened 2022-10-05
TILA-1916 | Read user's date of birth

nicobav:
Adds date of birth reading / determination in token authentication's resolve user method:

Changes.
 - Adds `date_of_birth` to `User` model.
 - Enables x-authorize headers
 - Overrides helusers resolve_user method
 - Reads user data from profile if user auth level (loa) is in configured values


users/utils/birthday_resolver.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def resolve_user(request, payload):
+    user = get_or_create_user(payload, oidc=True)
+
+    # If auth method is in configured loa levels we can try to read the date of birth of the user.
+    if (
+        payload.get("loa", "").lower()
+        in settings.OPEN_CITY_PROFILE_LEVELS_OF_ASSURANCES
+        and not user.date_of_birth
```
saaste, line 18: Is there any case where user's birthday can be changed? Like, is it possible that user entered a wrong birthday, realises it later and then changes it. Or is this always coming via strong authentication.

Just thinking, because with the current code, birthday is only added but never updated.

users/utils/birthday_resolver.py
```diff
@@ -0,0 +1,136 @@
...
+    def __get_token(self):
+        token = self.request.headers.get("X-Authorization", b"")
+        if isinstance(token, str):
+            token = token.encode(HTTP_HEADER_ENCODING)
+        return token
+
+    def get_user_birthday(self) -> [datetime.date, None]:
+        nin = self.__get_national_identification_number()
+        if not nin or len(nin) != 11:
+            return None
```
saaste, line 74: I'm wondering would be be a easy safety net for use log these cases 🤔 

Having an invalid length ID number and century look like corner cases we should never have, but if profile goes haywire, we just start to return `None`s without any information what is causing it. Unlikely, but if it happens, it could be a nightmare to debug and logs could help.

saaste (approved):
Looks great! Just two questions but I leave it up to you to decide if these are something that require code changes.

If I understood it correctly, user's national identification number is transferred in plain-text JWT token in headers. I really hope we use HTTPS everywhere because that feels slightly scary since the content of the JWT can be decoded very easily. Anyone who can get access to the token can get it ID number as well 🤔

users/utils/birthday_resolver.py
nicobav: We should use the loa levels of high and substantial and we should be able to trust those ones. Of course if/when we are using settings with low to then we need to do some updates to code or provide some other way to update the birthday if necessary. Or contact to support :upside_down_face: 
So shortly; for now were using high `loa`s and trust'em and in future this feature might be enhanced.

users/utils/birthday_resolver.py
nicobav: Good point; we could add a warning message to sentry? Whatya think?

users/utils/birthday_resolver.py
vergama: Somewhat esoteric but maybe possible scenario: foreign born person has ID assigned on 1st of January, then later gets new ID at which point actual date of birth is known and the new ID reflects that.

users/utils/birthday_resolver.py
saaste: Sounds great! 👍

users/utils/birthday_resolver.py
nicobav: Would the person be interpret as new user in profiili then?

PR #631, opened 2022-10-07
TILA-1910: Reservation unit product mapping

saaste:
When reservation unit is saved, we will call Product Experience API
and create a new product mapping. API is clever enough to return
existing mapping if it exists so we can call create every time.

Mapping is created/updated only if reservation unit is paid and has a
payment merchant defined. If merchant is removed, mapping is removed
too.

All this is is done as a Celery task

NOT TO SELF: This requires deployment configuration changes 🧠


merchants/verkkokauppa/tests/test_product_requests.py
```diff
@@ -1,60 +1,144 @@
...
-    post = mock_post({})
-    with suppress(Exception):
-        create_product(create_product_params, post)
-    post.assert_called_with(
-        url=settings.VERKKOKAUPPA_PRODUCT_API_URL,
-        json=create_product_params.to_json(),
-        headers={"api-key": settings.VERKKOKAUPPA_API_KEY},
-        timeout=REQUEST_TIMEOUT_SECONDS,
-    )
+class ProductRequestsBaseTestCase(TestCase):
```
nicobav, line 19: totally nitpicking :nerd_face: => would consider renaming to `ProductRequestsTestCaseBase` as its the base and not TestCase

api/graphql/tests/snapshots/snap_test_reservation_units.py
```diff
@@ -1775,7 +1775,7 @@
                         'nameFi': 'test name fi',
                         'paymentProduct': {
                             'merchantPk': '3828ac38-3e26-4501-8556-ba2ea3442627',
-                            'pk': '1018cabd-d693-41c1-8ddc-dc5c08829048'
+                            'pk': 'cf9798b2-c5c1-4bdc-877a-47722dc39f9e'
```
nicobav, line 1778: so this is coming from the mocked response? Its ok :innocent:

reservation_units/models.py
```diff
@@ -657,6 +659,40 @@ def state(self) -> ReservationUnitState:
...
+
+    def save(self, *args, **kwargs):
+        # Update product mapping if runit is not draft and is paid
+        if self._requires_product_mapping_update():
+            params = CreateProductParams(
+                namespace=settings.VERKKOKAUPPA_MERCHANT_NAMESPACE,
+                namespace_entity_id=self.pk,
+                merchant_id=self.payment_merchant.id,
+            )
+            api_product = create_product(params)
```
nicobav, line 684: this was the part that is to be run in a task?

merchants/verkkokauppa/tests/test_product_requests.py
saaste: Nitpick approved and class renamed 😄

api/graphql/tests/snapshots/snap_test_reservation_units.py
saaste: Yep, but I updated the tests and used the old UUID in the mock just to avoid pointless changes. I also managed to remove few extra lines from the test file 🧹

reservation_units/models.py
saaste: Exactly! I already have the code ready. I just need to update the tests 👍

reservation_units/models.py
```diff
@@ -657,6 +658,20 @@ def state(self) -> ReservationUnitState:
 
         return ReservationUnitStateHelper.get_state(self)
 
+    def save(
+        self,
+        force_insert=False,
+        force_update=False,
+        using=None,
+        update_fields=None,
+        run_product_refresh=True,
```
saaste, line 667: This is a bit of a hack to prevent infinite loop. `refresh_reservation_unit_product_mapping` task calls `save()` when product mapping is updated so I have to have a way to make sure the task is not triggered again.

nicobav: Also you could use queryset.update() => I'll add suggestion in proper place

reservation_units/tasks.py
```diff
@@ -14,3 +21,41 @@ def _update_reservation_unit_pricings() -> None:
...
+        payment_product, _ = PaymentProduct.objects.update_or_create(
+            id=api_product.product_id, defaults={"merchant": payment_merchant}
+        )
+        reservation_unit.payment_product = payment_product
+
+    # Remove product mapping if merchant is removed
+    if reservation_unit.payment_product is not None and payment_merchant is None:
+        reservation_unit.payment_product = None
+
+    reservation_unit.save(run_product_refresh=False)
```
nicobav, line 61: `ReservationUnit.objects.filter(id=reservation_unit.id).update(payment_product=payment_product)`
Wouldn't call instance's `save()` method and then the `run_product_refresh` could be ditched => but either way is proper  I guess

saaste: Awwwwyeessss! So much better 👏

nicobav (approved):
Nicey :sunglasses: 

left just one tiny miney comment and you can decide whether to act on it or not.

reservation_units/models.py
```diff
@@ -657,6 +658,19 @@ def state(self) -> ReservationUnitState:
 
         return ReservationUnitStateHelper.get_state(self)
 
+    def save(
```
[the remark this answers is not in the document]
saaste: Yep, I'll change this. Forgot to change this back after I removed the `update_product_mapping` argument.

reservation_units/models.py
```diff
@@ -657,6 +658,19 @@ def state(self) -> ReservationUnitState:
...
+        self,
+        force_insert=False,
+        force_update=False,
+        using=None,
+        update_fields=None,
+    ) -> None:
+        super(ReservationUnit, self).save(
```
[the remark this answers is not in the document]
saaste: Fixed! 👍

saaste:
I'll merge this now. There is a related pipeline update PR waiting on the devops side ☺️

PR #632, opened 2022-10-10
TILA-1961: Make unit merchant optional

saaste:
Fixing my own ooops 🙈 
Based on #631. Merge that first.


api/graphql/tests/snapshots/snap_test_reservation_units.py
```diff
@@ -1775,7 +1775,7 @@
                         'nameFi': 'test name fi',
                         'paymentProduct': {
                             'merchantPk': '3828ac38-3e26-4501-8556-ba2ea3442627',
-                            'pk': '1018cabd-d693-41c1-8ddc-dc5c08829048'
+                            'pk': 'cf9798b2-c5c1-4bdc-877a-47722dc39f9e'
```
nicobav, line 1778: :thinking:

nicobav (approved):

api/graphql/tests/snapshots/snap_test_reservation_units.py
saaste: I happened to use a different hard-coded UUID in my mock compared to the previous way the tests were set up. I updated to UUID in the mock and also removed some extra lines from the test that weren't really needed ☺️

PR #634, opened 2022-10-11
Add default order to be "rank" on ReservationUnit

nicobav:
the sortable library in django admin ui does not sort by default without this.


saaste (approved):
🚢  it!

PR #639, opened 2022-10-14
TILA-1944 | Remove leading zeros from reservation number in emails

nicobav:
This is the way.

Refs TILA-1944

### DIFF
--- a/email_notification/tests/test_notification_builders.py
+++ b/email_notification/tests/test_notification_builders.py
@@ -79,8 +79,8 @@ class ReservationEmailNotificationBuilderTestCase(ReservationEmailBaseTestCase):
         assert_that(self.builder._get_end_date()).is_equal_to("01.03.2022")
 
     def test_get_reservation_number(self):
-        resno = str(self.reservation.id).zfill(10)
-        assert_that(self.builder._get_reservation_number()).is_equal_to(f"{resno}")
+        resno = self.reservation.id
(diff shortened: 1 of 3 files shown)


email_notification/tests/test_notification_builders.py
```diff
@@ -227,9 +227,7 @@ def test_get_content(self):
     def test_padding_does_not_matter_in_curly_brackets(self):
         EmailTemplate.objects.all().delete()
         content_with_padding = "I have {{ reservation_number }} padding"
-        compiled_content_with_padding = (
-            f"I have {str(self.reservation.id).zfill(10)} padding"
-        )
+        compiled_content_with_padding = f"I have {self.reservation.id} padding"
```
saaste, line 230: Possible nitpick: What `padding` means here? If it is related to leading zeros, some variable renaming could be done here since zeroes are now gone ☺️

saaste (approved):
Approved with one nitpick comment. If it is just my misunderstanding, this is goooood to go 👍

email_notification/tests/test_notification_builders.py
nicobav: yup, this is quite obfuscating name; it means that in upper:
`content_with_padding = "I have {{ reservation_number }} padding"`
has a padding in the curly brackets and the we test that it isn't affecting inside of curlys... Perhas I'll change the variable name  to `content_with_space_inside_braces` and `compiled_content_with_space_inside_braces` ? :thinking:

PR #640, opened 2022-10-17
TILA-1939: Add a new permission for commenting reservations

saaste:
Also updated the description of `can_view_reservation` permission.


permissions/models.py
```diff
@@ -56,7 +56,7 @@
     ),
     (
         "can_view_reservations",
-        _("Can create, edit and delete equipment in the whole system"),
+        _("Can view equipments in the whole system"),
```
nicobav, line 59: "Can view reservations in the whole system"

permissions/models.py
```diff
@@ -87,6 +87,10 @@
         "can_view_users",
         _("Can view users in the whole system"),
     ),
+    (
+        "can_comment_reservations",
+        _("Can comment reservations"),
```
nicobav, line 92: perhapsies we could add the "whole system" line to here as well to be consistent and making it clear.

permissions/api_permissions/graphene_permissions.py
```diff
@@ -353,6 +354,18 @@ def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> b
         return False
 
 
+class ReservationCommentPermission(BasePermission):
+    @classmethod
+    def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> bool:
+        pk = input.get("pk")
+        if pk:
+            reservation = get_object_or_404(Reservation, pk=pk)
+            return can_handle_reservation(
```
nicobav, line 363: Should we add the can_handle helper func in the can_comment helper? Wouldn't need to combine here but would be just a straight helper function what to use here.

nicobav (changes requested):

permissions/models.py
saaste: Fixed 🤦

permissions/models.py
saaste: Added 👍

permissions/api_permissions/graphene_permissions.py
saaste: Good idea! I added the helper function and simplified this ✅

nicobav (approved):
Nicey :cookie:

PR #641, opened 2022-10-18
TILA-1917 | Remove reserveeDateOfBirth and add gql navigation to User from Reservation

nicobav:
Much more usable this way

Refs TILA-1917


saaste (approved):
Indeed, more usable ☺️ 👍

PR #642, opened 2022-10-19
TILA-1988: Fix product mapping removal bug

saaste:
Well that was a stupid mistake from me 🤦‍♂️

### DIFF
--- a/reservation_units/tasks.py
+++ b/reservation_units/tasks.py
@@ -59,7 +59,7 @@ def refresh_reservation_unit_product_mapping(reservation_unit_pk) -> None:
         )
 
     # Remove product mapping if merchant is removed
-    if reservation_unit.payment_product is not None and payment_merchant is None:
+    if reservation_unit.payment_product and not payment_merchant:
         ReservationUnit.objects.filter(pk=reservation_unit_pk).update(
-            payment_product=payment_product
+            payment_product=None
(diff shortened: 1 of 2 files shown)


reservation_units/tasks.py
```diff
@@ -61,5 +61,5 @@ def refresh_reservation_unit_product_mapping(reservation_unit_pk) -> None:
     # Remove product mapping if merchant is removed
     if reservation_unit.payment_product is not None and payment_merchant is None:
```
nicobav, line 62: I would suggest to change this `if` condition to this to:
`if reservation_unit.payment_product and not payment_merchant:`

obviously in every case `is not None` does not equal plain `if <something>` but in this case it will (and for the most cases).

nicobav: I know its not part the changeset but while you at it :innocent:

nicobav (approved):
Changes are 5/5 :+1: 

Added one optional fine-tuning while we're at this.

reservation_units/tasks.py
saaste: Sure, I'll change this! ☺️ 

I'm still a bit confused when in Python it is OK to check truthy and when it is recommended to explicitly check None value 🤔 I've seen some mixed guidelines 🤪

PR #648, opened 2022-10-20
TILA-1950 | Opening hours client enhancements

nicobav:
## Change log
- Adds new resource states to State enum in openinhours.enum; maintenance, not_in_use
- Add util method to State enum to return states that are to be considered when the reservations are possible
- Make the opening hours client know the usage of `end_time_on_next_day` value
- Make the opening hours client interpret resource to be reservable whole day when start_time and end_time are both nulls.

### DIFF
--- a/opening_hours/utils/opening_hours_client.py
+++ b/opening_hours/utils/opening_hours_client.py
@@ -6,12 +6,9 @@ from django.conf import settings
 from django.utils.timezone import get_default_timezone
 
 from opening_hours.decorators import datetime_args_to_default_timezone
-from opening_hours.hours import (
-    Period,
-    TimeElement,
-    get_opening_hours,
-    get_periods_for_resource,
-)
+from opening_hours.hours import Period
+from opening_hours.hours import State as ResourceState
+from opening_hours.hours import TimeElement, get_opening_hours, get_periods_for_resource
 
 TIMEZONE = get_default_timezone()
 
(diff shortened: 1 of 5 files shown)


opening_hours/utils/opening_hours_client.py
```diff
@@ -164,12 +168,20 @@ def get_resource_periods(self, resource) -> List[Period]:
...
-        for time in times:
-            if time.start_time <= start_time and time.end_time >= end_time:
+        for time in [
+            time
+            for time in times
+            if ResourceState(time.resource_state) in ResourceState.reservable_states()
+        ]:
+            if (
+                time.start_time <= start_time
+                and (time.end_time >= end_time or time.end_time_on_next_day)
```
saaste, line 182: Hmm, I wonder if the first two lines should be in parenthesis 🤔 The first and last condition are both checking `time.start_time` but on the latter it is a falsy check and not comparison to some other value. So something like this:
```
(
    time.start_time <= start_time
    and (time.end_time >= end_time or time.end_time_on_next_day)
)
or (not time.start_time and not time.end_time)
```

Or maybe even other way round just to make the falsy check first 🤔

saaste (changes requested):
Looks good! Just one question to make sure the condition check is correct. Also, tests seem to be failing.

opening_hours/utils/opening_hours_client.py
nicobav: nice ones :+1:  I'll refactor

nicobav: Added the falsy check as variable to be more readable

saaste (approved):
Splendid! :shipit:

PR #649, opened 2022-10-21
TILA-1967: Add Order

saaste:
## Change log
- Added `PaymentOrder` model
- Added `PaymentOrderItem` model
- Added tests for extra validation
- Fixed the flaky test, again (seemed to be case-sensitivity issue)
- Updated snapshot to get rid of the out-of-date warning

## Other notes
Extra validation is implemented in the model because these will be created as a result of Verkkokauppa
API calls instead of user requests.


merchants/models.py
```diff
@@ -47,3 +51,242 @@ class PaymentProduct(models.Model):
...
+class Language(models.TextChoices):
+    FI = "fi", _("Finnish")
+    SV = "sv", _("Swedish")
+    EN = "en", _("English")
+
+
+class PaymentOrder(models.Model):
+    order_id = models.UUIDField(
+        verbose_name=_("Order ID"),
+        help_text=_("Talpa eCommerce order ID"),
```
nicobav, line 80: This could be more common model for other eCommerce platforms so I would get rid off "Talpa"

merchants/models.py
```diff
@@ -47,3 +51,242 @@ class PaymentProduct(models.Model):
...
+    )
+
+    price_total = models.DecimalField(
+        verbose_name=_("Total amount"),
+        max_digits=10,
+        decimal_places=2,
+    )
+
+    created_at = models.DateTimeField(
+        verbose_name=_("Created at"), null=False, default=timezone.now
```
nicobav, line 124: We could remove default= and add `auto_now_add=True` It is the way.
Then if you want to have some `updated_at` it would need `auto_now=True`

nicobav (commented):
It seems that we're saving a lot of info regarding orders. I just wonder if they're something that we must maintain regularly and even need to have in our end?

merchants/models.py
saaste: Removed 👍

merchants/models.py
saaste: Changed 👍

saaste:
> It seems that we're saving a lot of info regarding orders. I just wonder if they're something that we must maintain regularly and even need to have in our end?

That was a very valid point. I had a discussion with Juhis and I also checked what fields we actually need and removed most of them. We still need some fields, but the main source of data should be Verkkokauppa API.

nicobav (approved):
<3

PR #651, opened 2022-10-21
TILA-1990 | Navigate to user from application

nicobav:
## Change log
- Adds ability to navigate to user from applicaiton. ApplicationType has userType.
- Remove permission check from UserType's date_of_birth field. The permission should be checked in application and reservation level when navigating to user.

Refs TILA-1990


saaste (approved):
Noice! 👍

PR #652, opened 2022-10-26
TILA-1986 | Add permissions to roles in userType

nicobav:
## Change log
- Adds permissions to role types in gql api
- Means that in apis using userType the user can query permissions

Refs TILA-1986


saaste (approved):
[Okily-Dokily-Doo!](https://www.youtube.com/watch?v=EzO-1FreLvw)

PR #654, opened 2022-10-28
TILA-2005 | Make get_opening_hours to read null start and end times

nicobav:
Previously it failed to do so since the fromisoformat fails. Fixes that to checking the data before parsing.

Refs TILA-2005


saaste (approved):
Nice catch! 🛠️

PR #656, opened 2022-11-02
TILA-2025 | Make changing to non-mandatory fields in metadata

nicobav:
In reservation update we check the metadatafields but there is some exceptions for the required fields. As two of those exceptions for individual persons are:
 - home_city is not required field even marked in metadataset
 - organisation_name is not required field even marked in metadataset

Refs TILA-2025


saaste (approved):
Looks good! Chipmunk time :shipit:

nicobav:
I just love chipmunks!

PR #657, opened 2022-11-03
TILA-1963 | Alter permission check for space mutations

nicobav:
Previously we checked only for unit_pk the unit level permissions when doing unit mutations. This faults in most cases to general permission check and thus the update/create wouldn't work for unit admins not service sector admins.

This changes the permission check logic so that it can handle update/delete/creation for unit admins => checks the permissions from the space itself (space.unit)

Refs TILA-1963


saaste (approved):
Superb! 👍

PR #658, opened 2022-11-04
TILA-1975 | Round resrevation prices using Decimal class quantize

nicobav:
This was originally introduced using built-in round which does not comply the rounding specs.

Refs TILA-1975

### DIFF
--- a/reservation_units/utils/reservation_unit_pricing_helper.py
+++ b/reservation_units/utils/reservation_unit_pricing_helper.py
@@ -1,3 +1,4 @@
+import decimal
 from datetime import date, datetime
 from decimal import Decimal
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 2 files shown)


reservation_units/utils/reservation_unit_pricing_helper.py
```diff
@@ -115,14 +116,12 @@ def calculate_vat_prices(cls, data: Dict[str, Any]) -> Dict[str, Any]:
...
-                    Decimal(highest_price_net * (1 + tax_percentage.decimal)),
-                    2,
-                )
-                pricing["lowest_price"] = round(
-                    Decimal(lowest_price_net * (1 + tax_percentage.decimal)),
-                    2,
-                )
+                pricing["highest_price"] = Decimal(
+                    highest_price_net * (1 + tax_percentage.decimal)
+                ).quantize(Decimal("1.0"), rounding=decimal.ROUND_HALF_UP)
```
saaste, line 121: Original code had two decimals so I think this should be:
```suggestion
                ).quantize(Decimal("1.00"), rounding=decimal.ROUND_HALF_UP)
```

Same with the next `quantize`

saaste (changes requested):

saaste (approved):
Fixers gonna fix! :shipit:

PR #660, opened 2022-11-04
TILA-2017 | Non subsidised prices

nicobav:
## Change log
- Adds `non_subsidised_price` and `non_subsidised_price_net` model fields to reservation unit
- Stores reservation prices to those fields when doing an update or create mutations. 
- Adds some tests also for net prices. (Only asserts to places having prices)

Refs TILA-2017


api/graphql/tests/test_reservations/test_reservation_create.py
```diff
@@ -1091,12 +1091,16 @@ def test_create_price_calculation_with_free_reservation_unit(
         pk = content.get("data").get("createReservation").get("reservation").get("pk")
         reservation = Reservation.objects.get(id=pk)
         assert_that(reservation).is_not_none()
-        assert_that(reservation.price).is_equal_to(
-            0.0
-        )  # Free units should always be 0 €
+        assert_that(reservation.price).is_zero()  # Free units should always be 0 €
+        assert_that(reservation.non_subsidised_price).is_equal_to(
+            reservation.non_subsidised_price
```
saaste, line 1096: I believe this should be `price`. Otherwise you compare `non_subsidised_price` to itself ☺️ 
```suggestion
            reservation.price
```

saaste (changes requested):

api/graphql/tests/test_reservations/test_reservation_create.py
nicobav: solv'd

saaste (approved):
Aaaaaamazing 👍

PR #661, opened 2022-11-08
TILA-1974: Create order and call Verkkokauppa API on reservation confirm

saaste:
## Change log
- Add support for staff to define payment type on confirm
- Make sure product is defined on confirm
- Store local order with minimal information on confirm
- Add reservation reference to order
- Add user UUID reference to order
- Call verkkokauppa API and create a proper order
- Prevent confirmation if order already exists
- Add a new reservation state: WAITING_FOR_ORDEr

## Other notes
This is a way too big change that should have been multiple smaller PRs. The ticket itself is relatively simple at first, but as it turned out, it was nothing but simple. I tried to separate changes with commits. There is a little bit code churn because I added TODO-comments for future changes and then I changed them in the later commit. Made my life a little bit easier and that way I avoided a brain malfunction.

Shame on me

## Deployment reminder
- No deployment changes needed


api/graphql/reservations/reservation_serializers.py
```diff
@@ -717,34 +721,142 @@ def check_metadata_fields(self, data, reservation_unit) -> None:
...
+                    price_net=price_net,
+                    price_vat=price_vat,
+                    price_total=price_total,
+                    reservation=self.instance,
+                    reservation_user_uuid=self.instance.user.uuid,
+                    order_id=payment_order.order_id,
+                    checkout_url=payment_order.checkout_url,
+                    receipt_url=payment_order.receipt_url,
+                )
+                Reservation.objects.filter(pk=self.instance.pk).update(
```
saaste, line 855: This is the part I wanted to ask you about. The point is to update reservation state AFTER we've created the order.

I moved the `instance = super().save(**kwargs)` up and this whole order mayhen happens after that. If everything goes well, we'll update the state again.

Can you think any better way to do this?

api/graphql/reservations/reservation_serializers.py
```diff
@@ -717,34 +721,142 @@ def check_metadata_fields(self, data, reservation_unit) -> None:
...
+        state = self.validated_data["state"]
+        if state == STATE_CHOICES.CONFIRMED:
+            payment_type = self.validated_data["payment_type"].upper()
+            price_net = round_decimal(self.instance.price_net, 2)
+            price_vat = round_decimal(
+                self.instance.price_net * (self.instance.tax_percentage_value / 100), 2
+            )
+            price_total = round_decimal(self.instance.price, 2)
+
+            if payment_type == PaymentType.ON_SITE:
```
nicobav, line 830: These order creations could be inside the payments app, could they? 🤔

merchants/verkkokauppa/order/requests.py
```diff
@@ -15,17 +15,19 @@
...
-            url=settings.VERKKOKAUPPA_ORDER_API_URL,
+            url=settings.VERKKOKAUPPA_ORDER_API_URL + "/",
             json=params.to_json(),
             headers={"api-key": settings.VERKKOKAUPPA_API_KEY},
             timeout=REQUEST_TIMEOUT_SECONDS,
         )
-        json = response.json()
+        json = (
+            response.json()
+        )  # TODO: In case of 501 there is no JSON response. Check response headers
```
nicobav, line 25: could a ticket numer

nicobav (changes requested):
Nicey nicer 👏 

Had some thoughts on mainly the structure of the code 🥯

tilavarauspalvelu/utils/date_util.py
```diff
@@ -29,3 +29,16 @@ def previous_or_current_matching_weekday(
     if days_ahead > 0:  # Target day already happened this week
         days_ahead -= 7
     return d + datetime.timedelta(days_ahead)
+
+
+def localized_short_weekday(weekday: int, lang_code: str) -> str:
```
nicobav, line 34: This could also could make use of django's translations; https://docs.djangoproject.com/en/3.2/topics/i18n/translation/

Then we'd need to accept that the English abbreviation would be Mon, Tue, Wed etc.. (unless we provide custom translations files and maintain those).

So lets say we want "Mon" to be in reservee language => ``translation.activate(lang_code)`` and then ``return _("Mon")`` (that can probably get from the date library) where the ``_`` is imported from ``from django.utils.translation import gettext_lazy as _``

api/graphql/reservations/reservation_serializers.py
saaste: I tried to draft the solution in my head, but it felt like it would just open a whole new can of worms 😅  I feel that the main problem is the fact that we don't want this action to be tied to any save but to be specific to confirm action. I guess it would be possible, but the solution would probably be on the same spot on the hackish-solutions scale 😄 

Perhaps I just leave this as is for now and let's improve if we notice that this does not work.

api/graphql/reservations/helpers.py
```diff
@@ -0,0 +1,130 @@
...
+from tilavarauspalvelu.utils.date_util import localized_short_weekday
+
+from ..application_errors import ValidationErrorCodes, ValidationErrorWithCode
+
+
+def get_reservation_unit_name(
+    reservation: Reservation, reservation_unit: ReservationUnit
+) -> str:
+    preferred_language = reservation.user.preferred_language or "fi"
+    if preferred_language == "fi" and reservation_unit.name_fi:
```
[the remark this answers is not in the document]
saaste: I always forget that! I'll change this, thanks for the reminder! 👍

api/graphql/reservations/reservation_serializers.py
nicobav: Yes, I agree 🤝

api/graphql/reservations/helpers.py
```diff
@@ -0,0 +1,130 @@
+import re
```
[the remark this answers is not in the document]
saaste: I'll move this whole thing under `merchants` since we already have a `helper.py` there 👍

api/graphql/reservations/helpers.py
```diff
@@ -0,0 +1,130 @@
...
+    OrderItemParams,
+)
+from reservation_units.models import ReservationUnit
+from reservations.models import Reservation
+from tilavarauspalvelu.utils.date_util import localized_short_weekday
+
+from ..application_errors import ValidationErrorCodes, ValidationErrorWithCode
+
+
+def get_reservation_unit_name(
```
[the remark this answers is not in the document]
saaste: I'll remove it. There is no need since `getattr` does it with one line 😄

api/graphql/reservations/helpers.py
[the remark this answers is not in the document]
saaste: How to have fun times:
1) Move files around
2) Add fixup-commits that alter commits somewhere in the middle of the commit stack
3) Do git rebase
4) Watch how forces of hell arrive on earth in the form of merge conflict
5) ?????
6) FUN TIMES! 😅

api/graphql/reservations/reservation_serializers.py
```diff
@@ -717,34 +721,142 @@ def check_metadata_fields(self, data, reservation_unit) -> None:
...
+                PaymentOrder.objects.create(
+                    payment_type=payment_type,
+                    status=PaymentStatus.PAID_MANUALLY,
+                    language=self.instance.reservee_language or Language.FI,
+                    price_net=price_net,
+                    price_vat=price_vat,
+                    price_total=price_total,
+                    reservation=self.instance,
+                )
+            else:
```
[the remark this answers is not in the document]
saaste: Payment type is always set when reservation is approved. If approver does not provide the type, there is small helper function that tries to figure out what to use: `reservation_serializers._get_default_payment_type()`. It follows the rules defined in the ticket.

That being said, I just realized that it is possible to have a payment type that is not support by the reservation unit. I need to fix that and add additional check 🛠️

merchants/verkkokauppa/order/requests.py
saaste: Added TODO comment for myself but forgot to implement the DO part 😄 I'll fix this.

utils/decimal_utils.py
```diff
@@ -0,0 +1,9 @@
+from decimal import ROUND_HALF_UP, Decimal
+
+
+def round_decimal(decimal: Decimal, ndigits: int) -> Decimal:
```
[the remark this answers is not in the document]
saaste: Created a ticket for it: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2049

tilavarauspalvelu/utils/date_util.py
saaste: Hmm, `translation.activate` applies the language selection to the current thread. I wonder if this could raise some "surprising features" 🤔 

Do you have any deeper understanding on how Django uses threads? I'm mostly thinking about the risk of setting the language and then serving requests in completely wrong language for other users.

utils/decimal_utils.py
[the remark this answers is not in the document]
nicobav: I think we should change those that already use the ``quantize``  within this introduction. 
I looked and we have it in use in https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/12284e18e594b4e5e67f5ab02b9480af0185ad6e/reservation_units/utils/reservation_unit_pricing_helper.py#L121
and
https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/12284e18e594b4e5e67f5ab02b9480af0185ad6e/reservation_units/utils/reservation_unit_pricing_helper.py#L124

tilavarauspalvelu/utils/date_util.py
nicobav: I think with current setup that we have one thread setupped in uwsgi conf it should be fine.

nicobav: oh, I backup a little => one thread but multiple processes so it might be funky

api/graphql/reservations/helpers.py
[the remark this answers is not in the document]
saaste: Fixed 👍

api/graphql/reservations/helpers.py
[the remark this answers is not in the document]
saaste: Fixed 👍

api/graphql/reservations/reservation_serializers.py
```diff
@@ -717,34 +721,142 @@ def check_metadata_fields(self, data, reservation_unit) -> None:
...
+        data = super().validate(data)
+
+        if self.instance.payment_order.exists():
+            raise ValidationErrorWithCode(
+                "Reservation cannot be changed anymore because it is attached to a payment order",
+                ValidationErrorCodes.CHANGES_NOT_ALLOWED,
+            )
+
+        if self.instance.reservation_unit.count() > 1:
+            raise ValidationErrorWithCode(
```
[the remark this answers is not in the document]
saaste: If I remember correctly, this is an adventure for the future.

api/graphql/reservations/reservation_serializers.py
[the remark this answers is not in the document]
saaste: And that is now fixed as well 👍

api/graphql/reservations/reservation_serializers.py
saaste: I started to move them there but I when I was almost done, I got a feeling that the code felt really out-of-place. 

Creating those orders is kinda specific to the confirm step and having this logic to be in the `merchants` app didn't feel ...right. If someone later opens those helper functions or whatnot, the code looks really confusing because it is out of the context.

So I decided to keep them here, but if you feel that this really isn't the place, I can move them ☺️

merchants/verkkokauppa/order/requests.py
saaste: Fixed 👍

utils/decimal_utils.py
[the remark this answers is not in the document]
saaste: Updated 👍

api/graphql/reservations/reservation_serializers.py
```diff
@@ -740,34 +744,154 @@ def check_metadata_fields(self, data, reservation_unit) -> None:
...
+                "Reservations with multiple reservation units are not supported.",
+                ValidationErrorCodes.MULTIPLE_RESERVATION_UNITS,
+            )
+
+        if not self._requires_handling():
+            payment_type = data.get("payment_type", "").upper()
+            reservation_unit = self.instance.reservation_unit.first()
+
+            if not reservation_unit.payment_product:
```
nicobav, line 822: This got me pondering; do we send every reservation_unit to webshop?

nicobav (approved):
Very nice 🚀

api/graphql/reservations/reservation_serializers.py
saaste: No. Actually, we don't send any runits to webshop.

However, for each runit with **payment merchant** defined, we will create a product mapping - meaning runit has an UUID both we and webshop knows. When order is created, we will send this product UUID and merchant UUID as part of the order. These are the only information webshop knows about the runits.

This is why we check runit's product here - webshop payment is possible only for runits with product mapping defined. Hopefully this clarified things ☺️

nicobav: Yup, that clarified, thanks! Starting to fall off a bit of this webshop wagon

PR #662, opened 2022-11-08
TILA-2036 | Make reservee_id not mandatory or individual reservee

nicobav:
Make reservee_id not mandatory or individual reservee even if defined in metadata to be mandatory.

Refs TILA-2036


saaste (approved):
Noice! 👍

PR #669, opened 2022-11-14
TILA-2057: Add price_net and make handling_details required

saaste:
## Change log
- Added `price_net` field to `ReservationApproveSerializer`
- Changed `handling_details` to be required.

## Other notes
In the code we copy `handling_details` to `working_memo`. If `handling_details` is missing, the code will throw a key error exception. Since client is already always sending the field, I just marked it as required.

## Deployment reminder
- No changes needed


nicobav (approved):
🚀

PR #671, opened 2022-11-17
TILA-2071: Return order data directly from confirm mutation

saaste:
## Change log
- Rename `order_id` to `remote_id` in the model
- Rename `order_id` to `order_uuid` in the API
- Return order data directly from the confirm mutation to avoid extra requests from the frontend

## Other notes
~~That `id <-> uuid` mess really annoys me. I don't think client needs the field but I double-checked it from Juhis. If field is not needed by the frontend, I'm considering dropping it from the API. My understanding is that if client needs order data, it can use `pk` field and that verkkokauppa-generated ID is only used between backend and verkkokauppa (which is fine)~~

Turns out, we do need the generated ID but we don't need our own primary key. So I changed the code so that the only ID we return is the generated one and it is called `order_uuid`. That way we have just one ID that frontend needs to use.

## Deployment reminder
- No changes needed


nicobav (approved):

nicobav (approved):
Let's :rocket:

api/graphql/reservations/reservation_types.py
```diff
@@ -327,7 +327,9 @@ def resolve_cancel_details(self, info: ResolveInfo) -> Optional[str]:
     def resolve_order_uuid(self, info: ResolveInfo) -> Optional[str]:
         payment_order = self.payment_order.first()
         return (
-            payment_order.order_id if payment_order and payment_order.order_id else None
+            payment_order.remote_id
```
nicobav, line 330: I'd suggest (not strongly) to refactor this to;
```
uuid = getattr(payment_order, "remote_id", None)
return uuid
```
For readability

nicobav (approved):

api/graphql/reservations/reservation_types.py
saaste: Fixed! 👍

PR #672, opened 2022-11-17
TILA-2003 | Reservation adjust time mutation

nicobav:
## Change log
- Add new mutation reservationAdjustTime
  - Enables reservation time change for reservee or staff user.

## Other notes
Done a bit of a refactor within the changes so commit kinda yuge 🙈


api/graphql/reservations/reservation_serializers.py
```diff
@@ -1167,3 +1167,137 @@ class Meta:
...
+            )
+
+    def check_cancellation_rules(self, reservation_unit):
+        now = datetime.datetime.now(tz=DEFAULT_TIMEZONE)
+
+        cancel_rule = reservation_unit.cancellation_rule
+        if not cancel_rule:
+            raise ValidationErrorWithCode(
+                "Reservation time cannot be changed thus no cancellation rule.",
+                ValidationErrorCodes.CANCELLATION_NOT_ALLOWED,
```
saaste, line 1272: Here, too, we have multiple errors with `CANCELLATION_NOT_ALLOWED` error codes. Can reservation be cancelled via time change mutation? 🤔

api/graphql/reservations/reservation_serializers.py
```diff
@@ -1167,3 +1167,137 @@ class Meta:
...
+            self.instance.begin - cancel_rule.can_be_cancelled_time_before
+        )
+        if must_be_cancelled_before < now:
+            raise ValidationErrorWithCode(
+                "Reservation time cannot be changed because the cancellation period has expired.",
+                ValidationErrorCodes.CANCELLATION_NOT_ALLOWED,
+            )
+        if cancel_rule.needs_handling:
+            raise ValidationErrorWithCode(
+                "Reservation time change needs manual handling.",
```
saaste, line 1284: I'd change the message to be more explaining. Something like:
```suggestion
                "Reservation time cannot be changed because cancel rule needs manual handling.",
```

saaste (changes requested):
Those mixins! 🧑‍🍳🤌 

I had just few comments regarding error messages, but otherwise this looks really good to me. I wonder if it would make sense to extract those mixing classes into a separate file just to cut down the number of lines in one file. Might be good, might be bad. Don't know, really 🤔

api/graphql/reservations/reservation_serializers.py
```diff
@@ -1167,3 +1167,137 @@ class Meta:
...
+            )
+
+        return instance
+
+    def validate(self, data):
+        data = super().validate(data)
+
+        if self.instance.state != STATE_CHOICES.CONFIRMED:
+            raise ValidationErrorWithCode(
+                "Only reservations in confirmed state can be cancelled through this.",
```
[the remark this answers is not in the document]
nicobav: yup, you're right => I'll change the text

api/graphql/reservations/reservation_serializers.py
nicobav: The error code itself is specsed in ticket to be `CANCELLATION_NOT_ALLOWED` since here is the cancellation rule that is causing the fail. We first pondered that should we return different codes for every situation but nah, the UI knows that it is cancellation rule related error when changing time when return `CANCELLATION_NOT_ALLOWED`

So shortly; the time change itself is kind of a cancellation for the old time.

nicobav:
>I wonder if it would make sense to extract those mixing classes into a separate file just to cut down the number of lines in one file. Might be good, might be bad. Don't know, really 🤔

Yes that would be nice and then I would actually refactor every serializer in its own file under `resrevation_serializers/` But chose not to do that within this. But something to perhaps do in the near future 👍

saaste (approved):
Yabba Dabba Doo! 🚀

PR #673, opened 2022-11-18
TILA-2023 | Adds mutation for deleting the reservation

nicobav:
Reservation can be deleted when it is in CREATED state or WAITING_FOR_PAYMENT

Refs TILA-2023

### DIFF
--- a/api/graphql/reservations/reservation_mutations.py
+++ b/api/graphql/reservations/reservation_mutations.py
@@ -1,8 +1,10 @@
 import graphene
-from graphene import ResolveInfo
+from django.core.exceptions import ValidationError
+from graphene import ClientIDMutation, ResolveInfo
 from graphene_django.rest_framework.mutation import SerializerMutation
+from rest_framework.generics import get_object_or_404
 
-from api.graphql.base_mutations import AuthSerializerMutation
+from api.graphql.base_mutations import AuthDeleteMutation, AuthSerializerMutation
 from api.graphql.merchants.merchant_types import PaymentOrderType
 from api.graphql.reservations.reservation_serializers import (
     ReservationApproveSerializer,
@@ -22,6 +24,7 @@ from permissions.api_permissions.graphene_permissions import (
     ReservationHandlingPermission,
     ReservationPermission,
 )
+from reservations.models import STATE_CHOICES as ReservationState
 from reservations.models import Reservation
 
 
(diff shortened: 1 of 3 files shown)


api/graphql/reservations/reservation_mutations.py
```diff
@@ -114,3 +117,21 @@ class Meta:
...
+    model = Reservation
+
+    @classmethod
+    def validate(self, root, info, **input):
+        reservation = get_object_or_404(Reservation, pk=input.get("pk", None))
+        if reservation.state not in (
+            ReservationState.CREATED,
+            ReservationState.WAITING_FOR_PAYMENT,
+        ):
+            raise ValidationError(
```
saaste, line 133: We could use `ValidationErrorWithCode` here

api/graphql/validation_errors.py
```diff
@@ -43,6 +43,7 @@ class ValidationErrorCodes(Enum):
     INVALID_PAYMENT_TYPE = "INVALID_PAYMENT_TYPE"
     MISSING_PAYMENT_PRODUCT = "MISSING_PAYMENT_PRODUCT"
     UPSTREAM_CALL_FAILED = "UPSTREAM_CALL_FAILED"
+    DELETION_NOT_ALLOWED = "DELETION_NOT_ALLOWED"
```
saaste, line 46: This is not used anywhere, probably because `ValidationErrorWithCode` is not used ☺️

saaste (changes requested):

api/graphql/reservations/reservation_mutations.py
nicobav: Its kinda decision to not use it here since it would be quite a change and would need a refactor and change other deletions to use it as well. Which sounds awesome, but would take a bit of more time so decided not to since we have the other ticket relating to error codes to root.

api/graphql/validation_errors.py
nicobav: Yup, I was implementing this with ``ValidationErrorWithCode`` but as mentioned above decided not to.

saaste: I guess this line could be removed then to get rid of the unused code.

Side note: searching and removing unused code could be fun Friday activity in general. I've found few of them just by luck so I think our static code analyzers don't inspect that 🤔 

Perhaps something for Friyays 😄 

Other than that, the PR looks great!

nicobav: its now removed.

saaste (approved):
Looks gggrrrreeat! 🚢  it!

PR #674, opened 2022-11-21
TILA-1971: Payment webhook endpoint to REST API

saaste:
## Change log
- Updated `drf-spectacular` package
- Updated `Payment API` client to be up-to-date with the latest implementation
- Extracted reservation email sending to own file
- Added a new payment webhook endpoint to REST API
- Moar tests

## Other notes
There are some known changes coming to Payment Experience API but since those are not yet deployed, I added a TODO comment as a reminder. I'll update the code as soon as I get an update from the Verkkokauppa team. 

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_serializers.py
+++ b/api/graphql/reservations/reservation_serializers.py
@@ -15,11 +15,6 @@ from api.graphql.choice_char_field import ChoiceCharField
 from api.graphql.duration_field import DurationField
 from api.graphql.primary_key_fields import IntegerPrimaryKeyField
 from applications.models import CUSTOMER_TYPES, City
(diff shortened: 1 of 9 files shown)


nicobav (approved):
awwwwsome 👍

saaste:
>Just noticed that there aren't any refs in commits. They're I find them useful when debugging and pondering the logic etc. since you could go and see the ticket based on the ref

I usually include the ticket ID in the title of the PR. Most of the time I review the code via GitHub UI so I rarely need the ticket id to be in the commits. But I'll try to add them in the future if it helps you ☺️ 

Keep reminding me because there is a high chance that I forget to do it 🧠 🚫

PR #677, opened 2022-11-22
TILA-2068 | Include reservations with same components to res unit reservation query

nicobav:
## Change log
- Refactor resunit tests to own files
- Add option to include clashing reservations with same components in reservationUnit query selecting the reservations.


saaste (approved):
Holy mackerel, Batman! What a nice refactoring! I definitely read every single line while doing this review 🤞 

Looks great! :shipit:

PR #679, opened 2022-11-22
TILA-2073 | Change staff event to reservation type

nicobav:
## Change log
- Removes the staff_event field from `Reservation` model 
- Obtain the `staffEvent` field in gql api but make it depricated
- Remove the staff_event from ``ReservationStatistics``

## Other notes
There is some mutations coming in other ticket later which make use of the new type and the permission in the other PR.


saaste (approved):
It feels like I added the field just few weeks ago 😄 
Good stuff! 🚢

nicobav:
> It feels like I added the field just few weeks ago 😄 Good stuff! 🚢

yup, fields grow so fast

PR #682, opened 2022-11-23
TILA-2014: Add refresh_order mutation

saaste:
## Change log
- Add `refresh_order` mutation
- Add `can_refresh_order` permission check helper
- Fix error message

## Other notes
The mutation is used in cases where client returns from the webshop after payment, but backend did not yet receive or process the webhook call that updates the order state.

If backend returns an order in `DRAFT` state, client can call this mutation and force the refresh. Backend will then checks and updates the order state from the webshop API.

Mutation can be called by users with `can_manage_reservations` role or by user who created the original reservation.

## Deployment reminder
- No changes required


api/graphql/merchants/merchant_mutations.py
```diff
@@ -0,0 +1,73 @@
...
+from merchants.models import PaymentOrder, PaymentStatus
+from merchants.verkkokauppa.payment.exceptions import GetPaymentError
+from merchants.verkkokauppa.payment.requests import get_payment
+from reservations.email_utils import send_confirmation_email
+from reservations.models import STATE_CHOICES
+
+from ..validation_errors import ValidationErrorCodes, ValidationErrorWithCode
+
+
+class RefreshOrderMutation(relay.ClientIDMutation, AuthMutation):
```
nicobav, line 16: Do we need some permissions to perform this operations? This is ``AuthMutation`` but no ``permission_classes`` are defined 🕵️

saaste: The caller cannot really "do" anything with this since this is just for triggering an action that should happen behind the scenes automatically. So there isn't big harm that can be done by doing this. I'll double check it.

saaste: Oh and if I'm not mistaken, this is not called via admin UI but via public UI after customer paid the order. I have a feeling we can't have any roles, but perhaps we could check that the caller is the same as the owner of the related reservation 🤔

saaste: There is now a permission checks in place. I couldn't figure out an easy way to do this with existing permission checks, so I created a new `can_refresh_order` helper function.

nicobav: We could add Permission class for this which uses the `can_refresh_order`

nicobav (approved):
Looks good 👍 

Added optional commenting about changing the mutation to be using permission classes which use `can_refresh_order` permission helper.

api/graphql/merchants/merchant_mutations.py
saaste: Done! 👍

PR #683, opened 2022-11-24
TILA-2074 | Reservation staff create mutation

nicobav:
## Change log
- Refactors reservation serializers to own sub package and files.
- Introduce new mutation createStaffReservation which is intended for staff users to create reservations

## Other notes
Happy reviewing 🥂


api/graphql/reservations/reservation_serializers/staff_create_serializers.py
```diff
@@ -0,0 +1,228 @@
...
+        self.set_buffer_times(data)
+
+        user = self.context.get("request").user
+        data["user"] = user
+
+        return data
+
+    def set_buffer_times(self, data):
+        if not data.get("buffer_time_after"):
+            data["buffer_time_after"] = None
```
saaste, line 179: Could you clarify what is the purpose of this? Initially it looked like you are trying to set `buffer_time_after` to `None` if it is `None`, but perhaps the point here is to replace other falsy values like empty strings with `None`?

api/graphql/reservations/reservation_serializers/staff_create_serializers.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if not data.get("buffer_time_before"):
+            data["buffer_time_after"] = None
```
saaste, line 182: Is this correct? Should it be `data["buffer_time_before"]`?

saaste (changes requested):
Once again, amazing work with the refactoring! 👏 

Just two small questions but otherwise looks great! 👍

api/graphql/reservations/reservation_serializers/staff_create_serializers.py
nicobav: should but as you pointed out should not be necessary to touch it and that's why tests passes 😂

api/graphql/reservations/reservation_serializers/staff_create_serializers.py
nicobav: Yup, no point, I was so heavily comparing the code to the "regular" create where the buffer time gets copied from reservation_unit. I'll change this. Good catch 👍

nicobav: its now removed 🧘‍♂️

saaste (approved):
Jokeri pokeri box! :shipit:

PR #684, opened 2022-11-24
TILA-2077: Send order payment expiration to order API

saaste:
## Change log
- Refactoring: use factory helper instead of building the object in the test
- Add order payment expiration to OrderCreate parameters

## Other notes
This is a new field Verkkokauppa guys added. We agreed that the default value is 5 minutes but it is configurable.

## Deployment reminder
- We could add this but we can do it when we need something else than the default value


nicobav (approved):
:shipit:

PR #685, opened 2022-11-24
TILA-2038 | Include REQUIRES_HANDLING to going_to_occur qs method

nicobav:
## Change log
- REQUIRES_HANDLING to `ReservationQueryset.going_to_occur` 

## Long notes
In reservations queryset we have util queryset function to get reservations that are going to occur e.g not cancelled or denied. The method is used in various places and it was lacking the REQUIRES_HANDLING state which also means that the reservation is going to occur.

This adds REQUIRES_HANDLING state to going_to_occur method.

Refs TILA-2038


saaste (approved):
One-liners ❤️

nicobav:
yes, oneliners, they often need biggest explanations 😅

PR #686, opened 2022-11-25
Add timezone to ones missing

nicobav:
There was missing timezone declaration in various places using datetimes. This adds the tzinfo to those.

Some is still missing since getting warnings but cannot found the source yet so this is the changeset for now.

## Change log
- Timzones to various places

### DIFF
--- a/allocation/allocation_runner.py
+++ b/allocation/allocation_runner.py
@@ -1,4 +1,5 @@
 from django.utils.datetime_safe import datetime
+from django.utils.timezone import get_default_timezone
 
 from allocation.allocation_data_builder import AllocationDataBuilder
 from allocation.allocation_solver import AllocationSolver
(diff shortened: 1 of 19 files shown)


merchants/verkkokauppa/helpers.py
```diff
@@ -23,7 +23,7 @@
 def parse_datetime(string: Optional[str]) -> Optional[datetime]:
     if string is None:
         return None
-    return datetime.strptime(string, "%Y-%m-%dT%H:%M:%S.%f")
+    return datetime.strptime(string, "%Y-%m-%dT%H:%M:%S.%f")  # what is the timezone?
```
nicobav, line 26: I pondered how does the verkkokauppa work with timezones. Does it send the times without zone info and we expect them to be UTC or UTC+3 🕵️

saaste: I asked this from Severi and he told me all times are UTC

saaste (approved):
Cooool beans!

merchants/verkkokauppa/helpers.py
saaste: Oh and feel free to change to to explicitly read the datetime as UTC. I already approved the rest of the PR 😄

PR #688, opened 2022-11-28
TILA-1518: Remove ununsed code

saaste:
## Change log
- Removed unused code

## Other notes
Cleaning day! 🧹 

Finished what I started on Friyay. I used [Vulture](https://github.com/jendrikseipp/vulture) to find dead code and removed the ones that felt obvious. Naturally, Vulture found a lot more but because we have some "magic methods" that are following certain naming conventions (like resolvers). Also, it does not understand classes/methods being used via decorators so naturally I didn't delete those.

## Deployment reminder
- No changes required


nicobav (approved):
Looks clean to me!

PR #689, opened 2022-11-28
TILA-2033: Add webshop order cancellation support

saaste:
## Change log
- Added support for webshop order cancellation
- Cancel webshop order when reservation is deleted
  - If cancellation is successful, also cancel our PaymentOrder

## Other notes
I just realized order status are still plain strings. I'll do a separate refactoring PR when I will change them to enums 🛠️ 

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_mutations.py
+++ b/api/graphql/reservations/reservation_mutations.py
@@ -3,6 +3,7 @@ from django.core.exceptions import ValidationError
 from graphene import ClientIDMutation, ResolveInfo
 from graphene_django.rest_framework.mutation import SerializerMutation
 from rest_framework.generics import get_object_or_404
+from sentry_sdk import capture_message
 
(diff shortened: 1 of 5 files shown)


api/graphql/reservations/reservation_mutations.py
```diff
@@ -136,6 +139,21 @@ def validate(self, root, info, **input):
...
+            try:
+                webshop_order = cancel_order(
+                    payment_order.remote_id, payment_order.reservation_user_uuid
+                )
+
+                if webshop_order and webshop_order.status == "cancelled":
+                    payment_order.status = PaymentStatus.CANCELLED
+                    payment_order.save()
+            except CancelOrderError as err:
+                capture_message(
```
nicobav, line 153: Should we also inform UI for the errors?

nicobav (approved):
Looks good! 👍 

One thing to ask what does not affect approval ✅

PR #690, opened 2022-11-29
TILA-1973: Scheduled payment order update

saaste:
## Change log
- Add changes
- Here

## Other notes
Additional notes if needed

## Deployment reminder


reservations/tasks.py
```diff
@@ -21,6 +22,7 @@
 def _prune_reservations() -> None:
     prune_inactive_reservations(PRUNE_OLDER_THAN_MINUTES)
     prune_reservation_with_inactive_payments(PRUNE_WITH_ORDERS_OLDER_THAN_MINUTES)
+    update_expired_orders(PRUNE_WITH_ORDERS_OLDER_THAN_MINUTES)
```
nicobav, line 25: I think this could be in its own task. What do you think? 🤔

nicobav (commented):
Overall looks good 👌 

Some ponders before partying the merge

reservations/tasks.py
saaste: Moved! 👍

merchants/tests/test_pruning.py
```diff
@@ -0,0 +1,151 @@
...
+from merchants.verkkokauppa.payment.types import PaymentStatus as WebShopPaymentStatus
+from reservations.models import STATE_CHOICES
+from reservations.tests.factories import ReservationFactory
+
+from ..pruning import update_expired_orders
+
+TIMEZONE = get_default_timezone()
+
+
+class UpdateExpiredOrderTestCase(TestCase):
```
nicobav, line 24: What I see here is using the freeze_time first as "base time" and then doing some action after some time has passed. I think we could add the freeze time as a decorator like;
```
@freeze_time(datetime(2022, 11, 28, 10, 10, 0, tzinfo=TIMEZONE))
class UpdateExpiredOrderTestCase(TestCase):
```
and then could remove the first `with freeze_time...` from the tests.


Note that this is fine as is now and no action needed. Surely I'm not even sure how the freeze time acts when you've used the decorator and then use it with the `with` statement

nicobav (approved):
Awesomeness 😎

PR #692, opened 2022-12-01
TILA-2029: Reservation - add sorting by order status and rename order status filter

saaste:
## Change log
- Add `order_status` to Reservation `order_by` fields
- Refactoring: renamed `payment_status` filter to `order_status` (see notes)

## Other notes
`ReservationType` has a field called `order_status` but the related filter was called `payment_status` which led a bit confusing queries. I renamed the filter so that now everything - field, filter and order_by - is using the same naming convention.

Filter is already used on the admin client, so this requires some syncing with the frontend code. @wuahi will work on this on the frontend side.

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/tests/test_reservations/test_reservation_queries.py
+++ b/api/graphql/tests/test_reservations/test_reservation_queries.py
@@ -296,7 +296,7 @@ class ReservationQueryTestCase(ReservationTestCaseBase):
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
 
-    def test_filter_payment_status_single_value(self):
+    def test_filter_order_status_single_value(self):
         self.maxDiff = None
         self.client.force_login(self.general_admin)
         metadata = ReservationMetadataSetFactory()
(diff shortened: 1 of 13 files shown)


api/graphql/tests/test_reservations/test_reservation_queries.py
```diff
@@ -1252,6 +1252,45 @@ def test_order_by_created_at(self):
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
 
+    def test_order_by_order_status(self):
+        res1 = ReservationFactory(name="this should be 1st")
+        PaymentOrderFactory(status=PaymentStatus.CANCELLED, reservation=res1)
```
wuahi, line 1257: Should these enums be also OrderStatus? For concistency? or does that mean a larger refactor?

nicobav: I second the motion 👍

nicobav (changes requested):

api/graphql/tests/test_reservations/test_reservation_queries.py
saaste: I very much approve the suggestion! Added a new commit with fancy find-and-replace changes 👍

nicobav (approved):
Awesomeness cannot be avoided 🚀

PR #696, opened 2022-12-05
TILA-1518: Configurable log levels

saaste:
## Change log
- Added a helper function that adds `tilavaraus` namespace to all loggers
- Made log level of `tilavaraus` namespace configurable via envs
- Changed all loggers to use the new function

## Deployment reminder
- No changes required. Log level can be added when needed.


nicobav (approved):
Very noice 🍻

PR #698, opened 2022-12-07
TILA-2074 | Check interval for staff reservations from times only

nicobav:
## Change log
- Uses time, not datetime, when checking start intervals for staff reservation

In staff reservation the opening hours does not matter so we can just check the times instead of datetimes.
There was also a bug that the used only current day to check the intervals and thus would fail when trying to reserve to future dates. This also is fixed when used only times.

Refs TILA-2074


saaste (approved):
Superb! :shipit:

PR #699, opened 2022-12-09
TILA-2098: Task for updating accounting info in web shop

saaste:
## Change log
- Some automatic formatting by `format.sh` 🤷‍♂️ 
- Fixed a wrong return type
- Added task for updating accounting data in web shop

## Other notes
This task is triggered by another task: `refresh_product_mapping`. Basically it makes sure that whenever we update
merchant information on our side, we also refresh the product mapping and as a result also send the new accounting data to web shop.

## Deployment reminder
- No changes required


reservation_units/tasks.py
```diff
@@ -58,8 +65,51 @@ def refresh_reservation_unit_product_mapping(reservation_unit_pk) -> None:
...
+            vat_code=accounting.vat_code,
+            internal_order=accounting.internal_order,
+            profit_center=accounting.profit_center,
+            project=accounting.project,
+            operation_area=accounting.operation_area,
+            company_code=accounting.company_code,
+            main_ledger_account=accounting.main_ledger_account,
+        )
+        try:
+            accounting = create_or_update_accounting(
```
saaste, line 107: I'm really tempted to rename this method to `upsert_accounting` just to save some characters. Hash tag code churn :rage4:

nicobav:
> Some automatic formatting by format.sh 🤷‍♂️

In my dev the format.sh does not re-format those 😕

nicobav (changes requested):

nicobav:
> In my dev the format.sh does not re-format those 😕

might be black version related👀

reservation_units/utils/reservation_unit_payment_helper.py
```diff
@@ -30,3 +30,15 @@ def requires_product_mapping_update(cls, reservation_unit) -> bool:
...
 
         return False
+
+    @classmethod
+    def get_accounting(cls, reservation_unit) -> Optional[PaymentAccounting]:
+        if reservation_unit.payment_accounting is not None:
+            return reservation_unit.payment_accounting
+        if (
+            reservation_unit.unit
+            and reservation_unit.unit.payment_accounting is not None
```
nicobav, line 40: Minor tweaky; I think we can check just `if reservation_unit.unit` and then return the `payment_accounting` even if `None` since we're returning `None` in the below

nicobav (approved):
Looks good! :shipit:

reservation_units/utils/reservation_unit_payment_helper.py
saaste: True! I simplified the code 👍

PR #701, opened 2022-12-12
TILA-2100 | New fields to RecurringReservation model

nicobav:
## Change log
- bunch of new fields to `RecurringReservation` model; created, reservation_unit, weekdays, interval, name
- Add WEEDAYS class/choices enum as to be common util 
   - Refactors places to be using it where weekdays were used.

## Other notes
Migration removes `RecurringReservations` which does not have `Reservations`. 
Also needed to introduce reservation_unit field as nullable first and then set up those and then finally altering it to be non nullable.


saaste (approved):
Awesome stuff! 👍

PR #702, opened 2022-12-12
TILA-2163: Fix payment product requirement

saaste:
## Change log
- Fix payment product requirement
- Removed payment_type check when price is zero
- Added tests

## Other notes
Payment product and payment_type check are not required when price of the reservation is zero.

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_serializers/confirm_serializers.py
+++ b/api/graphql/reservations/reservation_serializers/confirm_serializers.py
@@ -79,7 +79,7 @@ class ReservationConfirmSerializer(ReservationUpdateSerializer):
             payment_type = data.get("payment_type", "").upper()
             reservation_unit = self.instance.reservation_unit.first()
 
-            if not reservation_unit.payment_product:
+            if self.instance.price_net > 0 and not reservation_unit.payment_product:
                 raise ValidationErrorWithCode(
                     "Reservation unit is missing payment product",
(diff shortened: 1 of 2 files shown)


vergama:
I guess the payment_type validation is also only relevant for priced reservations.

saaste:
> I guess the payment_type validation is also only relevant for priced reservations.

@vergama I added the price check to there as well 👍

api/graphql/reservations/reservation_serializers/confirm_serializers.py
```diff
@@ -79,15 +79,20 @@ def validate(self, data):
             payment_type = data.get("payment_type", "").upper()
             reservation_unit = self.instance.reservation_unit.first()
 
-            if not reservation_unit.payment_product:
+            if self.instance.price_net > 0 and not reservation_unit.payment_product:
```
nicobav, line 82: what if reservation_unit `pricing_type` is `FREE`? In that case I think the price validations should be ignored.

saaste: Krista confirmed this and said that it is not possible to have `price > 0` if `pricing_type` is `FREE`. We already have validation in place in the UI and also price calculations are skipped for `FREE` reservation units. So I'd say we don't need extra validation for this because it basically means data has been tweaked directly via database or via Django Admin.

nicobav (approved):
Let's 🪨 !

PR #703, opened 2022-12-13
TILA-2119 | Remove leading zero from notification email begin and end dates

nicobav:
## Change log
- Removes leading zero from notification email begin and end dates
- Moves email_notification.helpers to email_notification.sender.helpers (it contained just sending helpers)

Refs TILA-2119


saaste (approved):
So good that we don't have to worry about localization and different date and time formats! 😅

PR #705, opened 2022-12-14
TILA-1926: Remove deprecated pricing fields from ReservationUnit

saaste:
## Change log
- Removed deprecated pricing fields from ReservationUnit
- Updated quite many tests to reflect the changes

## Other notes
Based on UI devs, the UI has been updated to reflect these changes. It should be safe to remove these fields.

## Deployment reminder
- No changes required

PR #706, opened 2022-12-14
TILA-2177: Fix payment_type validation in reservation confirm

saaste:
## Change log
- Fixed invalid check
- Updated tests

## Other notes
I changed the code so that validation logic kicks in only if price of the reservation > 0 OR reservation unit's pricing type is `PAID`. Both should be true at the same time if data has been created via UI, but just to be on the safe side I used `or` here.

This is based on #705 because I don't want to make test changes again 😅 
I'll rebase when #705 is merged.

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_serializers/confirm_serializers.py
+++ b/api/graphql/reservations/reservation_serializers/confirm_serializers.py
@@ -5,7 +5,10 @@ from api.graphql.reservations.reservation_serializers.update_serializers import 
 from api.graphql.validation_errors import ValidationErrorCodes, ValidationErrorWithCode
 from merchants.models import Language, OrderStatus, PaymentOrder
 from merchants.verkkokauppa.helpers import create_verkkokauppa_order
-from reservation_units.models import PaymentType
+from reservation_units.models import PaymentType, PricingType
+from reservation_units.utils.reservation_unit_pricing_helper import (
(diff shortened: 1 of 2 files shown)


api/graphql/reservations/reservation_serializers/confirm_serializers.py
```diff
@@ -79,28 +82,32 @@ def validate(self, data):
...
-                )
-
-            if not payment_type:
-                data["payment_type"] = self._get_default_payment_type()
-            elif (
-                self.instance.price_net > 0
-                and not reservation_unit.payment_types.filter(
-                    code=payment_type
-                ).exists()
+            active_price = ReservationUnitPricingHelper.get_active_price(
```
nicobav, line 85: Oh, just realized this helper class to fetch active pricing. I also added custom QuerySet manager to fetch active pricings in another PR.

nicobav (approved):
Looks good 🚀

api/graphql/reservations/reservation_serializers/confirm_serializers.py
saaste: It probably would be nice to have this inside `ReservationUnit` so that you could just call `reservation_unit.get_active_price()` and there is no need to remember any helper classes  🤔

PR #707, opened 2022-12-14
TILA-2100 | Fix recurring reservation changes migration

nicobav:
It was saving twice which caused the postgres to throw an error about pending trigger events

Refs TILA-2100


saaste (commented):
Looks good to me! 👍 
Then again, so did the previous one and it had some issues 😄

PR #708, opened 2022-12-14
TILA-2100 | Separate recurring reservation data mod to own migration file

nicobav:
Refs TILA-2100


saaste (commented):
I tested this on my machine and it seems to work 👍

PR #709, opened 2022-12-15
TILA-1978 | Add new fields to reservation unit exporter

nicobav:
## Change log
- Adds bunch of new fields to `ReservationUnit` exporter

## Other notes
These fields were already on previous tag (0.17.1) but not in main since that version of the tag there weren't yet pricing fields.


saaste (approved):
Go go! :shipit:

PR #711, opened 2022-12-16
TILA-2187: Fix app and SQL logging

saaste:
## Change log
- Add a new handler
- Use added handler for app and sql loggers

## Other notes
Logs didn't appear in the console but were written to a file.

## Deployment reminder
- No changes required


nicobav (approved):
Excellence!

PR #714, opened 2022-12-22
TILA-2131 | Add support for if else tags in email templates

nicobav:
## Change log
- Adds Jinja2 to requirements for Template validation.
- Adds new setting for adding supported template expressions (like if for)
   - Currently supported are if else elif


saaste (approved):
Looks great! Tests are failing but it feels like one more flaky test that is failing because ordering changes 🤔 

Feel free to merge this first. I'll update my PR after that since there will be small merge conflicts.

nicobav:
> Looks great! Tests are failing but it feels like one more flaky test that is failing because ordering changes 🤔

Yup, there's something with resunit ordering 🤔

PR #716, opened 2022-12-23
TILA-2219: Improve exception logging

saaste:
## Change log
- Use `capture_exception` instead of `capture_message`
- Use exception inheritance where it makes sense
- Add additional information to `capture_exception` where it makes sense

## Deployment reminder
- No changes required


nicobav (approved):
Awesome! 🚀

PR #718, opened 2022-12-23
TILA-2133 | Add subsidised price to email template variables

nicobav:
## Change log
- Adds subsidised price calc to reservation price calculation method
- Use the calculated price in email templates.


saaste (approved):
Awesome! 🤘

PR #728, opened 2023-01-10
Upgrade graphene and graphene-django packages

nicobav:
## Other notes
Bunch of stuff happened and decimal support exists now so kinda breaking mutation changes since the mutations need to have Decimals in strings. This is a breaking feature for UI for creating and updating reservation units at least.

**EDIT**
Now contains also rendering all the errors to root of the response doc.

Refs TILA-1982 (partially relates to it)


permissions/api_permissions/graphene_permissions.py
```diff
@@ -421,6 +422,10 @@ def has_mutation_permission(cls, root: Any, info: ResolveInfo, input: dict) -> b
 
 
 class AgeGroupPermission(BasePermission):
+    @classmethod
+    def has_permission(cls, info: ResolveInfo) -> bool:
+        return True
```
nicobav, line 427: some permissions checks improvements since this needed to be added

api/graphql/tests/test_reservations/test_reservation_approve.py
```diff
@@ -70,8 +70,8 @@ def get_valid_approve_data(self):
         return {
             "pk": self.reservation.pk,
             "handlingDetails": "You're welcome.",
-            "price": 10.59,  # This floating point number will float somewhere 10.58999...
-            "priceNet": 8.61,
+            "price": "10.59",  # This floating point number will float somewhere 10.58999...
```
saaste, line 73: I'm assuming this comment is not valid anymore ☺️

permissions/api_permissions/graphene_permissions.py
```diff
@@ -55,10 +55,11 @@ def has_permission(cls, info: ResolveInfo) -> bool:
     @classmethod
     def has_node_permission(cls, info: ResolveInfo, id: str) -> bool:
         user = info.context.user
-        application = Application.objects.filter(id=id)
+        application = Application.objects.filter(id=id).first()
 
         if application:
-            return user.is_authenticated and can_read_application(user, application)
+            access = user.is_authenticated and can_read_application(user, application)
+            return access
```
saaste, line 62: Nitpick: variable assignment feels a bit unnecessary here since we just return the value

saaste (commented):
Getting rid of floats, noice! 👨‍🍳 🤌

Just some nitpicks comments, but other than that looks great! This is good to go from my side, but I let frontend guys to do the approvement this so that we don't accidentally merge this before the UI is ready ☺️

permissions/api_permissions/graphene_permissions.py
nicobav: oh, this change shouldn't be here. I'll clean it.

nicobav: cleaned 🧹

api/graphql/tests/test_reservations/test_reservation_approve.py
nicobav: cleaned 🧹

api/graphql/schema.py
```diff
@@ -197,7 +197,7 @@ def resolve_queryset(
                 application_events__event_reservation_units__reservation_unit__unit__in=units
             )
             | Q(user=user)
-        ).distinct()
+        )
```
saaste, line 200: Could you clarify why `distinct` is not needed here anymore? Just wanting to make sure it isn't an accidental removal 😄

saaste (commented):
Just one verification question but otherwise this looks superb!

api/graphql/schema.py
nicobav: I certainly can't! :D So very nice catch 🎣

nicobav: fix'd

saaste (approved):
Guuuuuud staff! :shipit:

nicobav:
@kieferbonk approve when UIs are ready for these, then we can merge

PR #729, opened 2023-01-12
TILA-1495: Add custom metrics for verkkokauppa

saaste:
## Change log
- Add helper classes for custom metrics
- Use new helpers with verkkokauppa requests

## Other notes
The new file wraps `prometheus_client` details so that we don't have to worry about them when we want to collect metrics. The initial setup is for Verkkokauppa and it collects number of calls, latencies and responses. These can be filtered by service, method, path and status code.

This PR might be a bit annoying to review without knowledge of how Prometheus metrics work, but feel free to ask me in Slack if things are unclear ☺️ 

## Deployment reminder
- No changes required


nicobav (approved):
LGTM! 🚀

PR #732, opened 2023-01-17
TILA-2283 | Add hauki_department_id property to unit

nicobav:
Refs TILA-2283

## Change log
- When generating hauki link within gql response use prefix "tprek:" in organisation id.

## Other notes
Hauki uses the department id with the prefix of the origin_id. In this department id case it needs to in form "tprek:<id>" to make the link work properly.
This adds computed property called "hauki_department_id" to Unit model which returns the tprek_department id with prefix "tprek:".


saaste (approved):
Great! 👍

PR #733, opened 2023-01-18
TILA-2185: Count reservations only from same resunit

saaste:
## Change log
- Fix `check_max_reservations_per_user`

## Other notes
When max_reservations_per_user is checked, we should consider only reservations from the same reservation units. Currently, people are unable to make new reservations because globally they have multiple reservations.

## Deployment reminder
- No changes required


nicobav (approved):
🚀

PR #734, opened 2023-01-19
TILA-2101 | RecurringReservation GQL api - mutations and queries

nicobav:
## Change log
- Adds mutations for creating and updating RecurringReservations
- Adds queries for querying RecurringReservations

## Other notes
RecurringReservations are only visible for staff with correct permissions. Own RecurringReservations can be seen.


api/graphql/reservations/recurring_reservation_serializers/create_serializers.py
```diff
@@ -0,0 +1,124 @@
...
+            if weekday not in WEEKDAYS.VALUES:
+                raise ValidationErrorWithCode(
+                    f"Invalid weekday {weekday}.", ValidationErrorCodes.INVALID_WEEKDAY
+                )
+            days.append(str(weekday))
+
+        return ",".join(days)
+
+    def validate_interval(self, interval):
+        allowed_values = [i for i in range(170) if i % 7 == 0]
```
saaste, line 116: `170` feels like a magic number. Where does it come from?

api/graphql/reservations/recurring_reservation_serializers/create_serializers.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["user"].readonly = True
+
+    def validate(self, data):
+        begin = data.get("begin_time", getattr(self.instance, "begin_time", None))
+        end = data.get("end_time", getattr(self.instance, "end_time", None))
+
+        if end < begin:
```
saaste, line 80: This isn't probably a real world situation, but is it possible that we have a valid time that is invalid because of timezones?

Let's say we have a reservation from 02:00 - 04:00 @ UTC+3. If time is stored in UTC, this becomes 23:00 - 01:00. I didn't really get a grasp on how timezones are handled with `TimeField` but can this become a problem now that time is stored separately without date information? If I recall correctly, we just had an similar issue somewhere else.

saaste (commented):
Looks awesome! Just two small questions so didn't approve just yet ☺️

api/graphql/reservations/recurring_reservation_serializers/create_serializers.py
nicobav: Nice catch, it is a magic number 🪄 indeed! I'll need to change this a bit and ask is there any max value for it.

api/graphql/reservations/recurring_reservation_serializers/create_serializers.py
nicobav: Very valid point, those begin and end date and times could be datetimes 🤔  Or perhaps we can just define a timezone to the recurring model. Nevertheless a ponder.

nicobav: We talked about this and decided that we use the configured timezone in settings. So no need for changing anything here. So the UI will need know the timezone in which the backend is working in.

api/graphql/reservations/recurring_reservation_serializers/create_serializers.py
nicobav: Removed the magic number and also renamed the field to `recurring_in_days`

saaste (approved):
Great! Let's ship this 🥇

PR #735, opened 2023-01-20
TILA-2298 | Add ability to give recurring reservation in staff create

nicobav:
## Change log
- Add recurringReservationPk to `ReservationStaffCreateMutation` 

## Other notes
In staff create reservation there should be possibility to give the recurring reservation within the creation of the reservation.

Refs TILA-2298


saaste (approved):
Cool beans!

PR #736, opened 2023-01-20
TILA-2102 | Add task for deleting recurring reservation without reservations

nicobav:
## Change log
- Adds a task called `prune_recurring_reservations` which removes recurring reservations which does not have any reservations

## Deployment reminder
- [ ] Requires to configure the task to every environment when deployed!

Refs TILA-2102


saaste (approved):
Yey! 🚢  it!

PR #738, opened 2023-01-24
TILA-2269 | Opening hours normalization

nicobav:
## Change log
- Change `start_date` and `end_date` of the `OpeningHoursType` to `datetime`
- Change then `end_time_on_next_day` to be be determined by the start and end time. 

Refs TILA-2269


opening_hours/utils/opening_hours_client.py
```diff
@@ -55,6 +55,12 @@ def get_opening_hours_class_from_time_element(
                 start.minute,
             )
         )
+
+        full_day = time_element.full_day or (
+            time_element.start_time is None and time_element.end_time is None
+        )
+        end_time_on_next_day = time_element.end_time_on_next_day or full_day
```
saaste, line 62: I have hard time understanding this logic. How is end time on the next day when `full_day` is true? Is this some "from 00:00 to 00:00" thingie or what? 🤔

I tried to read the ticket but even with that I'm still bamboozled 😄  I wonder if some human-readable comment could help us understand the logic/reasoning of the code in the future.

saaste (commented):
Looks good, but because I couldn't quite understand the logic, I didn't approve this yet. This looks like a breaking change that requires changes in the frontend, but it looks like we have a ticket for that so that is bueno 👌

opening_hours/utils/opening_hours_client.py
nicobav: yup since we want to set the end_time to 00:00 instead of 23:59 so date must be pushed by a day and set the end time to be the next day.

@vergama do you have a say for this?

vergama: These are indeed breaking changes so frontend needs to sync (but that approach was okay'ed).
So yes the idea is that `full_day` is a shorthand which means the end time is as **24:00/00:00** which technically is the first moment of the next day.
The end time could also be later in the next day (say **01:00**) but in each case we want to ensure that the date part is uniformly interpreted and passed on from the GQL.
Interpretation as **23:59** would be the other option but (a) hardly anybody marks closing times like that and (b) it would be inconsistent with time slots like **13-17.00**.

saaste (approved):
Rebase needed but otherwise looks good! 👏

opening_hours/utils/opening_hours_client.py
vergama: However... 
`if time_element.end_time_on_next_day:` should be `if end_time_on_next_day:` so that the interpretation discussed above affects also this point.

Additionally, before `start_time = timezone.localize(...)` the **full_day** flag needs to be used to override start and end because API may have returned bogus (non-null) values for them over which **full_day = true** (in that case from the API, not interpreted) should take precedence; thus:
```
if full_day:
   start = datetime.time(0)
   end = datetime.time(0)
```
Then the `end_time_on_next_day` (possibly/probably interpreted from `full_day`) will lead to correct date part of `end_time`.

PR #742, opened 2023-01-27
TILA-2165: Email template test form for Django Admin

saaste:
## Change log
- Added [django-admin-extra-button](https://github.com/saxix/django-admin-extra-buttons) package
- Remove pinned version from `django-modeltranslation`
- Added `EmailNotificationContext` that contains all supported email template fields
- Updated `ReservationEmailNotificationBuilder` to support context
- Added new button to `EmailTemplate` list view that open testing form
- Created a custom testing form that sends emails in all three languages
- Updated tests to match the changes

## Other notes
I was surprised how tricky it is to add custom functionality to Django Admin without marrying the code to a specific Django version. Some people did the same by overriding Django's templates, but that felt even more complex and requires understanding of Django's internal implementation. Django's templates seem to change quite often with versions so this route felt like a maintenance nightmare.

That being said, I'm not sure at all if this is the best way to do this. Because I had to learn so much new stuff, the solution might have a lot of details that could be improved. So keep'em comments coming! 😄 

## Deployment reminder
- No changes required.

### DIFF
--- a/email_notification/admin.py
+++ b/email_notification/admin.py
@@ -1,12 +1,18 @@
+from admin_extra_buttons.api import ExtraButtonsMixin, button
 from django.contrib import admin
 from django.core.files.uploadedfile import InMemoryUploadedFile
 from django.forms import ModelForm, ValidationError
+from django.http import HttpResponse
+from django.template.response import TemplateResponse
 
+from email_notification.email_tester import EmailTestForm
 from email_notification.models import EmailTemplate, EmailType
 from email_notification.sender.email_notification_builder import (
+    EmailNotificationContext,
     EmailTemplateValidationError,
     EmailTemplateValidator,
 )
+from email_notification.sender.senders import send_reservation_email_notification
 
 
 class EmailTemplateAdminForm(ModelForm):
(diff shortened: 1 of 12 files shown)


email_notification/admin.py
```diff
@@ -84,7 +90,70 @@ def clean_html_content_sv(self):
...
+                context.unit_name = form.cleaned_data["unit_name"]
+                context.reservation_name = form.cleaned_data["reservation_name"]
+                context.reservation_unit_name = form.cleaned_data[
+                    "reservation_unit_name"
+                ]
+                context.price = form.cleaned_data["price"]
+                context.non_subsidised_price = form.cleaned_data["non_subsidised_price"]
+                context.subsidised_price = form.cleaned_data["subsidised_price"]
+                context.tax_percentage = form.cleaned_data["tax_percentage"]
+                context.confirmed_instructions = {
```
nicobav, line 119: if you choose to write some class for this logic here, as commented above ☝️ I'd suggest that getting these dicts could have common function.

nicobav: There could be some helper methods to be added in the form class for these 🤔

nicobav (approved):
I think I like it 🕺 

Consider my comment but if you choose to keep the long function, then its a long function 👍 Mergeable stuff 👏

email_notification/admin.py
```diff
@@ -84,7 +90,70 @@ def clean_html_content_sv(self):
...
 
 @admin.register(EmailTemplate)
-class EmailTemplateAdmin(admin.ModelAdmin):
+class EmailTemplateAdmin(ExtraButtonsMixin, admin.ModelAdmin):
     model = EmailTemplate
     form = EmailTemplateAdminForm
     exclude = ["html_content"]
+
+    @button(label="Email Template Testing")
+    def template_tester(self, request):
```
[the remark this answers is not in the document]
saaste: I thought about this and ended up separating `EmailNotificationContext` into a separate file. I also created a static `from_form` function there so that `admin.py` stays cleaner. This way this whole context thingie and stuff related to it are wrapped into one class.

email_notification/admin.py
saaste: This functionality is now a part of `EmailNotificationContext`. So all we need to do in `admin.py` is:
```
context = EmailNotificationContext.from_form(form)
```

Keeps admin stuff a lot cleaner ☺️

PR #745, opened 2023-01-31
TILA-2211: Use fake context in validation

saaste:
## Change log
- Added a helper function to create context with mock content
- Use context in email template validation

## Other notes
It looks like email template validation failed because Jinja tried to render the template with an empty context. Now it is using the context I created in the previous ticket and that seems to fix the issue.

## Deployment reminder
- No changes required

### DIFF
--- a/email_notification/sender/email_notification_context.py
+++ b/email_notification/sender/email_notification_context.py
@@ -40,6 +40,49 @@ class EmailNotificationContext:
         return "\n-\n".join(instructions)
 
     @staticmethod
+    def with_mock_data() -> "EmailNotificationContext":
+        """Initialize context with mock data"""
+        context = EmailNotificationContext()
+        context.reservee_name = "Email Test"
+        context.begin_datetime = datetime.datetime(2100, 1, 1, 12, 00)
+        context.end_datetime = datetime.datetime(2100, 1, 1, 13, 15)
+        context.reservation_number = 1234567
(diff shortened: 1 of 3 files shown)


email_notification/sender/email_notification_context.py
```diff
@@ -39,6 +39,49 @@ def _get_reservation_unit_instruction_field(self, reservation, name, language):
...
+        context.unit_location = "Testikatu 99999 Korvatunturi"
+        context.unit_name = "TOIMIPISTE"
+        context.reservation_name = "TESTIVARAUS"
+        context.reservation_unit_name = "VARAUSYKSIKKÖ"
+        context.price = Decimal("12.30")
+        context.non_subsidised_price = Decimal("15.00")
+        context.subsidised_price = Decimal("5.00")
+        context.tax_percentage = 24
+        context.confirmed_instructions = {
+            "fi": "[lisäohje: hyväksytty]",
```
nicobav, line 59: Is there some reason these are inside `[]`?

nicobav (approved):
superb 👍

email_notification/sender/email_notification_context.py
saaste: Those where defined like that in [the ticket](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2172). No other reason 😄

PR #748, opened 2023-02-01
TILA-2325: Add reservation unit selection to the email testing form

saaste:
## Change log
- Add a new sub-form for selecting a reservation unit
- Fill the form with reservation unit data if it is selected

## Other notes
This is not the prettiest or most user-friendly implementation, but it does the job. This was requested yesterday in the meeting.

## Deployment reminder
- No changes required


nicobav (approved):
🚀

PR #749, opened 2023-02-01
TILA-2267 | Check the is_reservable agains states

nicobav:
The openinghoursType's is_reservable flag is just a simplified boolean value against the resource states that are reservable.

Refs TILA-2267

### DIFF
--- a/api/graphql/opening_hours/opening_hours_types.py
+++ b/api/graphql/opening_hours/opening_hours_types.py
@@ -5,6 +5,7 @@ import pytz
 from django.conf import settings
 from django.utils.timezone import get_default_timezone
 
+from opening_hours.enums import State as ResourceState
 from opening_hours.utils.opening_hours_client import OpeningHoursClient
 
 DEFAULT_TIMEZONE = get_default_timezone()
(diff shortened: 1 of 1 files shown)


api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -144,11 +145,8 @@ def resolve_opening_hours(self, info, **kwargs):
                         end_time=time.end_time.time(),
                         state=time.resource_state,
                         periods=time.periods,
-                        is_reservable=opening_hours_client.is_resource_open_for_reservations(
-                            str(self.hauki_resource_origin_id),
-                            time.start_time,
-                            time.end_time,
-                        ),
+                        is_reservable=ResourceState(time.resource_state)
```
saaste, line 148: Is `opening_hours_client.is_resource_open_for_reservations` function still needed somewhere? I checked the code and looks like it is used, but does the same logic apply in other places?

Just wondering if we could remove it and replace with this simple state check 🤔

saaste (approved):
Nice simplification! Added a note about wider refactoring, but since I don't know the context that well, I leave it up to you to decide ☺️

api/graphql/opening_hours/opening_hours_types.py
nicobav: yes, it used when validating the reservation creation (via scheduler)

PR #750, opened 2023-02-01
TILA-2246 | Show only seasonally reservable reservation units in Application Round django admin

nicobav:
## Change log
- In ApplicationRoundAdmin show only reservation units that are seasonally bookable

Refs TILA-2246


Niilis (approved):
Looks good to me :ship:

PR #753, opened 2023-02-03
TILA-2166 | Hauki admin url api for mass edits

nicobav:
## Change log
- Changes the hauki url generation to create mass-edit url (with target_resources) if needed
- adds GQL api query for getting hauki url wth option to give other reservation units to edit aside with the "main" one.

### DIFF
--- a/api/graphql/reservation_units/reservation_unit_types.py
+++ b/api/graphql/reservation_units/reservation_unit_types.py
@@ -7,6 +7,7 @@ from easy_thumbnails.files import get_thumbnailer
 from graphene import ResolveInfo
 from graphene_django import DjangoObjectType
 from graphene_permissions.mixins import AuthNode
+from graphql import GraphQLError
 
 from api.graphql.base_connection import TilavarausBaseConnection
 from api.graphql.base_type import PrimaryKeyObjectType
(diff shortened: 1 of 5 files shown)


api/graphql/tests/test_reservation_units/test_reservation_unit_hauki_url.py
```diff
@@ -0,0 +1,162 @@
...
+        return """
+            query {
+                reservationUnitHaukiUrl(pk: %s reservationUnits: [%s]) {
+                    url
+                }
+            }
+        """ % (
+            res_unit_id,
+            target_res_units,
+        )
```
Niilis, line 65: Why not use f-strings?

Niilis (commented):
Some minor comments before we can call this done. Generally, looks good though.

SonarCloud spotted a code smell as well. This should be addressed.

api/graphql/tests/test_reservation_units/test_reservation_unit_hauki_url.py
nicobav: because of double {{ and }} does not go well with my taste 😂

api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -169,12 +170,42 @@ class Meta:
...
+            self.include_reservation_units = include_reservation_units
+
+    def resolve_url(self, info, **kwargs):
         if can_manage_units(info.context.user, self.unit):
+            target_uuids = []
+            include_reservation_units = getattr(self, "include_reservation_units", None)
+
+            if include_reservation_units:
+                for runit_id in include_reservation_units:
+                    runit = ReservationUnit.objects.filter(id=runit_id).first()
```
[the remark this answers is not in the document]
nicobav: yup, only few queries to validate the existence of the ids given in. But very good point indeed. But I could make some changes actually

api/graphql/reservation_units/reservation_unit_types.py
```diff
@@ -169,12 +170,47 @@ class Meta:
...
+                    {res_unit.id for res_unit in res_units_in_db}
+                )
+
+                if difference:
+                    raise GraphQLError(
+                        "Wrong identifier for reservation unit in url generation."
+                    )
+                for res_unit in res_units_in_db.filter(unit=self.unit):
+                    target_uuids.append(res_unit.uuid)
+
```
Niilis, line 208: The updated lines look great!

You can do away with the loop here with the `values_list()` queryset function:

```python
target_uuids = res_units_in_db.filter(unit=self.unit).values_list("uuid", flat=True)
```

Niilis (changes requested):
One more minor upgrade to have the code in mint condition.

api/graphql/tests/test_reservation_units/test_reservation_unit_hauki_url.py
Niilis: Sure, this is fine!

PR #754, opened 2023-02-09
TILA-2288 | Add new reservation unit states

nicobav:
## Change log
- Separate the reservation states in the reservation unit state to be in its own enum and add some new logic and values to both.
- Update reservation unit gql to new states

Refs TILA-2288


Niilis (approved):
This looks good to me, well done with the queries! :100:

PR #760, opened 2023-02-21
TILA-1983 | Prefill rervee info from profile

nicobav:
## Change log
- Add logic to fetch user info from profile
- Use the logic in reservation create serializer for fetching prefill data.

## Deployment reminder

Refs TILA-1983


vergama:
Address reading needs the foreign address option as last resort:
```
address = self.vp_info.get("permanentForeignAddress", {})
if address:
	return {
		"postalCode": "",
		"address": address.get("streetAddress"),
		"city": "",
	}
```

vergama (commented):
`permanentForeignAddress.additionalAddress` is now read into `city` but that won't do, that value from DVV can apparently contain anything... we'll just have to disregard it and leave both postal code and city default as empty and let the customer user fill in / split from the street address on UI form

vergama (approved):

PR #761, opened 2023-02-21
TILA-2346 | Add viewer user's name and email to PersonalInfoViewLog

nicobav:
## Change log
- Add viewer's full name and email to PersonalInfoViewLog
 - Add task for removing old PersonalInfoViewLogs

## TODO
- [ ] Add the task to envs after merge

Refs TILA-2346 TILA-2347


nicobav:
Really hard to write specific tests for those migrations.. They're tested in a way in every usage of the field which are tested obviously. Would need tweak the coverage to exclude migrations perhaps.

vergama (commented):
Removal test could check that logs newer than time limit are retained. Limit is 2 years, not 1.

nicobav:
> Removal test could check that logs newer than time limit are retained. Limit is 2 years, not 1.

Nice catch! There's something to change there 👍

nicobav:
Changed the deletion threshold to two years 😇

vergama (approved):

saaste (approved):
Good stuff!

PR #763, opened 2023-02-22
Exclude migrations from coverage

nicobav:


nicobav:
At least sonar nags about coverage of migration files - perhaps also some sonar configs need to be added. Any suggestions?

saaste (approved):

saaste:
> At least sonar nags about coverage of migration files - perhaps also some sonar configs need to be added. Any suggestions?

Nothing comes to my mind 🤔  I guess we can go with this and add more exclusions later when we spot unnecessary nagging.

PR #764, opened 2023-02-22
TILA-2365 | Add check logic to user data anonymization

nicobav:
Adds logic to where user data anonymization is checked to be valid for execution.

Refs TILA-2365


vergama (approved):

aceViilee (commented):
LGTM

PR #765, opened 2023-02-24
TILA-2376 | Make subscriptionId and phone not mandaroy for reading order

nicobav:
## CHANGES
 - Order from_json method had an expectation that subscriptionId would be always present. This isn't the case and this adds handling for non existent subscription ids. Same goes for the phone.
 - Send tax percentage value without comma to verkkokauppa

Refs TILA-2376

### DIFF
--- a/merchants/verkkokauppa/order/types.py
+++ b/merchants/verkkokauppa/order/types.py
@@ -205,7 +205,7 @@ class CreateOrderParams:
                     "priceNet": str(item.price_net),
                     "priceGross": str(item.price_gross),
                     "priceVat": str(item.price_vat),
-                    "vatPercentage": str(item.vat_percentage),
+                    "vatPercentage": str(int(item.vat_percentage)),
                     "meta": [
                         {
                             "key": meta.key,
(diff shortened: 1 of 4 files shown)


merchants/verkkokauppa/order/types.py
```diff
@@ -155,12 +161,10 @@ def from_json(cls, json: Dict[str, Any]) -> "Order":
...
                     last_name=json["customer"]["lastName"],
                     email=json["customer"]["email"],
-                    phone=json["customer"]["phone"],
+                    phone=json["customer"].get("phone", ""),
                 ),
                 status=json["status"],
-                subscription_id=UUID(json["subscriptionId"])
-                if json["subscriptionId"]
-                else None,
+                subscription_id=subscription_id if subscription_id else None,
```
saaste, line 167:
```suggestion
                subscription_id=subscription_id,
```
I have a feeling `if` statement is not needed here. On line 106 it is set to UUID and if that fails, it is set to None on line 108. The value should be safe to use use as is - unless I missed something 😄

saaste (changes requested):

saaste:
I'll merge this because there is need to test the whole flow in dev. The nitpicks can be addressed later in a separate PR ☺️

saaste (approved):

PR #768, opened 2023-03-01
TILA-2345 | Fix uuid migration

nicobav:
To Recurring reservation was added UUID field which was unique. We must always handle unique field adds afterwards with custom migration code and this wasn't exception. This adds the custom setting up the uuid field for RecurringReservation.

Refs TILA-2345

### DIFF
--- a/reservations/tests/test_recurring_reservation.py
+++ b/reservations/tests/test_recurring_reservation.py
@@ -1,0 +1,26 @@
+from assertpy import assert_that
+from django.db import connection
+from django.db.migrations.executor import MigrationExecutor
+from django.test import TestCase
+
+from reservations.models import RecurringReservation
+from reservations.tests.factories import RecurringReservationFactory
+
+
+class TestRecurringReservationUUIDMigration(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.rec = RecurringReservationFactory()
+        cls.rec_too = RecurringReservationFactory()
+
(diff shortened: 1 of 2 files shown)


reservations/tests/test_recurring_reservation.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class TestRecurringReservationUUIDMigration(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.rec = RecurringReservationFactory()
+        cls.rec_too = RecurringReservationFactory()
+
+        migration = MigrationExecutor(connection)
+        migration.migrate([("reservations", "0044_recurringreservation_description")])
```
saaste, line 17: My understanding is that **all** migrations are run when test run is started. How running individual migrations after that works?

If my understanding is correct, when this test is being run, the database already has all these new fields, including `uuid`. I'm bamboozled! 🤔

saaste (commented):
The fix looks great, but I want to understand migration tests before approving this 😄

reservations/tests/test_recurring_reservation.py
nicobav: Yes they are run once in my understanding and that's why I'm reverting the migrations

nicobav: the main reason I have this test is that it contains now existing data rows when doing the uuid assigning

saaste (approved):

PR #769, opened 2023-03-01
Add correct resolve_user path

nicobav:
The resolver_user path was changed within other changes/adds but the settings wasn't updates.
This updates the settings to have correct path.


saaste (approved):
One-liners FTW! 💎

PR #774, opened 2023-03-08
TILA-2370 | Save profile id when saving birthday

nicobav:
## Change log
- Adds new field to user "profile_id" 
- Reads profile id when resolving user's birthday

Refs TILA-2370


saaste (approved):
Good stuff! 👍

PR #777, opened 2023-03-10
TILA-2366 | Change default value for reservation_notification

nicobav:
## Change log
- Change User's reservation notification default value to Only handling required 
- Change existings "none" values to the new default.

Refs TILA-2366


saaste (approved):
Veeery noice! 👍 
Looks like this requires a rebase to get rid of the build errors.

PR #779, opened 2023-03-13
TILA-2397: Fix invalid reservation time in payment order

saaste:
## Change log
- Fixed invalid reservation time in PaymentOrder

## Other notes
The old code didn't use get_default_timezone() so we ended up having different times in our database and in web shop.

## Deployment reminder
- No changes required

### DIFF
--- a/merchants/verkkokauppa/helpers.py
+++ b/merchants/verkkokauppa/helpers.py
@@ -31,10 +31,14 @@ def get_formatted_reservation_time(reservation: Reservation) -> str:
     Weekday is localized based on user's preferred language, but rest
     of the format is always the same: ww dd.mm.yyyy hh:mm-hh:mm
     """
+    begin = reservation.begin.astimezone(get_default_timezone())
+    end = reservation.end.astimezone(get_default_timezone())
+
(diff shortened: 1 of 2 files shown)


merchants/verkkokauppa/helpers.py
```diff
@@ -31,10 +31,14 @@ def get_formatted_reservation_time(reservation: Reservation) -> str:
     Weekday is localized based on user's preferred language, but rest
     of the format is always the same: ww dd.mm.yyyy hh:mm-hh:mm
     """
+    begin = reservation.begin.astimezone(get_default_timezone())
```
nicobav, line 34: Our default timezone is `Europe/Helsinki` is that also what is in the verkkokauppa? Or should it be converted to UTC?

saaste: This is just a additional meta field we add to the order. We just send a string to Verkkokauppa and that string is displayed in the payment summary view. So we don't really send datetime to Verkkokauppa and it is not using this field internally.

nicobav (approved):
GR8! 🚀

PR #781, opened 2023-03-13
TILA-2377 | Make hauki reader handle unknown resource states

nicobav:
Make hauki reader to be failproof when encountering unknown resource states. Previously it blew up when non-existent state was present.  This makes it default to UNDEFINED status.

Refs TILA-2377


saaste (approved):
Grrreat!
Rebase from `main` should now solve the issue with `bandit`.

PR #782, opened 2023-03-14
TILA-2401 | Reservation state BEHALF

nicobav:
## Change log
- Add new ReservationState BEHALF
- Restrict reservation creation for staff users to allow only types  BEHALF, BLOCKED and STAFF
   - So NORMAL is not allowed.

Refs TILA-2401


api/graphql/reservations/reservation_serializers/staff_create_serializers.py
```diff
@@ -180,6 +180,21 @@ def validate(self, data):
...
+    def validate_type(self, reservation_type):
+        allowed_types = [
+            ReservationType.BLOCKED,
+            ReservationType.STAFF,
+            ReservationType.BEHALF,
+        ]
+
+        if reservation_type not in allowed_types:
+            raise ValidationErrorWithCode(
+                f"Reservation type {reservation_type} is not allowed in this mutation.",
```
saaste, line 192: The error message could list the allowed types to help solve the problem.

saaste (approved):
Good stuff! Just one suggestion 👏

PR #783, opened 2023-03-14
TILA-2101 | Make possible to empty RecurringReservation description

nicobav:
## Change log
- Make possible to empty the `RecurringReservation` description through the gql api mutations

Refs TILA-2101


saaste (approved):
Solid! 👍

PR #785, opened 2023-03-16
TILA-2170: Remove unnecessary emails

saaste:
## Change log
- Some of the emails are not needed anymore so I removed them

## Deployment reminder
- No changes required


nicobav (approved):
Nice 👍

PR #786, opened 2023-03-16
TILA-2162 | Reservation deny validation rule additions

nicobav:
## Change log
- Makes possible to deny also Reservations with state of CONFIRMED
- Makes impossible to deny reservation that end time has passed
- Send emails only when Reservation type is NORMAL


saaste (approved):
Looks awesome! Nothing to add 🚀

PR #788, opened 2023-03-17
TILA-2406 | Don't use graphiql when DEBUG is false

nicobav:
This excludes graphiql from staging and production. And we want it so.

Refs TILA-2406


saaste (approved):
Looks great!
Why did we want this? 😄

nicobav:
> Looks great! Why did we want this? 😄

Even though pubic api I think less exposure is better in production environment. What I can see in the pfofile there is also the graphiql excluded from the production.

PR #789, opened 2023-03-17
TILA-2410 | Separate reservationUnitImage gql to own endpoint

nicobav:
This because it needs to support image upload and that's going to have the csrf_exempt.

Refs TILA-2410

### DIFF
--- a/api/graphql/tests/test_reservation_unit_image.py
+++ b/api/graphql/tests/test_reservation_unit_image.py
@@ -16,13 +16,13 @@ from reservation_units.tests.factories import (
     ReservationUnitImageFactory,
 )
 
-DEFAULT_GRAPHQL_URL = "/graphql/"
-
 
 @override_settings(CELERY_TASK_ALWAYS_EAGER=True)
 class ReservationUnitImageCreateTestCase(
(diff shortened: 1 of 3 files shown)


api/graphql/tests/test_reservation_unit_image.py
```diff
@@ -15,12 +15,12 @@
     ReservationUnitImageFactory,
 )
 
-DEFAULT_GRAPHQL_URL = "/graphql/"
-
 
 class ReservationUnitImageCreateTestCase(
     GrapheneTestCaseBase, GraphQLFileUploadTestCase
 ):
+    GRAPHQL_URL = "/graphql/images/"
```
saaste, line 22: Looks like this could be in `GrapheneTestCaseBase` just to avoid code duplication in each TestCase class 🤔

saaste (approved):
Noice and clean. One suggestion to remove code duplication 👍

api/graphql/tests/test_reservation_unit_image.py
nicobav: It actually applies only to the reservationUnitImage gql since they're now in their own url so applying in the base level wouldn't do.

PR #791, opened 2023-03-22
TILA-2291 | ReservationStatistics export from admin

nicobav:
## Change log
- Adds ability to export `ReservationStatistic` model to csv file (with django-import-export package)
- Makes possible to filter by spesific date range in django admin (with the django-admin-rangefilter package)

Refs TILA-2291


saaste (approved):
Whoa! That turned out to be way simpler than expected! 🤯 
Good job! 🥇

PR #792, opened 2023-03-23
TILA-1680 | ReservationUnit export from admin

nicobav:
## Change log
- Makes possible for export the reservation unit csv from django admin
- Adds reservation_unit.state and reservation_state to the export csv

Refs TILA-1680


nicobav:
Didn't write a test for admin function. I should and would

saaste (approved):
Good stuff! 👏

PR #796, opened 2023-03-28
TILA-2442 | Make QueryPerformanceOptimizerMixin to be inherited in ReservationUnitType

nicobav:
## Change log
- Adds `QueryPerformanceOptimizerMixin` to `ReservationUnitType`
   - Some initial configuration to the queries
   
Refs TILA-2442


saaste (approved):
I need to spend some time to really understand how these query optimizations work. But it looks good with the knowledge I currently have so let's 🚢  and see what is the effect ☺️

PR #797, opened 2023-03-29
TILA-2428: Fix update_expired_orders

saaste:
## Change log
- Added checks for case where payment does not exist
- Unit test that catches the old problem and confirms the fix

## Other notes
It did not handle the case where payment does not exist in Verkkokauppa. I added the checks and if status cannot be checked, it is treated as expired.

## Deployment reminder
- No changes required


nicobav (approved):
Yes sir 🍪

PR #798, opened 2023-03-29
TILA-2464: Fix order creating issue with wrong datetime format

saaste:
## Change log
- Replaced `isoformat()` with `strftime()` and custom format

## Other notes
Order Experience API does not support standard ISO 8601 format with UTC offset. That is why we have to manually define a custom format. This is a bit risky since Verkkokauppa assumes that this date has to be in Europe/Helsinki timezone. Any change in timezones can cause problems.

## Deployment reminder
- No changes required

### DIFF
--- a/merchants/verkkokauppa/order/types.py
+++ b/merchants/verkkokauppa/order/types.py
@@ -192,7 +192,10 @@ class CreateOrderParams:
             "namespace": self.namespace,
             "user": str(self.user),
             "language": self.language,
-            "lastValidPurchaseDateTime": self.last_valid_purchase_datetime.isoformat(),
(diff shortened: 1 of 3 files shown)


merchants/verkkokauppa/order/types.py
```diff
@@ -192,7 +192,10 @@ def to_json(self) -> Dict[str, Any]:
             "namespace": self.namespace,
             "user": str(self.user),
             "language": self.language,
-            "lastValidPurchaseDateTime": self.last_valid_purchase_datetime.isoformat(),
+            # Order Experience API does not support standard ISO 8601 format with UTC offset
```
nicobav, line 195: 👍

nicobav (approved):
LGTM :shipit:

PR #800, opened 2023-03-29
TILA-2442 | ReservationUnits query performance configurations

nicobav:
## Change log
- Fix bug in `QueryPerformanceOptimizerMixin` class 
   - no prefetches were executed if used advanced optimization and the field did not exist in actual query.
- Fix Correct field_name for querying `reservationUnits` (was `reservationUnit`)
- Adds more query optimizations to reservationUnits query

Refs TILA-2442

## Other notes
There might be some things yet to do with QueryPerformanceOptimizer. Will see.


saaste (approved):
Looks good! I still don't quite understand the query optimization, but figuring that out is next on my todo list 😄

PR #801, opened 2023-03-30
TILA-2467: Change merchant form fields to be required

saaste:
## Change log
- Changed merchant form fields to be required
- Fixed GitHub action. This should fix SonarCloud as well.

## Other notes
Verkkokauppa requires these fields so they must be required on our side as well.

## Deployment reminder
- No changes required


nicobav (approved):
Looks nice🍻

PR #805, opened 2023-04-04
TILA-2448 | ReservationStatistic adds

nicobav:
## Change log
- Adds bunch of field to `ReservationStatistic` model
- Combine age_group min and max to field called age_group_name
- Save primary_reservation_unit as it was not saved before 🐛 

Refs TILA-2448


saaste (approved):
Awwwwyeah! :shipit:

PR #807, opened 2023-04-05
TILA-2471 | Staff can deny self created reservations

nicobav:
## Change log
- Adds new permission class  `ReservationDenyPermission` and use in deny mutation.
  - permission checks for staff user can either handle reservations or can create staff AND is self created one.

## Other notes
There should be possible for staff members to deny self created (staff created reservations).
Makes new permission for denying based on the
ReservationHandlingPermission called ReservationDenyPermission and use in deny mutation.

Refs TILA-2471


saaste (approved):
Nice and clean! 🧼

PR #809, opened 2023-04-05
TILA-2443 | Add db indexes to fields used in filtersets

nicobav:
## Change log
- Adds db_index=True to fields used in filtersets excluding char/textfields and date fields.


saaste (approved):
🏎️ !

PR #814, opened 2023-04-12
TILA-2510 | Generate image urls on reservation unit image save

nicobav:
## Change log
- Generates `ReservationUnitImage`'s urls when its saved
- Changes the `update_url` task to take optional parameter for giving single res unit image

Refs TILA-2510


saaste (approved):
Veeery noice! 👍

PR #815, opened 2023-04-13
TILA-2497 | Application and ApplicationEvent gql additions

nicobav:
## Change log
- Adds possibility to query `Applications` with `appliedCount`
- Adds possibility to query `ApplicationEvents` with `appliedCount`
- Adds possibility to query `ApplicationRounds` with `pk`

### DIFF
--- a/api/graphql/application_rounds/application_round_types.py
+++ b/api/graphql/application_rounds/application_round_types.py
@@ -83,7 +83,6 @@ class ApplicationRoundType(AuthNode, PrimaryKeyObjectType):
     class Meta:
         model = ApplicationRound
         fields = [
-            "id",
             "reservation_units",
             "application_period_begin",
             "application_period_end",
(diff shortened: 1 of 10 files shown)


api/graphql/application_rounds/application_round_types.py
```diff
@@ -83,7 +83,6 @@ class Input:
     class Meta:
         model = ApplicationRound
         fields = [
-            "id",
```
saaste, line 86: Removing this looks like a breaking change. Hopefully client is not using this field.

api/graphql/application_rounds/application_round_types.py
```diff
@@ -102,12 +101,6 @@ class Meta:
             "target_group",
         ] + get_all_translatable_fields(model)
 
-        filter_fields = {
```
saaste, line 105: Same as above. Do we know that client is not using these filters?

saaste (approved):
Approved with two comments about breaking change. If field/filters are not in use, this is good to go (as soon as tests are green 😄 )! 👍

api/graphql/application_rounds/application_round_types.py
nicobav: It shouldn't be in use and I've discussed that these ids are not our way to go, rather the `pk`

api/graphql/application_rounds/application_round_types.py
nicobav: Client is using the REST api currently for the application rounds and there are no such filters in use so this should be safe. This was only missing the `pk` filter as was in REST /application_round/<pk>

saaste:
🚢  it! 👏

PR #816, opened 2023-04-13
TILA-2450 | Read profile node id regardless of LOA level

nicobav:
## Change log
 - Reads the user's open city profile node id regardles of the LOA level

Refs TILA-2450

### DIFF
--- a/users/tests/test_basic_info_resolver.py
+++ b/users/tests/test_basic_info_resolver.py
@@ -1,13 +1,19 @@
 from typing import Dict, List
+from unittest import mock
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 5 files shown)


saaste (approved):
Grrrreat! :shipit: 

SonarCloud is nagging about the code coverage. Looks like we have no tests for failure cases. Might be useful to add and also makes SonarCloud happy ☺️

nicobav:
> Grrrreat! :shipit:
> 
> SonarCloud is nagging about the code coverage. Looks like we have no tests for failure cases. Might be useful to add and also makes SonarCloud happy ☺️

Very much agree 🤝

PR #818, opened 2023-04-14
TILA-2485: Add reservation number and language support

saaste:
## Change log
- Add reservation number to webshop order meta fields
- Add language support to meta field labels

## Other notes
The implementation might still change because I'm not sure if language should come from the reservation user or from `reservee_language` field

## Deployment reminder
- No changes required

### DIFF
--- a/merchants/verkkokauppa/helpers.py
+++ b/merchants/verkkokauppa/helpers.py
@@ -9,6 +9,7 @@ from sentry_sdk import capture_exception, push_scope
 
 from api.graphql.validation_errors import ValidationErrorCodes, ValidationErrorWithCode
 from applications.models import CUSTOMER_TYPES
+from merchants.verkkokauppa.exceptions import UnsupportedMetaKey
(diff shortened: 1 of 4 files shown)


nicobav (approved):

merchants/verkkokauppa/helpers.py
```diff
@@ -59,6 +59,26 @@ def get_validated_phone_number(phone_number: str) -> str:
...
+            "sv": "Bokningstiden",
+        },
+        "reservationNumber": {
+            "fi": "Varausnumero",
+            "en": "Booking number",
+            "sv": "Bokningsnummer",
+        },
+    }
+    if key not in labels:
+        raise Exception(f"Invalid meta label key '{key}")
```
nicobav, line 76: Consider creating a own Exception type for this

saaste: Done! I also added more tests 👍

PR #821, opened 2023-04-17
TILA-2517: Log invalid webhook calls

saaste:
## Change log
- Fixed webhook payload. Webshop documentation was wrong and that lead to wrong implementation
- Log invalid webhook calls to sentry
- Updated tests

## Other notes
In normal use webhooks should not fail. Failures are indication of changes in webshop side or someone bombarding API with malicious payload. By sending these to Sentry we get proper alerts when things go haywire.

## Deployment reminder
- No changes required


vergama (commented):
This looks ok, but in GetPaymentError capture should remote-id go to extra like in GetOrderError?

vergama (approved):

PR #824, opened 2023-04-18
TILA-1518: Skip allocation tests

saaste:
## Change log
- Upgrade urllib3 package in requirements_dev.txt
- Disable allocation tests

## Other notes
For some reason the tests are failing on my mac. The feature is not used so I just disabled these tests. We could remove the whole feature in the future but that might require a bit more work especially if there are some code written on the client side too.

## Deployment reminder
- No changes required


nicobav (approved):
It is what it is 👍

PR #826, opened 2023-04-18
TILA-2451 | Reservation unit state helper fixes

nicobav:
## Change log
- changes the ReservationUnitReservationStateHelper to return
reservation state as RESERVATION CLOSED when reservation unit's
reservation begins and *_ends dates are in past and the same
- Returns HIDDEN state when publish dates in the past and the same
- Shows correctly scheduled publishing reservation units when publish dates in the future

## Other notes
See commit messages for details

Refs TILA-2451

### DIFF
--- a/api/graphql/tests/test_reservation_units/test_reservation_units_state_filter.py
+++ b/api/graphql/tests/test_reservation_units/test_reservation_units_state_filter.py
@@ -222,3 +222,33 @@ class ReservationUnitsFilterStateTestCase(ReservationUnitQueryTestCaseBase):
         assert_that(self.content_is_empty(content)).is_false()
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
+
+    def test_filtering_by_scheduled_publishing_when_begin_after_end(self):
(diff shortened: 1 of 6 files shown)


api/graphql/tests/test_reservation_units/test_reservation_units_state_filter.py
```diff
@@ -222,3 +222,33 @@ def test_filtering_by_hidden(self):
         assert_that(self.content_is_empty(content)).is_false()
         assert_that(content.get("errors")).is_none()
         self.assertMatchSnapshot(content)
+
+    def test_filtering_by_scheduled_publishing_when_begin_after_end(self):
```
saaste, line 226: Just interested: in which case having begin time being after end time makes sense? 🤔

saaste (approved):
Great stuff! 👏

api/graphql/tests/test_reservation_units/test_reservation_units_state_filter.py
nicobav: When it's scheduled for publishing 🥁 So this case is when the reservation unit it is scheduled for closing at some time and to be republished after that.

PR #827, opened 2023-04-19
TILA-2525: Remove timestamp fields from validation

saaste:
## Change log
- Remove timestamp fields from webhook validation

## Other notes
We are not using these fields so just to avoid shadowboxing with all the mixed implementations on webshop side, I just removed'em.

## Deployment reminder
- No changes required


nicobav (approved):
🧹

PR #828, opened 2023-04-19
TILA-2465: Check refund status from webshop API on refund webhook

saaste:
## Change log
- Added client support for the new endpoint in webshop
- Use client in the refund webhook handler

## Other notes
This endpoint wasn't available before but now it is so we are using it.

## Deployment reminder
- No changes required

### DIFF
--- a/merchants/verkkokauppa/payment/requests.py
+++ b/merchants/verkkokauppa/payment/requests.py
@@ -15,11 +15,13 @@ from ..constants import METRIC_SERVICE_NAME, REQUEST_TIMEOUT_SECONDS
 from ..exceptions import VerkkokauppaConfigurationError
 from .exceptions import (
     GetPaymentError,
+    GetRefundStatusError,
     ParsePaymentError,
     ParseRefundError,
+    ParseRefundStatusError,
     RefundPaymentError,
 )
-from .types import Payment, Refund
+from .types import Payment, Refund, RefundStatusResult
 
 
 def _get_base_url():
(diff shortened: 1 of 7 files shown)


merchants/verkkokauppa/payment/requests.py
```diff
@@ -64,6 +71,49 @@ def get_payment(order_id: UUID, namespace: str, get=_get) -> Optional[Payment]:
...
+
+        # API documentation says it returns 404 but in reality it
+        # returns 500 with error code. They have plans to fix this
+        # but just to avoid breaking things with unnotified changes we
+        # are handling both cases. 500 handling can be removed after
+        # webshop API is updated.
+        if (
+            response.status_code == 500
+            and errors
+            and errors[0].get("code") == "failed-to-get-refund-payment-for-order"
```
nicobav, line 102: This could blow up with index out of bounds

nicobav, line 102: 'len(errors) > 0'

nicobav (changes requested):
Looks good 👍 Found one miney thing to perhaps take care of

nicobav (approved):
Very nice sir 🥂 

My comments about index out of bounds was highly exaggerated

merchants/verkkokauppa/payment/requests.py
saaste: As discussed in Slack, Python's truthful tests return False for empty lists so this should not cause index out of bounds error. However, because this check is not super obvious, I'll change it to `len(errors) > 0` as you suggested. It is more clear for everyone 👍

PR #829, opened 2023-04-19
TILA-2271: Separate Dockerfile for running the service locally

saaste:
## Change log
- Created separate Dockerfile for local running
- Updated README to contain simplified instructions

## Other notes
Running core locally can be a pain in the 🍑 . So, I decided to simplify it for the frontend devs. We don't need all that RedHat crap and service runs just fine in with standard python image. This drops some configuration and RedHat requirements and service can be run with a single command.

## Deployment reminder
- No changes required

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,0 +1,13 @@
+FROM python:3.8.13-bullseye
+
+RUN pip install --upgrade pip
+COPY ./requirements.txt .
+RUN pip install -r requirements.txt
+
+COPY . /tvp
+WORKDIR /tvp
+
+RUN apt-get update && apt-get -y install libgdal-dev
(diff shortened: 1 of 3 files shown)


Dockerfile.local
```diff
@@ -0,0 +1,13 @@
+FROM python:3.8.13-bullseye
+
+RUN pip install --upgrade pip
+COPY ./requirements.txt .
+RUN pip install -r requirements.txt
+
+COPY . /tvp
+WORKDIR /tvp
+
+RUN apt-get update && apt-get -y install libgdal-dev
```
nicobav, line 10: so these gdallies can be found also within this image. Noice. Since afaik this was the main blocker for using python images... Ubuntu would've also an option but me likes this the best.

PR #832, opened 2023-04-24
TILA-2531 | Read only regular user's profile_id

nicobav:
## Change log
- Add util property check whether user is supposed to be staff user (since django's is_staff won't work in this case)
- Read profile id only for regular users - not staff users
- Catch exceptions reading profile_id and send'em to sentry 

Refs TILA-2531

### DIFF
--- a/users/utils/open_city_profile/birthday_resolver.py
+++ b/users/utils/open_city_profile/birthday_resolver.py
@@ -7,7 +7,10 @@ from helusers.user_utils import get_or_create_user
 from requests import RequestException
 from sentry_sdk import capture_exception, capture_message
 
-from users.utils.open_city_profile.basic_info_resolver import ProfileNodeIdReader
+from users.utils.open_city_profile.basic_info_resolver import (
+    ProfileNodeIdReader,
+    ProfileReadError,
+)
 from users.utils.open_city_profile.mixins import ProfileReaderTokenMixin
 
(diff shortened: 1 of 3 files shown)


users/utils/open_city_profile/birthday_resolver.py
```diff
@@ -16,10 +19,14 @@
...
-        user.profile_id = profile_id_reader.get_user_profile_id()
-        user.save()
+    if not user.profile_id and not user.has_staff_permissions:
+        try:
+            profile_id_reader = ProfileNodeIdReader(request)
+            user.profile_id = profile_id_reader.get_user_profile_id()
+        except ProfileReadError as e:
+            capture_exception(e)
+        else:
+            user.save()
```
saaste, line 29: I did not know one can use `else` with `try`!

I'm struggling a bit to understand the benefit, but I guess you did it like this because you want `try` block to be about profile reading and not about saving the user. So any errors thrown from `user.save()` and not handled. Or did I miss something?

saaste (approved):
Good stuff! Very much approved, I just added a comment about my own ponderings 😄

users/utils/open_city_profile/birthday_resolver.py
nicobav: Yes exactly. Try to read the profile id and if success save it. Of course it could also be inside the try block but I like it like this.

PR #833, opened 2023-04-25
TILA-1518: Remove task scheduling from the code

saaste:
## Change log
- Removed task scheduling from the code

## Other notes
We schedule tasks using Admin UI. It is quite likely that these lines are the reason why we see so many scheduling errors in Sentry.

## Deployment reminder
- No changes required


nicobav (approved):
Nicey🧹

PR #836, opened 2023-05-02
TILA-2536: Save confirmation only if API call is successful

saaste:
## Change log
- Fixed invalid order of operations

## Other notes
We had an issue where reservation was marked as confirmed even API call failed. We should save confirmation only if API call is successful (or when API is not called). This PR fixes the issue.

## Deployment reminder
- No changes required


vergama:
Can this be removed?
`Reservation.objects.filter(pk=self.instance.pk).update(`
`   state=STATE_CHOICES.WAITING_FOR_PAYMENT`
`)`
`self.instance.refresh_from_db()`

nicobav (approved):

nicobav:
> Can this be removed? `Reservation.objects.filter(pk=self.instance.pk).update(` ` state=STATE_CHOICES.WAITING_FOR_PAYMENT` `)` `self.instance.refresh_from_db()`

Good guestion

saaste:
> Can this be removed? `Reservation.objects.filter(pk=self.instance.pk).update(` ` state=STATE_CHOICES.WAITING_FOR_PAYMENT` `)` `self.instance.refresh_from_db()`

Removed. It was a leftover from the previous implementation. Good catch! 👍

PR #837, opened 2023-05-03
Tila 2530/update task schedules

saaste:
## Change log
- Fixed null field handling (web shop does not allow them even they are in the response we get from them)
- Increased task schedule interval to 10 minutes and fixed a tiny parameter issue
- Wait longer for order expiration if user moved from web shop to payment pipeline
- Fixed `CREATED` enum string value
- Added a new required field to `CreateOrUpdateAccountingParams` JSON

## Deployment reminder
- [ ] Increase `VERKKOKAUPPA_ORDER_EXPIRATION_MINUTES` to 10 minutes in each environment


merchants/pruning.py
```diff
@@ -32,6 +32,14 @@ def update_expired_orders(older_than_minutes: int) -> None:
...
             elif result and result.status == WebShopPaymentStatus.PAID_ONLINE.value:
                 order.status = OrderStatus.PAID
+            elif (
+                result
+                and result.status == WebShopPaymentStatus.CREATED.value
+                and result.timestamp
+                and result.timestamp > expired_datetime
+            ):
+                continue
+                # user has entered actual payment phase from web shop; skip expire until more time passed
```
nicobav, line 42: This comment should be in above the `continue`

nicobav (approved):

PR #840, opened 2023-05-04
TILA-2540: Remove padding zeroes from date

saaste:
## Change log
- Removed padding zeroes from the date

## Other notes
I did not know that Python's date formatting does not support values without padding zeroes. So, ended up building the proper format by hand.

## Deployment reminder
- No changes required

### DIFF
--- a/merchants/verkkokauppa/helpers.py
+++ b/merchants/verkkokauppa/helpers.py
@@ -39,9 +39,11 @@ def get_formatted_reservation_time(reservation: Reservation) -> str:
 
     preferred_language = reservation.reservee_language or "fi"
     weekday = localized_short_weekday(begin.weekday(), preferred_language)
+    date = "{d.day}.{d.month}.{d.year}".format(d=begin)
+    start_time = begin.strftime("%H:%M")
     end_time = end.strftime("%H:%M")
 
-    return begin.strftime(f"{weekday} %d.%m.%Y %H:%M-{end_time}")
+    return begin.strftime(f"{weekday} {date} {start_time}-{end_time}")
 
 
 def get_validated_phone_number(phone_number: str) -> str:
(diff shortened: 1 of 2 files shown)


merchants/verkkokauppa/helpers.py
```diff
@@ -39,9 +39,11 @@ def get_formatted_reservation_time(reservation: Reservation) -> str:
 
     preferred_language = reservation.reservee_language or "fi"
     weekday = localized_short_weekday(begin.weekday(), preferred_language)
+    date = "{d.day}.{d.month}.{d.year}".format(d=begin)
```
nicobav, line 42: 🪄 `date = begin.strftime("%-d.%-m.%Y")` 🪄

nicobav (changes requested):

merchants/verkkokauppa/helpers.py
saaste: Initially I was going to use this solution but apparently this is [OS-specific solution](https://www.enthought.com/blog/no-zero-padding-with-strftime/). Looks like this does not work on Windows so that is why I chose to not use it. I'd prefer to use a solution that works on all OSs. What are your thoughts on this? ☺️

nicobav: It is good point indeed. However my two cents is that this app is supposed to be run in *nix based system and so our container images are linux images (distros varies).
I don't know how well this would run in windows but I guess there would be some oddities.

saaste: This probably isn't issue in production since we know it is a *nix based system. I'm more worried about the future developer who uses Windows machine without Linux subsystem 😅  This is going to fail on their machine and same happens with unit tests. I read more about this and it looks like this is not only a Windows issue and can happen in *nix systems as well, depending on the setup/configuration. Very hazy thing overall 🤪 

So I think I'll keep the `.format` solution. It is a bit ugly but it is guaranteed to work on all machines and won't backfire in the future 😄

PR #842, opened 2023-05-04
TILA-2561 | Add task to rebuild space hierarchy and rebuild on save

nicobav:
Sometimes the tree hierarchies gets corrupted and need a rebuilding. This adds task for rebuilding the hierarchies and does partial_rebuild when saving Space instance.


saaste (approved):
I never realized this is a thing! Learned something new today 🧠 
Great stuff! 👍

PR #844, opened 2023-05-05
TILA-2554: Limit non_subsidised_price decimals to 2

saaste:
## Change log
- Set decimal limit to 2

## Other notes
I started to wonder if we have some logic that need to be adjusted as well. Out of my head I couldn't remember anything and test were all green so model change was the only thing I did.

## Deployment reminder
- No changes required


nicobav (approved):
:shipit:

PR #845, opened 2023-05-05
TILA-2503 | Mutation for adjusting time as a staff user

nicobav:
## Change log
- Adds mutation for adjusting reservation time. For staff users only

Refs TILa-2503

### DIFF
--- a/api/graphql/reservations/reservation_serializers/staff_adjust_time_serializers.py
+++ b/api/graphql/reservations/reservation_serializers/staff_adjust_time_serializers.py
@@ -1,0 +1,87 @@
+import datetime
+
+from django.utils.timezone import get_default_timezone
+
+from api.graphql.base_serializers import PrimaryKeyUpdateSerializer
+from api.graphql.reservations.reservation_serializers.mixins import (
+    ReservationSchedulingMixin,
+)
+from api.graphql.validation_errors import ValidationErrorCodes, ValidationErrorWithCode
+from reservations.email_utils import send_reservation_modified_email
+from reservations.models import STATE_CHOICES as ReservationState
+from reservations.models import Reservation, ReservationType
+
+DEFAULT_TIMEZONE = get_default_timezone()
+
+
+class StaffReservationAdjustTimeSerializer(
+    PrimaryKeyUpdateSerializer, ReservationSchedulingMixin
+):
+    class Meta:
+        model = Reservation
+        fields = [
+            "pk",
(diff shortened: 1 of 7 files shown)


api/graphql/reservations/reservation_serializers/staff_adjust_time_serializers.py
```diff
@@ -0,0 +1,87 @@
...
+        if new_begin > new_end:
+            raise ValidationErrorWithCode(
+                "End cannot be before begin",
+                ValidationErrorCodes.RESERVATION_MODIFICATION_NOT_ALLOWED,
+            )
+
+        now = datetime.datetime.now(tz=DEFAULT_TIMEZONE)
+
+        if now.hour <= 1:
+            now = now - datetime.timedelta(days=1)
```
saaste, line 75: What is the reason we do this? 🤔

saaste (commented):
There was a one part I couldn't figure out the reasoning for so did not yet approve. But other than that looks great! 👏

api/graphql/reservations/reservation_serializers/staff_adjust_time_serializers.py
vergama: The reason is that we allow modification/correction of the reservation time even after it has passed as long as the change is done within the same day (date) or in advance. But in case someone is working really long hours, a short "grace period"/leeway is added so that the first hour after the midnight is considered part of the day that strictly speaking just ended. That specific cut-off is sort of arbitrary but on the other hand should still prevent changing reservations that really are in the past (changes done the next morning or later; `RESERVATION_MODIFICATION_NOT_ALLOWED`). Also if one _is_ allowed to modify the reservation, it cannot be moved further into the past at the date level, but can be kept on the date that it was already set; hence if the modification is done post-midnight of the day that it occurred, also need to validate against that special interpretation of "**now**" so that adding +1 day to the reservation time & date that has actually passed is not forced (but the hours within the date (yesterday) can be changed to something else; `RESERVATION_BEGIN_IN_PAST`). Of course this kind of validation also allows moving future reservations slightly to the past, for which there may be no real use case, but insuring against that would maybe over-complicate things... (Actually a case might be that the wrong reservation was first moved into the future and now needs to be put back where it was as the error was immediately noticed.) Balance of assuming the handlers use common sense, while trying to avoid unintentional modifications to past reservations.

saaste (approved):

PR #846, opened 2023-05-08
TILA-2516: Adjust profile address parser

saaste:
## Change log
- Adjust address parsing

## Other notes
Sometimes we had addresses where city value container something like "Helsinki FI". This happened with foreign addresses that can have all kinds of weird address formats. Now, only address field is used from the foreign address and other fields are ignored.

## Deployment reminder
- No changes required


vergama (approved):
Looks good!

PR #850, opened 2023-05-11
TILA-2582: Update Django to 4.2.1 and Python to 3.11

saaste:
## Change log
- Updated Django to 4.2.1
- Updated a ton of packages to latest version
- Updated code to work with upgraded packages and with Python 3.10

## Other notes
Well that was an adventure, but at least it works on my machine. All tests are green and I also tested Django Admin and GQL API and didn't find any issues.

PostgreSQL 11 is not supported anymore so for local development image must be upgraded to version 13. This might mean database reset unless you want to somehow backup and restore all the data. Existing database files are not compatible with version 13.

## Deployment reminder
- [ ] Update environments to use Python 3.10


nicobav:
> Existing database files are not compatible with version 13.

🙈

api/graphql/opening_hours/opening_hours_types.py
```diff
@@ -1,7 +1,7 @@
 import datetime
+from datetime import timezone
```
nicobav, line 2: perhaps we should lean towards to `from django.utils import timezone`?

reservation_units/tests/test_admin.py
```diff
@@ -30,7 +31,7 @@ def test_reservation_kind_is_direct_are_excluded_from_queryset(self):
             },
         )
 
-        queryset, _ = self.radmin.get_search_results(get_request, self.queryset, "")
```
nicobav, line 33: Why the `self` reference does not work anymore?

nicobav (approved):

.github/actions/setup-python-env/action.yml
```diff
@@ -7,10 +7,10 @@ runs:
         sudo apt update
         sudo apt install libpq-dev gdal-bin graphviz graphviz-dev -y
       shell: bash
-    - name: Set up Python 3.8
+    - name: Set up Python 3.10
       uses: actions/setup-python@v4
       with:
-        python-version: 3.8
+        python-version: 3.10.10
```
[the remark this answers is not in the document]
saaste: Yeah, I tried with 3.10 and it complained that no such version is found and provided a link with all the supported versions. There was 3.10.11 available too so we could use that. I just went with 3.10.10 since that was the version I tested on my machine 😄

.github/workflows/build.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
+      - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.10.10
       - name: Install black and isort
         run: |
           pip install isort==5.6.4
-          pip install black==20.8b1
+          pip install black==22.12.0
```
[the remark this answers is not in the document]
saaste: Indeed! The only change was the two spaces removed from the comments 😅

reservation_units/tests/test_admin.py
saaste: I couldn't fully figure that out. The error message I got was:
```
cannot pickle 'module' object
```

Based on my searches it seems to be somehow related to Python not being able to serialize the class or something like that. It was kinda weird and I couldn't figure out any proper way to fix it. I tried to debug the test and the class instance was initialized properly in the setup. But as soon as we hit the test and tried to access the instance, we got the error. We didn't have any issues with the other variables.

If you can figure out a proper fix for this, please let me know 😄

api/graphql/opening_hours/opening_hours_types.py
saaste: Sure, I can give it a go 👍

.github/actions/setup-python-env/action.yml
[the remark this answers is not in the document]
nicobav: Yup, it seems that 3.10.11 is the latest of 3.10 👍

api/graphql/opening_hours/opening_hours_types.py
nicobav: There might be better deprecation warnings etc 🤷

reservation_units/tests/test_admin.py
nicobav: Okay, I also did some quick tryouts for it and couldn't figure out main reason but noticed that the field works if put in `setUp` which is instance method.

api/graphql/opening_hours/opening_hours_types.py
saaste: Updated 👍  I also did some refactoring to remove repeating `get_default_timezone()` calls.

reservation_units/tests/test_admin.py
saaste: This was an awesome find! I changed the code so that the object is initialized in `setUp`.

PR #852, opened 2023-05-15
TILA-2546 | Use translated reservation unit and unit names in email context

nicobav:
## Change log
- Translate unit and reservation unit names in `EmailNotificationContext`

## Other notes
Previously the email context didn't translate unit's nor reservation unit's names. Now it does so based on reservations or reservees language choices.

Refs TILA-2546


saaste (approved):
Noice! 👍

PR #859, opened 2023-05-16
TILA-2574: Update price calculation logic

saaste:
## Change log
- Changed shorter pricing unit calculation to be based on 15 minutes intervals
- Changed longer pricing units to be fixed prices

## Other notes
Prices are calculated based on 15 minutes intervals, rounded up. This applies to price units that are less than half day. Other prices units are treated as fixed prices.

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_serializers/mixins.py
+++ b/api/graphql/reservations/reservation_serializers/mixins.py
@@ -87,6 +87,13 @@ class ReservationPriceMixin:
             PriceUnit.PRICE_UNIT_PER_WEEK: 10080,
         }
 
+        fixed_price_units = [
+            PriceUnit.PRICE_UNIT_FIXED,
+            PriceUnit.PRICE_UNIT_PER_HALF_DAY,
+            PriceUnit.PRICE_UNIT_PER_DAY,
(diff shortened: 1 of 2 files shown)


api/graphql/reservations/reservation_serializers/mixins.py
```diff
@@ -122,29 +129,33 @@ def calculate_price(
 
             # Time-based calculation is needed only if price unit is not fixed.
             # Otherwise, we can just use the price defined in the reservation unit
-            if pricing.price_unit != PriceUnit.PRICE_UNIT_FIXED:
+            if pricing.price_unit not in fixed_price_units:
                 reservation_duration_in_minutes = (end - begin).seconds / Decimal("60")
+
+                # Prices are calculated based on the 15 minutes intervals rounded up
+                reservation_duration_in_15mins = math.ceil(
```
nicobav, line 136: kinda funny variable name 😄

nicobav (approved):

api/graphql/reservations/reservation_serializers/mixins.py
saaste: I was really struggling with this one 😅

PR #860, opened 2023-05-16
Upgrade cryptography and cffi

nicobav:
Previous cffi caused build problems with python 3.11 (in ubuntu) so this bumps the cffi and as the cffi comes with dependency of cryptography bumps that as well.


saaste (approved):
Interesting how these things behave differently in different systems 😄

PR #861, opened 2023-05-17
TILA-1518: Simplify local docker running

saaste:
## Change log
- Added management command for creating the initial admin user
- Updated docker-compose setup so that it is easier to run
- Updated README

## Other notes
This should simplify things even more. Now client devs don't need to worry about migration or creating the initial user.

## Deployment reminder
- No changes required

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -5,9 +5,11 @@ COPY ./requirements.txt .
 RUN pip install -r requirements.txt
 
 COPY . /tvp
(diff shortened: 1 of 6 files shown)


nicobav (approved):
Niceness 🍾

nicobav:
Something to do with the container `E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing`

PR #862, opened 2023-05-17
TILA-2616: Remove deprecated timezone handling

saaste:
## Change log
- Removed deprecated timezone handlings
- Removed USE_DEPRECATED_PYTZ settings. Not needed anymore.

## Other notes
We still had places where pytz library was used to handle timezones. This causes problems and confusing behaviour with the new Python version so I fixed all of them.

I also removed USE_DEPRECATED_PYTZ setting because it is not needed anymore.

## Deployment reminder
- No changes required


nicobav (approved):
Time... date... zones

PR #864, opened 2023-05-17
TILA-2501 | Add mutation for modifying reservation details

nicobav:
## Change log
- Adds mutation with serializer and permissions for modifying reservation as staff user

## Other notes
Time adjusting is done via different mutation. 

Refs TILA-2501


saaste (approved):
Gooooood! For some reason Sonar is not happy about the code duplication, but in the report I'm not really seeing where the duplication happens 😄

PR #865, opened 2023-05-17
TILA-2519 | Modify reservation handling permission

nicobav:
## Change log
- Changes permission logic to handling permissions 
- Adds new mutation for refund mutation - it works as was previous mutation

## Other notes
This makes possible for reserver role staff user to perform handling permission based mutations to own reservations.

Within this change refund permission is introduced like the handling was originally working => so refund permissions are not changed but only the Approve and RequiresHandling

Refs TILA-2519


saaste (approved):
Good stuff! 👍

PR #866, opened 2023-05-18
TILA-2614: Fixed order cancellation logic

saaste:
## Change log
- Skip webshop call if order is already in `CANCELLED` state
- Log errors in web shop call, but don't prevent order cancellation/reservation deletion

## Other notes
Web shop API is not called if order is already cancelled. Failures in API call will not prevent reservation cancellation.

## Deployment reminder
- No changes required


nicobav (approved):
LGTM!

PR #870, opened 2023-05-24
Bump requests to latest

nicobav:
Security issues with old versions


requirements_dev.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.11
 # by the following command:
 #
-#    pip-compile --resolver=backtracking requirements_dev.in
+#    pip-compile requirements_dev.in
```
saaste, line 5: I don't know if it makes any difference in the generated file, but should we generate these with `--resolver=backtracking` just to avoid using the [deprecated](https://github.com/jazzband/pip-tools/issues/1659) resolver?

When I did the version upgrade, I had some issues that were solved with `backtracking`.

saaste (approved):
Grreeeeeat! Just comment about not using the deprecated resolver. Feel free to check it out if it makes any difference.

requirements_dev.txt
nicobav: only difference was that line `-c requirements.txt` was missing in comments and obivously the command specified had the `--resolver...` So I would go as is

PR #872, opened 2023-05-24
TILA-2644 | Use correct model in ApplicationEventFilterSet class

nicobav:
## Change log
- Changes `ApplicationEvent` as model for the `ApplicationEventFilterSet` class.

## Other notes
This filterset class was using wrong class (Application) and caused oddities.

Refs TILA-2644


nicobav:
> Funny how this has been like this for a year and no one noticed 🤣 Or was this one of those things where REST API was still used?

Yup, UI just rewrote this to gql

PR #874, opened 2023-05-25
TILA-2628: Ignore blocked reservations when buffers are checked

saaste:
## Change log
- Fixed a bug with `get_next_reservation`
- Ignore `BLOCKED` reservations when buffers are checked
- When BLOCKED reservation is created/updated, buffer check is skipped

## Deployment reminder
- No changes required

### DIFF
--- a/api/graphql/reservations/reservation_serializers/mixins.py
+++ b/api/graphql/reservations/reservation_serializers/mixins.py
@@ -1,7 +1,7 @@
 import datetime
 import math
 from decimal import Decimal
-from typing import Iterable
+from typing import Iterable, Optional
 
 from django.utils import timezone
 from django.utils.timezone import get_default_timezone
@@ -12,6 +12,7 @@ from reservation_units.models import PriceUnit, PricingType, ReservationUnit
 from reservation_units.utils.reservation_unit_pricing_helper import (
     ReservationUnitPricingHelper,
 )
+from reservations.models import ReservationType
 
 
 class PriceCalculationResult:
@@ -305,10 +306,23 @@ class ReservationSchedulingMixin:
                 ValidationErrorCodes.RESERVATION_TIME_DOES_NOT_MATCH_ALLOWED_INTERVAL,
             )
 
-    def check_buffer_times(self, reservation_unit, begin, end):
-        reservation_after = reservation_unit.get_next_reservation(end, self.instance)
+    def check_buffer_times(
+        self,
+        reservation_unit,
+        begin,
+        end,
+        reservation_type: Optional[ReservationType] = None,
+    ):
+
+        current_type = getattr(self.instance, "type", reservation_type)
+        if current_type == ReservationType.BLOCKED:
+            return
+
+        reservation_after = reservation_unit.get_next_reservation(
+            end, self.instance, exclude_blocked=True
+        )
         reservation_before = reservation_unit.get_previous_reservation(
-            begin, self.instance
+            begin, self.instance, exclude_blocked=True
         )
 
         buffer_before = max(
--- a/reservation_units/models.py
+++ b/reservation_units/models.py
@@ -604,8 +604,13 @@ class ReservationUnit(ExportModelOperationsMixin("reservation_unit"), models.Mod
 
         return qs.exists()
 
-    def get_next_reservation(self, end_time: datetime.datetime, reservation=None):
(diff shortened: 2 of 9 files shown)


api/graphql/reservations/reservation_serializers/mixins.py
```diff
@@ -306,9 +306,11 @@ def check_reservation_duration(self, reservation_unit: ReservationUnit, begin, e
             )
 
     def check_buffer_times(self, reservation_unit, begin, end):
```
nicobav, line 308: Got me pondering should we add the `if ResrevationType.BLOCKED` to here to have the iffing only one place 🤔 
On the other hand its more flexible as is now so that the buffers can be checked to blocked ones in some serializers although I don't see the need and that could be arranged if needed. 

So whatya think?

reservation_units/models.py
```diff
@@ -615,10 +620,18 @@ def get_next_reservation(self, end_time: datetime.datetime, reservation=None):
         if reservation:
             qs = qs.exclude(id=reservation.id)
 
-        return qs.order_by("-begin").first()
+        if not include_blocked:
```
nicobav, line 623: I'd change this to `if exclude_blocked` => not clauses are kinda harder to read.

reservation_units/models.py
```diff
@@ -604,8 +604,13 @@ def check_reservation_overlap(self, start_time, end_time, reservation=None):
 
         return qs.exists()
 
-    def get_next_reservation(self, end_time: datetime.datetime, reservation=None):
-        from reservations.models import STATE_CHOICES, Reservation
+    def get_next_reservation(
+        self,
+        end_time: datetime.datetime,
+        reservation=None,
+        include_blocked: bool = True,
```
nicobav, line 611: So as below mentioned this could be `exclude_blocked`

api/graphql/reservations/reservation_serializers/mixins.py
saaste: I added an optional `reservation_type` parameter which can be given in `CREATE` cases where `self.instance` does not exist, and then then I moved the `if` statement inside `check_buffer_times`. Simplified the code quite nicely 👍

reservation_units/models.py
saaste: True! Fixed! 👍

reservation_units/models.py
saaste: Fixed 👍

api/graphql/reservations/reservation_serializers/mixins.py
```diff
@@ -305,10 +306,23 @@ def check_reservation_duration(self, reservation_unit: ReservationUnit, begin, e
...
-    def check_buffer_times(self, reservation_unit, begin, end):
-        reservation_after = reservation_unit.get_next_reservation(end, self.instance)
+    def check_buffer_times(
+        self,
+        reservation_unit,
+        begin,
+        end,
+        reservation_type: Optional[ReservationType] = None,
+    ):
+        if (
```
nicobav, line 316: This could be also like:

`current_type = getattr(self.instance, "type", reservation_type)`
`if current_type == ReservationType.BLOCKED:`

api/graphql/reservations/reservation_serializers/adjust_time_serializers.py
```diff
@@ -61,7 +61,8 @@ def validate(self, data):
             self.check_reservation_time(reservation_unit)
             self.check_reservation_overlap(reservation_unit, begin, end)
             self.check_reservation_duration(reservation_unit, begin, end)
-            self.check_buffer_times(reservation_unit, begin, end)
+            if self.instance.type != ReservationType.BLOCKED:
```
nicobav, line 64: Should we get ridoff this `if` here?

nicobav (changes requested):

api/graphql/reservations/reservation_serializers/mixins.py
saaste: Fixed! 👍

api/graphql/reservations/reservation_serializers/adjust_time_serializers.py
saaste: Fixed! 👍

nicobav (approved):
Oui! 👍

PR #875, opened 2023-05-26
TILA-2568: Fix date filter

saaste:
## Change log
- Fixed `to` filter to include the given date
- Set reservations to be ordered by `begin` field

## Other notes
Reservation filter didn't work properly with `to` filter. Now it includes the given date and uses default timezone when reservations are filtered. Also, default sorting is set to be the `begin` field.

## Deployment reminder
- No changes required


nicobav (approved):
🤌

PR #876, opened 2023-05-26
TILA-2617: Anonymization does not change empty fields

saaste:
## Change log
- Updated string anonymization to leave empty fields empty

## Deployment reminder
- No changes required


nicobav (approved):
🤌

PR #877, opened 2023-05-29
TILA-2662: Set authenticated user in create and update

saaste:
## Change log
- Set user in `create` and `update` methods

## Other notes
In the REST API, creating and updating an application did not save the user. This is probably because of the updated `djangorestframework`. The simple fix was to just set it manually in create and update methods.

Would be really nice to get rid of these REST endpoints 😄 

## Deployment reminder
- No changes required

### DIFF
--- a/api/applications_api/serializers.py
+++ b/api/applications_api/serializers.py
@@ -665,6 +665,8 @@ class ApplicationSerializer(serializers.ModelSerializer):
             request.user if request and request.user.is_authenticated else None
         )
 
+        validated_data["user"] = request_user
+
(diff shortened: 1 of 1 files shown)


nicobav (approved):
:shipit:

saaste:
I'm going to postpone merge. While investigating this I found out even more problems with these REST endpoints. Also, the existing behaviour seems to be wrong so we need to figure out if we focus on fixing the REST endpoint or create a working mutation to GraphQL and get rid of the REST stuff.

PR #878, opened 2023-05-30
TILA-2618: Remove roles when user is anonymized

saaste:
## Change log
- Disable special roles and notifications when user is anonymized
- Disable general, service sector and unit roles when user is anonymized

## Deployment reminder
- No changes required


nicobav (approved):
🚂 choochoo last call to the `main` branch 🚂

PR #880, opened 2023-05-31
TILA-2620: Fix reservation notification permissions

saaste:
## Change log
- Fixed the logic with `get_reservation_notification`
- Updated tests

## Other notes
We need to use `has_staff_permissions` instead of `is_staff`. Thise caused some changes to test data as well. It is actually a bit confusing that we have the `is_staff` flag but we an alternative definition for `staff_permissions`.

## Deployment reminder
- No changes required


saaste:
It is that flaky application export test that is failing. I already tried to run the tests twice with no success.

nicobav (approved):

PR #881, opened 2023-06-01
TILA-2596: Fix reservation statistics update + migration script

saaste:
## Change log
- Fixed reservation statistics update
- Added migration script for updating existing statistics without reservation unit
- Remove invalid reference from fixtures

## Other notes
The root cause for the bug was the fact that Django `save` does not have many-to-many references saved at that time. That is why, when reservation is created, reservation unit ends up being NULL. Likewise, when reservation is updated, statistics object ends up having the previous reservation unit instead the updated on.

The solution is to use Django signals. `post_save` signal is needed when reservation is updated without reservation unit changes. That way every other field gets updated to statistics object. `m2m_changed` signal is used when reservation units are changed.

I added a lengthy comment in the code explaining to behaviour and also the reasoning.

## Deployment reminder
- No changes required


nicobav (approved):
Awesome 💪

PR #882, opened 2023-06-02
TILA-2410 - Small tweaks

saaste:
## Change log
- Added `celery.pidbox.exchange` to `.gitignore`
- Added `SESSION_COOKIE_DOMAIN`

# Notes
This PR used to be about file upload, but yesterday we agreed that we just keep the CSRF-token disabled, because there is no easy way to make it work and with the limited time we have, we can't focus on figuring out how to make it work on both frontend and backend.

So, now it is just about small tweaks.

## Deployment reminder
- No changes required

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -62,6 +62,9 @@ CONN_MAX_AGE=0
 # Django setting: prepended to CSRF_COOKIE_NAME and SESSION_COOKIE_NAME
 COOKIE_PREFIX=tilavarauspalvelu
 
+# Session cookie domain if Django's default behavior needs to be overridden
+# SESSION_COOKIE_DOMAIN=local-tilavaraus.hel.fi
+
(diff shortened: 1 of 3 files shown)


nicobav (approved):
Oh yes please! 🙏

api/graphql/reservation_units/reservation_unit_serializers.py
```diff
@@ -589,7 +589,7 @@ def validate_image_type(self, type):
         return type.lower()
 
     def validate(self, data):
-        image = self.context.get("request").FILES.get("image")
+        image = self.context.get("request").FILES.get("0")
```
nicobav, line 592: I bet this broke the test.

nicobav (approved):
This PR was a roller coaster of feelings 🥲

PR #884, opened 2023-06-06
TILA-2458 | Elasticsearch TextSearch

nicobav:
## Change log
- Adds elasticsearch image to ci pipeline and local dev (to docker compose)
- Adds ReservationUnits to elastic index

## Other notes
Testing instructions are in README.md => how to not messup your local dev index with test data.

## Deployment reminder
- [x] Conifgure elastic url and secrets: https://dev.azure.com/City-of-Helsinki/tilavarauspalvelu/_git/tilavarauspalvelu-pipelines/pullrequest/5102


reservation_units/models.py
```diff
@@ -685,6 +698,74 @@ def save(self, *args, **kwargs) -> None:
...
+                "purposes_name_fi": ",".join(
+                    [p.name_fi or "" for p in self.purposes.all()]
+                ),
+                "purposes_name_en": ",".join(
+                    [p.name_en or "" for p in self.purposes.all()]
+                ),
+                "purposes_name_sv": ",".join(
+                    [p.name_sv or "" for p in self.purposes.all()]
+                ),
+                "reservation_unit_type_name_fi": self.reservation_unit_type.name_fi,
```
saaste, line 751: I tried to run the migration on my machine and it failed to this line. `reservation_unit_type` can be `None` so I believe some null checks should be in place here.

elastic_django/tests/test_reservation_unit_query_builder.py
```diff
@@ -0,0 +1,54 @@
...
+        query_str = ReservationUnitQueryBuilderMixin.build_elastic_query_str("search")
+
+        assert_that(query_str).is_equal_to("(*search*)")
+
+    def test_parse_search_word_when_two_search_word(self):
+        query_str = ReservationUnitQueryBuilderMixin.build_elastic_query_str(
+            "search this"
+        )
+
+        assert_that(query_str).is_equal_to("((*search) AND (this*))")
```
saaste, line 20: Does this also match to text when there are words in between the two key words? For example, does `((*search) AND (this*))` match to `search foobar this`? If yes, is this the intended behaviour or should it be something like `(*search AND this*)`?

api/graphql/reservation_units/reservation_unit_filtersets.py
```diff
@@ -145,45 +151,13 @@ class Meta:
...
-                | Q(reservation_unit_type__name_sv__icontains=word)
-                | Q(equipments__name_fi__icontains=word)
-                | Q(equipments__name_en__icontains=word)
-                | Q(equipments__name_sv__icontains=word)
-                | Q(unit__name_fi__icontains=word)
-                | Q(unit__name_en__icontains=word)
-                | Q(unit__name_sv__icontains=word)
-            )
-        query = reduce(operator.or_, (query for query in queries))
+        sq = SearchQuery.do_search(
```
saaste, line 156: 👏 👏 👏

saaste (changes requested):

reservation_units/models.py
nicobav: Excellent catch. I fixed this and also possible unitless reservation unit as well with `getattr`

saaste (approved):
Awesome job! I tried this on my machine and it works SO well! I didn't do any performance tests yet but just querying stuff manually gave a feeling this is so much faster even with smaller databases and simpler queries. 🦜 🥇

PR #887, opened 2023-06-09
TILA-2679 | Deny user data delete through gdpr api when there's reservations in one month backwards

nicobav:
## Change log
- Deny deleting user data through gdpr api when user has reservations one month backwards
- Add flake local_settings to flake ignore.
- Add reservation begin and end to gdpr output data
- Add tests for reservation and application data
- Change applications to be returned list inside a list rather than single list with all the application data.

Refs TILA-2679


saaste (approved):
Veeeery noice! 👏

PR #888, opened 2023-06-13
TILA-2673 reservation permission changes

nicobav:
## Change log
- Adds permission logic for staff users to see working memos, staff_event, type and handled_at in `Reservations` query
- Reorders the `ReservationType` fields and methods.
- Adds field `is_blocked` to `ReservationType` to indicate whether the reservation is blocked type or not. Public field.

## Other notes
Strongly suggest to review by commits since the refactoring confuses the changeset a bit here.

Refs TILA-2673 TILA-2674


saaste (approved):
Nice work! I like the decorator cleanup and big 👏  for sorting fields and resolve-functions 😄

PR #892, opened 2023-06-16
TILA-2676: Add support for image cache and purging of individual images

saaste:
## Change log
- Added a helper function for purging individual images
- Updated Purpose save so that it'll purge images every time Purpose is saved
- Unit tests for new functionality

## Other notes
Images are purged every time Purpose is saved. I didn't bother checking if new image is actually uploaded since this is just about the cache. All it causes is an addition backend hit before image is cached again. I decided to do this in as a background task because it is using HTTP calls that can delay the `save` call. Because cache purge is a secondary action, in case of error it does not prevent Purpose saving. However, errors are logged.

Vitalijus already set up an image cache for us in the dev environment. We just need to enabled it before this one is merged.

## Deployment reminder

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -283,3 +283,12 @@ VERKKOKAUPPA_ORDER_EXPIRATION_MINUTES=5
 
 # Redis cache URL
 REDIS_URL=redis://127.0.0.1:6379/0
+
+# Is image cache enalbed
+IMAGE_CACHE_ENABLED=False
+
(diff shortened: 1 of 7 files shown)


nicobav (approved):
🚀 GR8!

saaste:
Failing tests are not related to this PR and because you are already fixing those flaky tests in your PR, I'll merge this PR so that I can continue with the cache stuff.

PR #893, opened 2023-06-19
TILA-2678 | Application export field adds and admin functionality

nicobav:
## Change log
- Adds new fields to application export
- Adds application export to admin
- Fix some flickery flicker tests.

Refs TILA-2678


applications/admin.py
```diff
@@ -93,13 +95,38 @@ class Meta:
...
+    @admin.action(description="Export first selected reservation unit to CSV")
+    def export_to_csv(self, request, queryset):
+        try:
+            app_round = queryset.first()
+            path = ApplicationDataExporter.export_application_data(
+                application_round=app_round
+            )
+
+        except Exception as e:
+            self.message_user(
```
saaste, line 115: Should we log these to sentry so that we are not depending on user's screenshot when debuging possible errors? Simple `capture_exception(e)` should do the trick.

saaste (approved):
Looks great! One suggestion for easier debugging.

applications/admin.py
nicobav: very good idea, i'll do that

saaste (commented):
Great stuff! :shipit:

PR #896, opened 2023-06-19
TILA-2685 | Make incoming verkkokauppa timestamps tz aware

nicobav:
## Change log
- Make incoming verkkokauppa timestamps tz aware
- Adds also a setting to be used to determine the verkkokauppa's timezone since the strings come without tz info

## Deployment reminder
- [x] PR requires deployment changes
- [x] Azure devops library updated if needed

Refs TILA-2685


saaste (approved):
Thanks for fixing this! Such a nice improvement and bonus points for making this as a setting 👏

PR #897, opened 2023-06-19
TILA-2565 | Remove District model

nicobav:
## Change log
- Remove district model
- Fix reservation_unit surface are ordering tests

## Other notes
The district will not be used in 2.5.

Refs TILA-2565


saaste (approved):
These red PRs are awesome 👨‍🍳 🤌

joonatank (approved):
There is District type definitions in the frontend that are not used anywhere.
=> should be removed there also, but not a problem merging this.

PR #900, opened 2023-06-21
TILA-2676: Purge image cache on Reservation Unit Image save

saaste:
## Change log
- Added image purge call to ReservationUnitImage save
- Added unit tests

## Other notes
There is no need to purge images when ReservationImageUnit item is deleted because these images are not referenced anywhere anymore. Client should not receive these image urls from the API. Yes, cached images are still visible but cache has 24h TTL so these images will disappears automatically.

## Deployment reminder
- No changes required


nicobav (approved):
Awesome 😎

PR #902, opened 2023-06-21
TILA-1870 | Validate reservation unit terms in django admin

nicobav:
## Change log
- Adds form validation for *_terms in `ReservationUnitAdminForm`

Refs TILA-1870


saaste (approved):
🚀

PR #903, opened 2023-06-22
TILA-1059 | Add fields to reservations from applications

nicobav:
## Change log
- Increase `Reservation.reservee_id` max_length to 255 as is in `Organisation` model
- Fill bunch of fields from application data to reservations when creating from schedule results.

Refs TILA-1059


saaste (approved):
🚀

vergama (approved):

PR #904, opened 2023-06-27
TILA-2670: Fix buffer validation in staff reservation mutations

saaste:
## Change log
- Updated mutations so that updated buffers are used in validation
- Fixed a bug in validation logic

## Deployment reminder
- No changes required


nicobav (approved):
👍

PR #907, opened 2023-06-30
TILA-2661 | Take into account collapsing closed opening hours

nicobav:
Takes into an account collapsing opening hours e.g when the reservation unit is open 10-18 and there is maintenance break 13-14 the opening hour needs to be splitted into two (10-13 and 14-18)

Refs TILA-2661

### DIFF
--- a/opening_hours/enums.py
+++ b/opening_hours/enums.py
@@ -1,3 +1,5 @@
+from types import DynamicClassAttribute
+
 from django.utils.translation import pgettext_lazy
 from enumfields import Enum
 
@@ -47,6 +49,14 @@ class State(Enum):
     @classmethod
     def closed_states(cls):
         return [None, cls.CLOSED, cls.UNDEFINED, cls.NOT_IN_USE, cls.MAINTENANCE]
+
+    @DynamicClassAttribute
+    def is_open(self) -> bool:
+        return self in State.open_states()
+
+    @DynamicClassAttribute
+    def is_closed(self) -> bool:
+        return self in State.closed_states()
 
     @classmethod
     def reservable_states(cls):
--- a/opening_hours/enums.py
+++ b/opening_hours/enums.py
@@ -36,12 +36,17 @@ class State(Enum):
     def open_states(cls):
         return [
             cls.OPEN,
+            cls.OPEN_AND_RESERVABLE,
             cls.SELF_SERVICE,
             cls.WITH_KEY,
             cls.WITH_RESERVATION,
             cls.WITH_KEY_AND_RESERVATION,
             cls.ENTER_ONLY,
         ]
+
+    @classmethod
+    def closed_states(cls):
+        return [None, cls.CLOSED, cls.UNDEFINED, cls.NOT_IN_USE, cls.MAINTENANCE]
 
     @classmethod
     def reservable_states(cls):
--- a/opening_hours/enums.py
+++ b/opening_hours/enums.py
@@ -35,36 +35,57 @@ class State(Enum):
         MAINTENANCE = pgettext_lazy("State", "Maintenance")
 
     @classmethod
-    def open_states(cls):
+    def accessible_states(cls):
+        """
+        States indicating the space can be accessed in some way,
+        whether the access is restricted (e.g. via key or reservation)
+        or not.
+        """
         return [
+            cls.ENTER_ONLY,
             cls.OPEN,
             cls.OPEN_AND_RESERVABLE,
             cls.SELF_SERVICE,
             cls.WITH_KEY,
-            cls.WITH_RESERVATION,
             cls.WITH_KEY_AND_RESERVATION,
-            cls.ENTER_ONLY,
+            cls.WITH_RESERVATION,
+        ]
+
+    @classmethod
+    def reservable_states(cls):
+        """
+        States indicating the space can be reserved in some way.
+        """
+        return [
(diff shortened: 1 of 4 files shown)


nicobav:
For the 2.5 team to take over. More tests needed.

vergama (changes requested):
I think there may be some problems in this:

**1.** A closed hour-period could theoretically occur inside more than one open and/or reservable period.
While the times should not be put into the Opening Hours Application like that (for Varaamo resources anyway), but for example, if it had three overlapping and conflicting periods like this defined:
* _closed_ 11-12
* _reservable_ 8-13
* _open_ 10-18
... then this code will now use the _closed_ period to completely split only the _open_ 10-18 period, because the inner loop creates a `new_open` addition only for the last item in `open_hours` encountered which encompasses the _closed_ item.
It will correctly also shorten the _reservable_ period to end when the _closed_ starts, but the issue is that the "left-over" hour (12-13) after _closed_ ends will be omitted (and the _open_ part won't be a substitute since it won't allow making reservations).
I.e., the result would be (as the hours array is returned):
* _closed_ 11-12
* _reservable_ 8-11 (end time changed)
* _open_ 10-11 (ditto)
* _open_ 12-18 (new_open appended)
... but ideally we'd like to have (re-ordered for clarity here):
* reservable 8-11
* open 10-11
* closed 11-12
* reservable 12-13
* open 12-18
So I guess `new_open` should be an array too, and then the code will be more tolerant when data of that sort has been defined in **Aukiolosovellus**.

**2.** A closed period can encompass an open/reserable period, which is now transformed to zero or negative length (`start_time > end_time`), and these are also left in the returned array. Not sure if zero-length ones would actually cause any problems in the UI (negative ones more likely) but might be better to purge them in any case.
For example if we have this input:
* _closed_ 9-12
* _reserable_ 9-10 (overriden entirely by the special closed)
* _open_ 10-17 (no conflict with reservable, closed overlaps with the start)
... now results in:
* _closed_ 9-12
* _reservable_ 12-9 (set to reverse of closed)
* _open_ 12-17 (no problem here)
... but we rather want just this with the reservable period nixed:
* _closed_ 9-12
* _open_ 12-17
Solvable by having the `new_open` array inside the _closed_ loop include both the split open hours and those we which to keep in shortened form (the transformed result still had positive length), and then swap that for `open_hours` before the next item in `closed_hours` is handled? And finally sort what will be the combined returned result (closed_hours + the transformed open_hours) so that it's easier to parse from the API for humans (even if the UI doesn't care about the order).

**3.** There is also the fact that hours which are considered neither `closed` nor `open` by their status (is that possible?) will be ignored and not passed through?

matti-lamppu:
@vergama What should be done about point number 3? There are currently 2 opening hour states that do not fall under reservable or closed status: `EXIT_ONLY`, and `WEATHER_PERMITTING`.

vergama:
> @vergama What should be done about point number 3? There are currently 2 opening hour states that do not fall under reservable or closed status: `EXIT_ONLY`, and `WEATHER_PERMITTING`.

Let's just pass them through as they are. Such statuses are not not supposed to be used for reservable-resources of Tilavarauspalvelu in Aukiolosovellus, but should they later come into play, at least there's some reasonable logic in place.

vergama:
Also the effects on `end_time_on_next_day` by splitting or shortening a time entry are now not yet considered; it can change from true to false due to that.

vergama (commented):
One nitpick: it would be more accurate to name `tracked_reservable_hours` as `tracked_open_hours` since it can include also non-reservable hours where the space is nonetheless "open"; e.g. `OPEN` = the library doors are open but the space is in public use; `WITH_KEY` = space can be accessed with a key/pin code (only) but not nevertheless cannot be reserved... don't actually know what scenario that latter would be but the state exists in the Opening Hours Application (somebody has seen fit to define it there). (`WITH_KEY_AND_RESERVATION`, on the other hand, would be the state used in the future for reservable spaces where a pin code is needed to access during those hours.)

Then it also goes hand in hand with `open_states` and `is_open()` -- strictly speaking "open_or_reservable" would be even more accurate but for the sake of simplicity we could treat those as "open" in the widest sense: can be accessed, possibly with some restrictions -- or come to think of it, maybe `tracked_accessible_hours` &c., then?

Background: While the frontend is (now) only interested in the hours that are truly reservable, and the merely generally open (or with key) hours are not supposed to be / need not be entered into the Opening Hours Application for the resources representing the reservation units (that info would more properly be at the parent/unit level, such as the opening hours of the library which contains a reservable room), that might change in the future, so we handle the strictly open hours here too. If there is such additional data, it can for example be shown in the admin-ui if some kind of overview of the hours of the reservation units of a unit is implemented.

**More importantly**: @ line 430: `if ResourceState(time.resource_state).is_closed`:
`!is_closed` also includes all the open/accessible states which are not necessarily reservable; only the 3 `reservable_states` are.
Perhaps a better name would be `is_resource_open_for_reservations()` => `is_resource_reservable() `to indicate that.

matti-lamppu:
I think "accessible" is a better name for "open" opening hours, so I'll make that refactor. Thanks for the background on that one!

I made the refactor for `is_resource_open_for_reservations` for performance reasons and didn't understand that `open != reservable` at that point. Will fix that!

(BTW, in the future, if your comment is about a specific line, you can make a comment on that line specifically. Makes it easier to see which line you are talking about 😊)

vergama (approved):

PR #909, opened 2023-08-09
Add local test data

matti-lamppu:
### Test data generation script

Adds a Django management script to generate test data.

- `Users` & `Permissions` for them based on [these rules](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/8051753147/Henkil+kunnan+k+ytt+j+roolit#Konventio-testihetuilla)
    - 010101-900V: Unit reserver (All)
    - 020202-900V: Unit viewer (All)
    - 030303-900V: Unit handler (All)
    - 040404-900V: Unit admin (All)
    - 010111-900V: Service Sector reserver (All)
    - 020222-900V: Service Sector viewer (All)
    - 030333-900V: Service Sector handler (All)
    - 040444-900V: Service Sector admin (All)
    - 090909-900D: Overall Admin
- 10 `Units` & `Locations` for each
    - 3 Are linked to Hauki entities
- 3 `Service Sectors`
- 3 `Spaces`
- 3 `Reservation Unit Types`
- `Terms of Use` for each term type
- 1 `Reservation Cancellation Rule`
- 3 `Reservation Metadata Sets`
- 10 `Equipments` & 1 `Equipment Category`
- 1 `Qualifier`
- 10 `Purposes`
- 10 `Resources`
- 10 `Services`
- 300 `Reservation Units` & `Reservation Unit Pricings` for each
    - Some have "free", some "paid", and some both
- 10 `Reservation Purposes`
- 7 `Age Groups`
- 10 `Reservation Cancel Reasons`
- 10 `Reservation Deny Reasons`
- 200 `Reservations` for user "090909-900D"
    - Divided (about) equally into past, upcoming, and canceled
 - 15 `Application Rounds`
     - 1/3 past, 1/3 present, and 1/3 future ones 
 - 20 `Applications` & `Location`, `Address`, and `Organisation` for each
 - 1-3 `Applitation Events` for each `Application`
 - 1-4 `Event Reservation Units` for each `Applitation Event`
 - 1-4 `Application Event Schedules` for each `Applitation Event`

### Misc

- Adds Makefile for running common commands
- Adds tests for the generation script
- Updates cryptography due to vulnerability
- Removes the old data fixture & tests

## Deployment reminder

---

Jira Issue: [TILA-2700](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2700)

[TILA-2700]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matti-lamppu:
Reconsidering making this into a Django management command for maintainability, since the JSON fixture is so big.

matti-lamppu:
Added Django management script for generating test data. This will replace the JSON fixture. Some seasonal booking data is still needed.

matti-lamppu:
Seasonal booking data added. Merchant data is still lacking, although I couldn't get it working locally anyway. Maybe something to add in the future, but would like to move on to other tasks for a while at least 😄

matti-lamppu:
Sorry for the huge PR, there is just a lot of stuff 😅

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+        tax_percentage: TaxPercentage = random.choice(tax_percentages)
+        multiplier = Decimal(f"1.{int(tax_percentage.value)}")
+
+        lowest_net = random.randint(0, 100)
+        highest_net = random.randint(lowest_net, 1000)
+        lowest = float(lowest_net * multiplier)
+        highest = float(highest_net * multiplier)
+
+        # Around 1/10 of the pricings are inactive
+        pricing_status: str = random.choices(PricingStatus.values, weights=[1, 10, 0])
```
vergama, line 1294: **Pricings**: does the randomization sometimes produce two active pricings for the same reservation unit? In any case with overlapping time-period = same begins.

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+                "num_persons",
+                "purpose",
+            ],
+            required=[
+                "purpose",
+            ],
+        ),
+    ]
+
+    field_combinations_loop = cycle(field_combinations)
```
vergama, line 853: **Metadata**: are the sets sufficient or should they be more like in the dev/test/stage/prod environments (the specific "form" combinations)? @joonatank

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+        end = begin + timedelta(hours=random.choice(range(min_hours, max_hours)))
+
+        state = STATE_CHOICES.CONFIRMED
+
+        pricing: ReservationUnitPricing = random.choice(
+            list(reservation_unit.pricings.all())
+        )
+        if pricing.highest_price != Decimal("0"):
+            state = random.choice(
+                [STATE_CHOICES.WAITING_FOR_PAYMENT, STATE_CHOICES.CONFIRMED]
```
vergama, line 1379: **Reservations**: might not need `state = WAITING_FOR_PAYMENT` ones, they would be deleted by batch job (and can't be paid before that since don't exist in Verkkokauppa).
But some `REQUIRES_HANDLING` could be useful? (Could be those with the non-zero price + `applying_for_free_of_charge` set to true and a `free_of_charge_reason` on those.)

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+            reservee_address_street=faker_fi.street_name(),
+            reservee_address_zip=faker_fi.postcode(),
+            reservee_email=faker_fi.email(),
+            reservee_first_name=user.first_name,
+            reservee_id=faker_fi.company_business_id(),
+            reservee_language="fi",
+            reservee_last_name=user.last_name,
+            reservee_organisation_name=faker_fi.company(),
+            reservee_phone=faker_fi.phone_number(),
+            reservee_type=CUSTOMER_TYPES.CUSTOMER_TYPE_INDIVIDUAL,
```
vergama, line 1429: `reservee_type`:

- CUSTOMER_TYPE_INDIVIDUAL: no reservee/business_id or reservee_organisation_name
- CUSTOMER_TYPE_BUSINESS: has those
- CUSTOMER_TYPE_NONPROFIT: has both unless `reservee_is_unregistered_association` is true in which case only the org.name
(The latter two are OK to have the same person details are INDIVIDUAL; it is the contact person info then.)

common/management/commands/create_test_data.py
vergama: No confirmed_at then if REQUIRES_HANDLING -- or DENIED (which would have handled_at).

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+            deny_reason=deny_reason,
+            description=faker_fi.sentence(),
+            end=end,
+            free_of_charge_reason=faker_fi.sentence(),
+            name=f"Reservation {i}",
+            num_persons=persons,
+            price=pricing.highest_price,
+            price_net=pricing.highest_price_net,
+            purpose=random.choice(reservation_purposes),
+            reservee_address_city=faker_fi.city(),
```
vergama, line 1419: + `home_city` ("kotikunta")

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+        )[0]
+
+        application = Application(
+            applicant_type=applicant_type,
+            contact_person=contact_person,
+            user=user,
+            organisation=organisation,
+            application_round=random.choice(application_rounds),
+            billing_address=billing_address,
+            additional_information=faker_fi.sentence(),
```
vergama, line 1550: + `home_city`

common/management/commands/create_test_data.py
matti-lamppu: Yes, there might be two active pricings. Is this an issue?

common/management/commands/create_test_data.py
matti-lamppu: I showed the configured field combinations to @joonatank previously, though if there are some more "common" combinations, it would benefit us to add those here.

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+        user: User = random.choices(
+            users, weights=[len(users)] + ([1] * (len(users) - 1))
+        )[0]
+
+        application = Application(
+            applicant_type=applicant_type,
+            contact_person=contact_person,
+            user=user,
+            organisation=organisation,
+            application_round=random.choice(application_rounds),
```
vergama, line 1548: Statuses for:
- **_ApplicationRound_**: **past** = `HANDLED`, **current** = `DRAFT` [_sic_], **future** = `DRAFT`
- **_Application_**: **past** = `HANDLED` (but decisions not sent) or `CANCELLED` or `EXPIRED`, **current** = `DRAFT` or `RECEIVED` or `CANCELLED`, **future** = _no applications for this kind of round_
- **_ApplicationEvent_**: **past** = `APPROVED` (would be `RESERVED` and the parent statuses `SENT` if process inside the system extended that far) or `DECLINED`, **current** = `CREATED` (i.e., not yet decided/handled)

common/management/commands/create_test_data.py
```diff
@@ -0,0 +1,1775 @@
...
+            event_units.append(event_unit)
+
+    return EventReservationUnit.objects.bulk_create(event_units)
+
+
+@with_logs(
+    text_entering="Creating application event schedules...",
+    text_exiting="Application event schedules created!",
+)
+def _create_application_event_schedules(
```
vergama, line 1755: Allocated schedules only for `APPROVED` application events; `CREATED` and `DECLINED` have none.

vergama (changes requested):
**Merchants** data can wait for now, its model needs to be changed slightly anyway because Verkkokauppa made a breaking change into theirs.
Other comments on **create_test_data.py**.

common/management/commands/create_test_data.py
vergama: The combinations can be seen in Django admin of all environments (dev has some others too); they are basically:
- **Lomake 1**: name and email + phone is asked (mandatory), use on reservation units for individual users only
- **Lomake 2**: adds description (req) and number of persons (opt), plus the fields for other customer types (all required but reservee_type + is_unreg... will be taken into account)
- **Lomake 3**: adds home_city and (reservation) purpose (both req.) and reservation name (opt, the shorter description); these two forms used on reservation units which require more information, possibly in order to make a decision whether to confirm or deny a reservation which requires staff handling
- **Lomake 4**: adds age_group to the above (req)
- the variants of 3 & 4 ("alennus sallittu") add applying_for_free_of_charge; for use on priced reservation units where subvention of the price can be applied for

In theory the UI would also support other sets too, so maybe the current code is desirable, but at least FYI those are the combos that in practice exist and are dealt with.
Implemented that way for historical reasons, maybe would now be done some other way in light of current information.
Actual address fields ended up not being asked on any form currently. But `home_city` is separate from those (in practice the choices are HELSINKI vs OTHER, just to make the distinction between those living there and everyone else).

common/management/commands/create_test_data.py
vergama: Yeah there should only be one = the current on one particular res.unit (begins <= now). The rest are its past or future.

matti-lamppu: Ok, thought that maybe there could be one active free and paid pricing. I'll simplify this bit then.

common/management/commands/create_test_data.py
matti-lamppu: Done.

common/management/commands/create_test_data.py
matti-lamppu: Appropriate status entities are create with the above rules.

common/management/commands/create_test_data.py
matti-lamppu: Done.

common/management/commands/create_test_data.py
matti-lamppu: Reservee type and the related fields are now determined according to the above rules.

common/management/commands/create_test_data.py
matti-lamppu: Changed implemented.

common/management/commands/create_test_data.py
matti-lamppu: Added these combinations to the created sets. Also made sure that the variant 3 and 4 are only use id said field is true.

common/management/commands/create_test_data.py
matti-lamppu: Done.

vergama (approved):

matti-lamppu:
Waiting fo @vincit-matu or @joonatank to test and comment on if this works for them before I merge it.

matti-lamppu:
Approval given in Slack by @joonatank to merge this.

PR #931, opened 2023-09-20
Lint tests

ranta:
## 🛠️ Changelog
- Enable linting of tests
- Fix linter issues in test files

## 🎫 Tickets
- TILA-2836

### DIFF
--- a/api/graphql/tests/test_recurring_reservations/test_recurring_reservations_queries.py
+++ b/api/graphql/tests/test_recurring_reservations/test_recurring_reservations_queries.py
@@ -190,22 +190,21 @@ class ReservationQueryTestCase(ReservationTestCaseBase):
         ]
         for filter_name, filter_value, field_name in test_cases:
             response = self.query(
-                """
-                query {
-                    recurringReservations(%s: "%s") {
+                f"""
+                query {{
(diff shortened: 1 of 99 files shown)


api/graphql/tests/test_recurring_reservations/test_recurring_reservations_queries.py
```diff
@@ -190,22 +190,21 @@ def test_filter_by_reservation_unit_name(self):
         ]
         for filter_name, filter_value, field_name in test_cases:
             response = self.query(
-                """
-                query {
-                    recurringReservations(%s: "%s") {
+                f"""
+                query {{
+                    recurringReservations({filter_name}: "{filter_value}") {{
```
matti-lamppu, line 195: Having double curly braces like this is a bit ugly, so I prefer the format strings in this case. Although a better solution would be to use variables like [here](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/a6a7d4947520dc47d4074cf9d5faa0683bf48ffe/tests/test_graphql_api/test_banner_notification/test_filtering.py#L235). But we can leave that for later.

matti-lamppu (approved):

PR #940, opened 2023-10-20
Fix erros when cancelling helsinki authentication

matti-lamppu:
## 🛠️ Changelog
- Fixes errors when returning from Helsinki authentication by setting social auth middleware and appropriate error return URL.

## 🗒️ Other notes
Additional notes if needed

## 🚚 Deployment reminder

## 🎫 Tickets
- ...


matti-lamppu:
Note that the error return URL is by default to the backend admin panel. This should be configured with the env variable. There might be a way to set this from frontend via some query parameter, but didn't look into it yet. Still, this can be used as a quick fix.

vergama (approved):
Looks ok as a quick fix.

PR #958, opened 2023-11-03
Migrate to python 3.12

matti-lamppu:
## 🛠️ Changelog
- Migrates the project to Python 3.12

## 🧪 Test plan

- Create a new virtual environment locally for python 3.12:
    - `poetry env use "<path to python 3.12 executable"` -> `poetry install`
- Run tests with new environment
- Run Django admin and graphQL with new environment

CI already runs tests and builds image, so those should be tested with that


matti-lamppu:
@ranta If we want to migrate to 3.12, we need to remove all snapshot tests, since the library requires the `imp`-module, [which was removed in 3.12](https://docs.python.org/3/whatsnew/3.12.html#:~:text=The%20asynchat%2C%20asyncore%2C%20and%20imp%20modules%20have%20been%20removed%2C%20along%20with%20several%20unittest.TestCase%20method%20aliases.).

[There is an open issue about this](https://github.com/syrusakbary/snapshottest/issues/166), but little activity.

ranta:
I don't think we are in any hurry to upgrade to 3.12. We can start slowly refactoring away the snapshot tests along with all of the other refactoring stuffs and upgrade python at some later point :)

ranta (approved):

PR #974, opened 2023-11-13
Celery redis config

matti-lamppu:
## 🛠️ Changelog
- Configure Redis broker for celery

## 🎫 Tickets
- TILA-2938

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 .PHONY: bash
+.PHONY: celery
 .PHONY: clear-db
 .PHONY: db-reset
 .PHONY: dev
@@ -32,6 +33,7 @@ define helptext
  Commands:
 
   bash                               Open bash in backend container.
+  celery                             Run a local celery worker.
   clear-db                           Reset database.
   dev                                Run local backend.
   flush-db                           Flush database.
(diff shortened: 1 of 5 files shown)


matti-lamppu:
Tested this in dev like this: 
1. Open worker pod and core pod terminals side by side
2. In worker pod: `cat /broker/worker.log` -> note the last entry.
3. In core pod: `python manage.py shell_plus` -> `from reservations.tasks import update_expired_orders_task` -> `update_expired_orders_task.delay()`. This runs this celery job using the broker.
4. In worker pod: `cat /broker/worker.log` again -> note the last entry. It should be for the task you just ran in the core pod, and something similar to below:

```shell
[2023-11-13 16:05:00,047: INFO/MainProcess] Task prune_reservations[6564bcf2-3da8-4dc7-8723-805251e9ca3a] received
[2023-11-13 16:05:00,048: INFO/ForkPoolWorker-14] Pruning inactive reservations older than 20 minutes...
```

This should prove that celery task initiated in the core pod are now run in the worker pod celery worker. Celery worker in the core pods has been removed.

matti-lamppu:
Does not require any new variables, since the redis variables are used to configure celery as well. Azure library needs a new config: `CELERY_FILESYSTEM_BACKEND` set to `'false'` (note the quotes!). This is already transferred to the container, but the default value is `'true'`.

vergama (approved):

PR #990, opened 2023-11-20
Overhaul opening hours app

ranta:
## 🛠️ Changelog

- Cache Hauki Resources and their ReservableTimeSpans in our own db
- Convert everything doing Hauki API calls to use our db instead
- Add a task to allow periodically fetching new time spans for all OriginHaukiResources

## 🧪 Test plan

- Automated tests
- Try to import some reservable time spans

## 🎫 Tickets

- TILA-2820


api/graphql/tests/test_reservation_units/conftest.py
```diff
@@ -9,3 +11,8 @@ def _setup_hauki_env_variables(settings):
     settings.HAUKI_SECRET = "HAUKISECRET"  # noqa: S105
     settings.HAUKI_ORGANISATION_ID = None
     settings.HAUKI_ADMIN_UI_URL = "https://test.com"
+
+
+@pytest.fixture()
+def graphql() -> GraphQLClient:
```
matti-lamppu, line 17: Should move any tests that require this fixture to the `/tests/` folder.

api/graphql/tests/test_reservation_units/test_reservation_units_query_opening_hours.py
```diff
@@ -0,0 +1,274 @@
...
+        latest_fetched_date="2021-05-03",
+    )
+    return ReservationUnitFactory(origin_hauki_resource=origin_hauki_resource)
+
+
+def _get_date(*, day: int = 1, hour: int = 0, tzinfo: ZoneInfo | None = None):
+    return datetime.datetime(2023, 5, day, hour, 0, 0, tzinfo=tzinfo or DEFAULT_TIMEZONE)
+
+
+GET_QUERY = """
```
matti-lamppu, line 38: How about using `tests.helpers.build_query` here? It would need some modification to enable filtering sub-selections, e.g. with `reservable_time_spans__start_date="2023-05-01"`.

api/graphql/types/opening_hours/types.py
```diff
@@ -1,187 +1,58 @@
...
 
 
-class OpeningTimesType(graphene.ObjectType):
-    date = graphene.Date()
-    start_time = graphene.DateTime()
-    end_time = graphene.DateTime()
-    state = graphene.String()
-    periods = graphene.List(graphene.Int)
-    is_reservable = graphene.Boolean()
+class ReservableTimeSpansGraphQLMixin:
```
matti-lamppu, line 22: Seems like this mixin is only used once (by `api.graphql.types.reservation_units.types.ReservationUnitByPkType`). Should just merge into that class.

opening_hours/models.py
```diff
@@ -0,0 +1,103 @@
...
+    latest_fetched_date = models.DateField(blank=True, null=True)
+
+    def __str__(self) -> str:
+        return str(self.id)
+
+    def is_reservable(self, start_datetime: datetime, end_datetime: datetime) -> bool:
+        return self.reservable_time_spans.filter_period(start=start_datetime, end=end_datetime).exists()
+
+
+class ReservableTimeSpanQuerySet(models.QuerySet):
```
matti-lamppu, line 26: Would put this QuerySet class to another file, e.g. `/opening_hours/querysets.py` so that the models.py is dedicated to only model classes.

api/graphql/types/opening_hours/types.py
```diff
@@ -1,187 +1,58 @@
...
-            init_times = False
-
-        opening_hours_client = OpeningHoursClient(
-            resources=self.hauki_resource_origin_id,
-            start_date=start,
-            end_date=end,
-            init_periods=init_periods,
-            init_opening_hours=init_times,
-            hauki_origin_id=self.hauki_resource_data_source_id,
+        origin_hauki_resource.reservable_time_spans.filter(
```
matti-lamppu, line 46: Unnecessary call? The next command uses `.filter_period` which uses `start_datetime__lt=end_date` & `end_datetime__gt=start_date`. There is a slight difference: `start_datetime__lte` vs `start_datetime__lt`. Intentional?

api/graphql/types/reservations/serializers/create_serializers.py
```diff
@@ -136,8 +136,8 @@ def __init__(self, *args, **kwargs):
         self.fields["purpose_pk"].required = False
 
     def validate(self, data, prefill_from_profile=True):
-        begin = data.get("begin", getattr(self.instance, "begin", None))
-        end = data.get("end", getattr(self.instance, "end", None))
+        begin: datetime.datetime = data.get("begin", getattr(self.instance, "begin", None))
+        end: datetime.datetime = data.get("end", getattr(self.instance, "end", None))
```
matti-lamppu, line 140: Correct type is `datetime.datetime | None`, since `getattr` default is `None`.

opening_hours/utils/hauki_api_types.py
matti-lamppu, line 1: Could `opening_hours.enums.State` be renamed to `ResourceState`?

opening_hours/models.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+    start_datetime = models.DateTimeField(null=False, blank=False)
+    end_datetime = models.DateTimeField(null=False, blank=False)
+
+    objects = ReservableTimeSpanQuerySet.as_manager()
+
+    class Meta:
+        ordering = ["resource", "start_datetime", "end_datetime"]
+        constraints = [
+            models.CheckConstraint(
```
matti-lamppu, line 84: How about a unique constraint that there can't be a duplicate timespan for the same resource?

opening_hours/utils/hauki_api_client.py
```diff
@@ -14,18 +21,84 @@
...
-        url = urljoin(settings.HAUKI_API_URL, f"/v1/{endpoint}/")
-        if resource_id:
-            url += f"{resource_id}/"
+    ############
+    # resource #
+    ############
+
+    def get_resources(self, *, hauki_resource_ids: list[int], **kwargs: Any) -> HaukiAPIResourceListResponse:
+        # Prepare the URL
+        url = urljoin(settings.HAUKI_API_URL, "/v1/resource/")
```
matti-lamppu, line 34: `urljoin` is a bit overkill here I think, could just use a simple f-string.

opening_hours/utils/hauki_api_client.py
```diff
@@ -14,18 +21,84 @@
...
+        query_params = {
+            "resource_ids": ",".join(str(id_) for id_ in hauki_resource_ids),
+            **kwargs,
+        }
+
+        # Get the data from Hauki API
+        response: HaukiAPIResourceListResponse = self.get(url=url, params=query_params)
```
matti-lamppu, line 41: Unnecessary assignment before return (I think this is a [`ruff` lint rule](https://docs.astral.sh/ruff/rules/unnecessary-assign/), should we enable it?)

opening_hours/utils/hauki_exporter.py
```diff
@@ -14,128 +16,110 @@ class HaukiResource:
...
-    def convert_to_request_data(self):
+
+class ReservationUnitHaukiExporter:
+    reservation_unit: ReservationUnit
+    hauki_api_client: HaukiAPIClient = HaukiAPIClient()
+
+    def __init__(self, reservation_unit: ReservationUnit):
+        self.reservation_unit = reservation_unit
+
+        if self.reservation_unit is None:
```
matti-lamppu, line 31: `reservation_unit` typed as just `ReservationUnit` and not `ReservationUnit | None`, so it should be the caller's responsibility to filter the value before it is given if it could be `None`.

opening_hours/utils/hauki_exporter.py
```diff
@@ -14,128 +16,110 @@ class HaukiResource:
...
+
+        # Unit doesn't have a hauki resource set, so try to get it from Hauki API
+        # If the unit doesn't have a tprek_id, we can't get it from hauki
+        if parent_unit.tprek_id is None:
+            return None
 
+        unit_origin_resource_id = f"tprek:{parent_unit.tprek_id}"
         try:
-            resource_data: HaukiAPIResource = HaukiAPIClient.get(url=url)
+            resource_data = HaukiAPIClient().get_resource(hauki_resource_id=unit_origin_resource_id)
```
matti-lamppu, line 101: Could use `self.hauki_api_client` instead of `HaukiAPIClient()`?

opening_hours/utils/hauki_exporter.py
```diff
@@ -14,128 +16,110 @@ class HaukiResource:
...
     organization: str
     origin_id: str
-    origin_data_source_name: str
-    origin_data_source_id: str
-    resource_type: ResourceType = ResourceType.RESERVABLE
 
-    def convert_to_request_data(self):
+
+class ReservationUnitHaukiExporter:
+    reservation_unit: ReservationUnit
```
matti-lamppu, line 25: Based on the fact that this class only needs the `reservation_unit` instance to work, these methods could be included in a `ReservationUnitActions` class.

utils/db.py
```diff
@@ -0,0 +1,12 @@
+from django.db.models import DecimalField, Subquery
+
+
+class SQSum(Subquery):
```
matti-lamppu, line 4: SQSum -> SQLSum?

ranta: Comes from **S**_ub_**Q**_uery_**Sum** :)

utils/db.py
matti-lamppu, line 1: I would put this in `/common` instead, perhaps `/common/expressions.py`. 

Side note: I'd like to get rid of the "utils" module since it has the same purpose as the common module - common code for all other apps. Or perhaps dedicate one module for all models (datamodel?) and another for common code (common/utils?).

utils/db.py
matti-lamppu: Ok, then I would just name it `SubquerySum`: it's not that long, and makes it more obvious what this does.

reservation_units/models.py
```diff
@@ -363,7 +362,14 @@ class ReservationUnit(SearchDocumentMixin, ExportModelOperationsMixin("reservati
 
     buffer_time_after = models.DurationField(verbose_name=_("Buffer time after reservation"), blank=True, null=True)
 
-    hauki_resource_id = models.CharField(verbose_name=_("Hauki resource id"), max_length=255, blank=True, null=True)
+    origin_hauki_resource = models.ForeignKey(
+        "opening_hours.OriginHaukiResource",
+        verbose_name=_("Origin Hauki resource"),
```
matti-lamppu, line 367: A verbose name is not needed.

opening_hours/utils/hauki_api_client.py
```diff
@@ -14,18 +21,84 @@
...
             raise HaukiConfigurationError("HAUKI_API_URL environment variable must to be configured.")
 
-        url = urljoin(settings.HAUKI_API_URL, f"/v1/{endpoint}/")
-        if resource_id:
-            url += f"{resource_id}/"
+    ############
+    # resource #
+    ############
+
+    def get_resources(self, *, hauki_resource_ids: list[int], **kwargs: Any) -> HaukiAPIResourceListResponse:
```
matti-lamppu, line 32: Could you document that `**kwargs` can be? Can be either with a TypedDict:

```python
_(..., **kwargs):
    kwargs: ResourceExtraParams
```

... or a link to the appropriate documentation on the Hauki side.

api/graphql/tests/test_reservation_units/test_reservation_units_query_opening_hours.py
```diff
@@ -0,0 +1,274 @@
...
+    assert response.get("errors") is None
+    assert response.data == {
+        "reservationUnitByPk": {
+            "reservableTimeSpans": None,
+        }
+    }
+
+
+def test__gql__reservation_unit__reservable_time_spans__no_start_date(graphql, reservation_unit):
+    reservation_unit.origin_hauki_resource = None
```
matti-lamppu, line 64: I think this should be removed, since if the `if not (start_date and end_date)` would be removed, then the `not origin_hauki_resource` would still return `None` and the test would pass.

utils/db.py
ranta: Good point, added this task to `TILA-2836`

api/graphql/tests/test_reservation_units/test_reservation_units_query_opening_hours.py
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+        """
+        % reservation_unit.id
+    )
+
+    assert response.get("errors") is None
+    assert response.data == {
+        "reservationUnitByPk": {
+            "reservableTimeSpans": None,
```
matti-lamppu, line 83: Not sure this is what I'd expect. Seem like you'd want to enforce that both `startDate` and `endDate` are required in the GraphQL schema, and give an error back if they are missing. Right now, they are typed as `Date`, which means they aren't required (Should be `Date!` if required). I think you can add `required=True`.

opening_hours/utils/hauki_api_client.py
ranta: `urljoin` should be fine here 🤔 
If we use simply join strings we must be able to assume that `HAUKI_API_URL` always or never ends in `/`

matti-lamppu: Urljoin also has some quirks that one might not expect:

```python
urljoin("https://github.com/City-of-Helsinki", "foo")
'https://github.com/foo'
```

matti-lamppu: I think you could just use 

```python
base = str(settings.HAUKI_API_URL).removesuffix('/')
url = f"{base}/v1/resource/"
```

opening_hours/tests/test_hauki_resource_hash_updater.py
```diff
@@ -0,0 +1,203 @@
...
+    updater._fetch_hauki_resources()
+
+    assert HaukiAPIClient.get_resources.call_count == 1
+    assert updater.fetched_hauki_resources == ["foo"]
+
+
+# First request uses `get_resources`
+@patch_method(HaukiAPIClient.get_resources, return_value={"results": ["foo"], "next": "page2"})
+# Later requests use `get`
+
```
matti-lamppu, line 61: Extra space here is a bit strange, seems like the decorator above is just "floating" even though it's attached to the test function.

opening_hours/tests/test_reservable_time_spans_client.py
```diff
@@ -0,0 +1,707 @@
...
+    Day 1:
+        10:00 - 16:00 (with reservation)
+        10:00 - 12:00 (open and reservable)
+        14:00 - 16:00 (closed)
+    Day 2:
+        10:00 - 20:00 (undefined)
+        18:00 - 06:00 (open and reservable)
+
+    Day 1 is reservable from 10:00 to 12:00 and from 14:00 to 16:00
+    Day 2 is reservable from 18:00 to 06:00
```
matti-lamppu, line 47: Really nice!

spaces/models.py
```diff
@@ -61,7 +61,15 @@ class Unit(models.Model):
     email = models.EmailField(verbose_name=_("Email"), blank=True, max_length=255)
     phone = models.CharField(verbose_name=_("Telephone"), blank=True, max_length=255)
 
-    hauki_resource_id = models.CharField(verbose_name=_("Hauki resource id"), max_length=255, blank=True, null=True)
+    origin_hauki_resource = models.ForeignKey(
+        "opening_hours.OriginHaukiResource",
+        verbose_name=_("Origin Hauki resource"),
```
matti-lamppu, line 66: Verbose name not required.

tests/decorators.py
matti-lamppu, line 1: This could probably be in `/tests/helpers.py`. Looks great though!

tests/test_graphql_api/test_reservation/test_create.py
```diff
@@ -16,14 +26,20 @@
 
 
 @pytest.mark.parametrize("arm", ["suomi_fi", "heltunnistussuomifi"])
+@freezegun.freeze_time("2021-01-01")
 def test_create_reservation__prefilled_with_profile_data(graphql, settings, arm: str):
     # given:
     # - Prefill setting is on
     # - There is a reservation unit in the system
```
matti-lamppu, line 33: Should include in this text that "the reservation unit has reservable time spans".

matti-lamppu (changes requested):
Finally was able to go through all of this. Discussed already what might need changes, but nothing major.

opening_hours/tests/test_reservable_time_spans_client.py
```diff
@@ -0,0 +1,707 @@
...
+    Contains cases of opening hours, to test the parsing logic.
+    Day 1:
+        10:00 - 16:00 (with reservation)
+        10:00 - 12:00 (open and reservable)
+        14:00 - 16:00 (closed)
+    Day 2:
+        10:00 - 20:00 (undefined)
+        18:00 - 06:00 (open and reservable)
+
+    Day 1 is reservable from 10:00 to 12:00 and from 14:00 to 16:00
```
vergama, line 46: I think the logic here is correct but is this specific comment invalid or am I misinterpreting it? Should it read (14-16.00 of day is closed): 
Day 1 is reservable from **10:00 to 14:00**
Day 2 is reservable from **20:00 to 06:00**

ranta: Good catch, yes looks like the comment part is incorrect.

api/graphql/tests/test_reservation_units/test_reservation_units_query_opening_hours.py
ranta: Adding filters to fields like so would require quite heavy modifications to the method, so I only changed these to use `build_query` by simply passing the filters in `fields` argument.

opening_hours/models.py
ranta: In theory there should never be a duplicate timespan for a resource, and even if there was, i dont think that other related functionality would be negatively affected by those :D

api/graphql/tests/test_reservation_units/test_reservation_units_query_opening_hours.py
matti-lamppu: OK, I can have a crack at implementing this elsewhere.

opening_hours/models.py
matti-lamppu: I guess given the current functionality duplicates don't really affect things. 

But, what about later when the frontend fetches reservable timespans for the calendar view on the reservation page? If there is a duplicate, would that affect drawing that view?

On the other hand, adding the constraint would mean that you'd need to modify the `ReservableTimeSpanClient` to account for duplicates, possibly with a hashmap instead of a list somewhere in gathering the TimeSpanElements. 

I'd add the constraint in favor of data correctness, but I'll leave it to you to what you want to do.

ranta: The `ReservableTimeSpanClient` already checks if the last-already-existing `ReservableTimeSpan` overlaps with the first-new-fetched-time-span, so duplicates should already be handled.

Also in `_init_date_range` fetching past dates is prevented by limiting the `start_date`, so we shouldn't be able to fetch any dates before our last existing `ReservableTimeSpan`.

matti-lamppu: Now that I think about it more, I guess the whole point of `ReservableTimeSpanClient` is to prevent duplicate ReservableTimeSpan elements due to the merging/splitting rules. So, I think it would be safe to add the constraint without any modifications.

actions/reservation_unit.py
```diff
@@ -0,0 +1,105 @@
+from typing import Any
+
+from opening_hours.errors import HaukiAPIError, HaukiRequestError, HaukiValueError
+from opening_hours.models import OriginHaukiResource
+from opening_hours.utils.hauki_api_client import HaukiAPIClient
+from opening_hours.utils.hauki_api_types import HaukiAPIResource, HaukiTranslatedField
+from reservation_units.models import ReservationUnit
+
+
+class ReservationUnitHaukiExporter:
```
matti-lamppu, line 10: Could rename this to `ReservationUnitActions`, create a connector in `/common/connectors.py` and then add that to the ReservationUnit model. Then, in `/api/graphql/types/reservation_units/mutations.py:ReservationUnitMutationMixin` instead of:

```python
exporter = ReservationUnitHaukiExporter(reservation_unit)
try:
    exporter.send_reservation_unit_to_hauki()
except (HaukiRequestError, HaukiAPIError):
    ...
```

...you could do:

```python
try:
    reservation_unit.actions.send_reservation_unit_to_hauki()
except (HaukiRequestError, HaukiAPIError):
    ...
```

opening_hours/utils/hauki_api_client.py
```diff
@@ -1,31 +1,127 @@
...
+    ############
+    # resource #
+    ############
+
+    @classmethod
+    def get_resources(
+        cls,
+        *,
+        hauki_resource_ids: list[int],
+        **kwargs: HaukiGetResourcesParams,
```
matti-lamppu, line 47: This is incorrect, `kwargs` are now typed as `dict[str, HaukiGetResourcesParams]`. This is the reason python 3.12 added [typing.Unpack](https://docs.python.org/3/library/typing.html#typing.Unpack). For now, you need to  untype the kwags, and add the annotation in the method body like I showed (or add [typing_extensions](https://pypi.org/project/typing-extensions/) and use it that way).

matti-lamppu (approved):
LGTM, should we go over this over video? @ranta @vergama I can do today after 13:00.

PR #1011, opened 2023-12-08
Reservation unit no longer reservable if no payment_product

matti-lamppu:
## 🛠️ Changelog

- `ReservationUnit` reservable state is no longer `RESERVABLE` if it has no `payment_product`

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-2944](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2944)

[TILA-2944]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vergama (changes requested):
Also need to check that the reservation unit needs payment, if it's free then the payment product is not needed.
Cf. [ReservationConfirmSerializer](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/c958235e67f22ba101da6a6cc772daba96fe93b8/api/graphql/types/reservations/serializers/confirm_serializers.py#L81): `active_price.pricing_type == PricingType.PAID`

matti-lamppu:
> Also need to check that the reservation unit needs payment, if it's free then the payment product is not needed. Cf. [ReservationConfirmSerializer](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/c958235e67f22ba101da6a6cc772daba96fe93b8/api/graphql/types/reservations/serializers/confirm_serializers.py#L81): `active_price.pricing_type == PricingType.PAID`

Would this be correct? @vergama 

```python
active_price = ReservationUnitPricingHelper.get_active_price(reservation_unit)

...

    (
        active_price is not None
        and active_price.pricing_type == PricingType.PAID
        and reservation_unit.payment_product is not None
    ) or (
        active_price is not None
        and active_price.pricing_type == PricingType.FREE
    )

```

vergama:
> Would this be correct? @vergama

Yeah that should do the trick.

matti-lamppu:
> > Would this be correct? @vergama
> 
> Yeah that should do the trick.

Fix in the newest commit.

vergama (changes requested):
I now realised that:

1. `ReservationState.SCHEDULED_CLOSING` also allows reservations being made, so the same extra conditions can be added to it, too. (Currently reservable but the res.unit has an ending time set in the future.)
2. Furthermore, if paid res.units missing the payment product are excluded from that state and `RESERVABLE`, then they should be included in the scope of `RESERVATION_CLOSED` to avoid an unknown state (assuming that their time range would have otherwise resolved as `RESERVABLE` or `SCHEDULED_CLOSING` but not one of the other states).

However, part **2** might lead to unnecessarily complex conditions for the `RESERVATION_CLOSED` check... I suppose that state could instead just be the default, if nothing else matches; it would work out for the UI by (a) disallowing reservations unless conditions expressly allow them by another state (the two above), or (b) informing users that reservation will be possible after a future date (`SCHEDULED_RESERVATION`, `SCHEDULED_PERIOD`).

matti-lamppu:
Should be finally fixed now. Refactored tests and added documentation diagrams.

reservation_units/utils/reservation_unit_reservation_state_helper.py
```diff
@@ -1,177 +1,285 @@
...
-                and (
-                    reservation_unit.reservation_begins
-                    and reservation_unit.reservation_begins <= now
-                    or reservation_unit.reservation_begins is None
-                )
+                and reservation_unit.reservation_begins is not None
+                and reservation_unit.reservation_begins <= now
+            )
+            # Reservations begin somewhere in the past (or right now),
+            # and end in the past, but begin before they end
```
vergama, line 137: Typo here: `but begin before they end` => `but begin (again) after they end`

vergama (changes requested):
Nice! Just one typofix to a single comment line needed.

reservation_units/utils/reservation_unit_reservation_state_helper.py
matti-lamppu: Fxed

vergama (approved):

PR #1020, opened 2023-12-13
Add new filters to application and application event schedule endpoints

matti-lamppu:
## 🛠️ Changelog

- Add new filters to applications and application event schedules according to the ticket specifications (application events already had everything)

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-2985](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2985)

[TILA-2985]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/common/db.py
+++ b/common/db.py
@@ -1,3 +1,4 @@
+from django.contrib.postgres.search import SearchQuery
 from django.db.models import DecimalField, Subquery
 
 
@@ -10,3 +11,15 @@ class SubQuerySum(Subquery):
     def __init__(self, queryset, output_field=None, *, sum_field="", **extra):
         extra["sum_field"] = sum_field
         super().__init__(queryset, output_field, **extra)
+
+
+def raw_prefixed_query(text: str) -> SearchQuery:
+    """
+    Create a query that searches for each word separately and matched partial words if match is a prefix.
+    Remove any whitespace and replace any single quote mark with two quote marks.
+    https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSQUERY
+    """
+    return SearchQuery(
+        " | ".join(f"'{val}':*" for value in text.split(" ") if (val := value.strip().replace("'", "''")) != ""),
+        search_type="raw",
+    )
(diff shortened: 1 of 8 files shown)


common/db.py
```diff
@@ -10,3 +11,15 @@ class SubQuerySum(Subquery):
...
+def raw_prefixed_query(text: str) -> SearchQuery:
+    """
+    Create a query that searches for each word separately and matched partial words if match is a prefix.
+    Remove any whitespace and replace any single quote mark with two quote marks.
+    https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSQUERY
+    """
+    return SearchQuery(
+        " | ".join(f"'{val}':*" for value in text.split(" ") if (val := value.strip().replace("'", "''")) != ""),
+        search_type="raw",
+    )
```
ranta, line 25: SonarCloud doesn't like this and we can get rid of the walrus operator and make it a bit cleaner like this:
```
    return SearchQuery(
        " | ".join(f"'{val.replace("'", "''")}':*" for value in text.split(" ") if value != ""),
        search_type="raw",
    )
```

- Since you are splitting the text by ` ` (space), there is no need to `strip` it since there will never be strings with the space character in it.
- The `.replace("'", "''")` can be moved left, since its not relevant in the equals check
- Finally what we have left is `... if value != ""`. Since empty strings evaluate to false, the `!= ""` can also be removed or kept for clarity's sake

matti-lamppu: Good simplification, although I do unfortunately need to use tripe quotes for the f-string since I'm already using single and double ones:

```python
def raw_prefixed_query(text: str) -> SearchQuery:
    return SearchQuery(
        " | ".join(f"""'{value.replace("'", "''")}':*""" for value in text.split(" ") if value != ""),
        search_type="raw",
    )
```

ranta (approved):

vergama (approved):

PR #1035, opened 2024-01-02
Update to Django 5.0

matti-lamppu:
## 🛠️ Changelog

- Updates Django to 5.0 release.
    - Required to backport PickleSerializer, as it was removed due to security issues. `django-helusers` still requires it.

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- TILA-####


matti-lamppu:
@vergama Here is the situation for updating to latest major version of Django. Some small things were removed/changed, and one package we use doesn't yet support it (needed to use lower version). Should we update, or wait until this is fixed?

vergama:
Need to get the pending features to production anyway, so can at least wait a little while. Looks like a new version of django-celery-beat just might be coming in the near future. But let's say, if it doesn't happen in the next few weeks then we'll go forward without it by the downgrade route.

cclauss:
https://github.com/celery/django-celery-beat/issues/680#issuecomment-1891665075 still needs a release to PyPI.

matti-lamppu:
This should be good to go now with `django-celery-beat` v2.6.0 released. @vergama @ranta

ranta (approved):
We can drop `django-prometheus` from the dependencies, as all things that previously used it have now been removed

PR #1041, opened 2024-01-04
Admin panel improvements

matti-lamppu:
## 🛠️ Changelog

- Various improvements to how admin panel shows data:

Show term type in terms of use list. Add filtering by term type:
terms of use

Show payment merchant and accounting in units list. Add filtering by them as well.
units list

Show unit counts for unit groups and service sectors in list views
unit groups list
service sectors list

Improve unit selection in unit groups and service sectors, and add selected unit count to edit view:
improved select unit group

Show unit after reservation unit in the reservation unit list
reservation units

Show space and space unit after resource in the resource list:
resource list

## 🧪 Test plan

- The admin panel still works, will get feedback from users and fix if something broke

## 🎫 Tickets

- [TILA-3035](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3035), [TILA-3036](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3036), [TILA-3037](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3037), [TILA-3038](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3038)

[TILA-3036]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TILA-3035]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spaces/models/space.py
+++ b/spaces/models/space.py
@@ -60,7 +60,7 @@ class Space(MPTTModel):
     def __str__(self) -> str:
         value = self.name
         if self.unit is not None:
-            value += f", {self.unit.name}"
+            value += f", {self.unit!s}"
         return value
 
     def save(self, *args, **kwargs):
--- a/spaces/models/space.py
+++ b/spaces/models/space.py
@@ -59,8 +59,6 @@ class Space(MPTTModel):
 
     def __str__(self) -> str:
         value = self.name
-        if self.building is not None:
-            value += f" ({self.building.name})"
         if self.unit is not None:
             value += f", {self.unit.name}"
         return value
--- a/spaces/models/space.py
+++ b/spaces/models/space.py
@@ -58,7 +58,12 @@ class Space(MPTTModel):
         base_manager_name = "objects"
 
     def __str__(self) -> str:
-        return f"{self.name} ({getattr(self.building, 'name', '')})"
+        value = self.name
+        if self.building is not None:
+            value += f" ({self.building.name})"
+        if self.unit is not None:
+            value += f", {self.unit.name}"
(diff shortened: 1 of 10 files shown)


spaces/models/space.py
```diff
@@ -60,7 +60,7 @@ class Meta:
     def __str__(self) -> str:
         value = self.name
         if self.unit is not None:
-            value += f", {self.unit.name}"
```
ranta: I think the `!s` here is redundant, as wouldn't`{self.unit}` just call the str() function on the `unit`, which is the same as `!s`? 🤔

ranta (approved):

spaces/models/space.py
```diff
@@ -60,7 +60,7 @@ class Meta:
...
+            value += f", {self.unit!s}"
```
ranta, line 63: I think the !s here is redundant, as wouldn't{self.unit} just call the str() function on the unit, which is the same as !s? 🤔

matti-lamppu: Wanted to be explicit here that I mean to use the str method, and not just the name property, as the unit str method has other info as well.

PR #1044, opened 2024-01-05
Fix overlapping reservation buffers

matti-lamppu:
## 🛠️ Changelog

- Fixes issues with reservation units whose reservations affect each other, but which have different buffer times.

Some reference pictures used in designing the solution:

image

image

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- TILA-####


tests/test_graphql_api/test_reservation_unit/test_query_reservable_fields.py
```diff
@@ -1250,3 +1272,82 @@ def test__query_reservation_unit_reservable__reservations__dont_include_cancelle
...
+    ReservationFactory.create(
+        begin=_datetime(year=2024, month=1, day=9, hour=13),
+        end=_datetime(year=2024, month=1, day=9, hour=15, minute=30),
+        buffer_time_before=reservation_unit.buffer_time_before,
+        buffer_time_after=reservation_unit.buffer_time_after,
+        reservation_unit=[reservation_unit],
+        state=ReservationStateChoice.CREATED,
+    )
+
+    # Monday | 2024-01-09 | 12:00 - 20:00 (10h)
```
vergama, line 1320: 1**2**:00 - 20:00 (1**0**h) => **10**:00 - 20:00 (**12h**)

matti-lamppu: Fixed

opening_hours/utils/reservable_time_span_client.py
```diff
@@ -79,40 +113,242 @@ def create_from_time_element(
...
+                reservation_unit.buffer_time_before.total_seconds() / 60
+            )
+        if reservation_unit.buffer_time_after is not None:
+            buffered_reservation_unit_minimum_duration_minutes += (
+                reservation_unit.buffer_time_after.total_seconds() / 60
+            )
+
+        return (
+            self.duration_minutes >= reservation_unit_minimum_duration_minutes
+            and self.buffered_duration_minutes >= buffered_reservation_unit_minimum_duration_minutes
```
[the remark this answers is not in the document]
matti-lamppu: True true, modified the check to check the buffed lengths from both ends.

reservation_units/querysets.py
```diff
@@ -170,17 +168,14 @@ def with_first_reservable_time(
         if filter_date_end is None:
             filter_date_end = two_years_from_now
 
-        # Time inputs without timezone information are interpreted as local time
+        # Time inputs are interpreted as local time even if they have timezone information.
+        # This is because we cannot know whether the time might be daylight saving time or not.
+        # Still, we remove any timezone information from the time input, so that time comparisons work correctly.
+        # (See. TODO)
```
ranta, line 174: TODO?

common/date_utils.py
```diff
@@ -1,93 +1,121 @@
...
-    ref: datetime.datetime | None = None,
-) -> datetime.datetime | datetime.time:
-    """
-    Convert a datetime to local timezone.
 
-    Use this instead of `datetime.astimezone()` or `datetime.datetime(..., tzinfo=local_timezone())`
-    to avoid issues with daylight savings time.
-    """
-    raise TypeError(f"Unsupported type: {_input.__class__.__name__}")
+class compare_datetimes(_TZComparator[datetime.datetime, datetime.datetime]):  # noqa: N801, RUF100
```
ranta, line 72: This is not actually used anywhere (yet?).

tests/test_utils/test_with_indices.py
```diff
@@ -0,0 +1,14 @@
+from common.utils import with_indices
+
+
+def test_with_indices():
+    test_list = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9]
+
+    for i, item in (gen := with_indices(test_list)):
+        if item % 2 == 0:
+            del test_list[i]
+            gen.item_deleted = True
```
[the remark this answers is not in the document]
matti-lamppu: Nice improvement! Done.

opening_hours/utils/reservable_time_span_client.py
```diff
@@ -32,45 +44,67 @@ class TimeSpanElement:
     start_datetime: datetime.datetime
     end_datetime: datetime.datetime
     is_reservable: bool
+    buffer_time_after: datetime.timedelta | None = None
```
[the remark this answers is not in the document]
matti-lamppu: Done.

common/date_utils.py
matti-lamppu: Not yet, should probably use it inside the TimeSpanElement class.

matti-lamppu: Actually, I don't think there is a need for it currently, as lt/lte/gt/gte comparisons already work correctly for datetime-objects, and there aren't any (not) equals checks. Code is simpler if we don't use it.

opening_hours/utils/time_span_element.py
```diff
@@ -0,0 +1,364 @@
...
+            )
+        if reservation_unit.buffer_time_after is not None:
+            back_buffered_reservation_unit_minimum_duration_minutes += (
+                reservation_unit.buffer_time_after.total_seconds() / 60
+            )
+
+        return (
+            self.duration_minutes >= reservation_unit_minimum_duration_minutes
+            and self.front_buffered_duration_minutes >= front_buffered_reservation_unit_minimum_duration_minutes
+            and self.back_buffered_duration_minutes >= back_buffered_reservation_unit_minimum_duration_minutes
```
vergama, line 277: ?Should there nevertheless be a 4th condition: `... and self.buffered_duration_minutes >= buffered_reservation_unit_minimum_duration_minutes`

Maybe this possibility is taken care of otherwise but even so for this logic to be self-contained entity:
- time span is 90 minutes with no buffers before or after
- min. res. duration is 30 minutes but res. unit has 60 min. buffer settings both before and after
- thus (90 >= 30 and 90 >= [90 = 60 + 30] and 90 >= [90 = 30 + 60]) = true, but the 90 min slot cannot fit the 60 + 30 + 60 = **150** total duration with buffers -- unless the slot boundaries are defined purely by opening hours, not other reservations

Furthermore: is it so that we are now not excluding those kind of slots where self.front/back_buffered... = self.duration_minutes because nominally there is no buffer (determined by existing reservation before or after) but actually there is buffer (before and/or after), in practice, when there is (hard) closed/non-reservable time, over which the buffers of the hypothetical reservation could be plaed?

matti-lamppu: Yes, fixed this and added test.

vergama (approved):

ranta (approved):

PR #1060, opened 2024-01-15
Reservee name

ranta:
## 🛠️ Changelog

- Extend Reservation.reservee_name with more options where to get the name
- Refactor some related old tests

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3022](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3022)

[TILA-3022]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


email_notification/sender/email_notification_context.py
```diff
@@ -88,11 +87,7 @@ def from_reservation(reservation: Reservation) -> "EmailNotificationContext":
         """Build context from reservation"""
         context = EmailNotificationContext()
 
-        if not reservation.reservee_type or reservation.reservee_type == CustomerTypeChoice.INDIVIDUAL.value:
-            context.reservee_name = f"{reservation.reservee_first_name} {reservation.reservee_last_name}"
-        else:
-            context.reservee_name = reservation.reservee_organisation_name
-
+        context.reservee_name = reservation.reservee_name
```
vergama, line 90: This is used in the salutation of the email text ("Hi FirstName LastName") so defaulting to reservation.name would be undesirable.

tests/test_graphql_api/test_reservation/test_query.py
```diff
@@ -138,7 +138,7 @@ def test_reservation__query__all_fields(graphql):
         "reserveeId": reservation.reservee_id,
         "reserveeIsUnregisteredAssociation": reservation.reservee_is_unregistered_association,
         "reserveeLastName": reservation.reservee_last_name,
-        "reserveeName": f"{reservation.reservee_first_name} {reservation.reservee_last_name}",
+        "reserveeName": f"{reservation.user.first_name} {reservation.user.last_name}",
```
matti-lamppu, line 141: `reservee_first_name` + `reservee_last_name` is not the same as `user.first_name` + `user.last_name` . The former names are saved to the reservation at the time of creation, which they might be changed after. @vergama can also check if this should be the case here.

ranta: This was an intentional change, as the `reservee_type` here is `None`, so user's name is used as fallback when we don't know if the reservee is an individual or organisation (and also `reservation.name` is not defined)

reservations/models.py
```diff
@@ -432,6 +432,40 @@ def requires_handling(self) -> bool:
...
 
+    @property
+    def reservee_name(self) -> str:
+        # Internal recurring reservations are created by STAFF
+        if self.type == ReservationTypeChoice.STAFF.value and self.recurring_reservation is not None:
+            return self.recurring_reservation.name
+
+        # Blocking reservation
+        if self.type == ReservationTypeChoice.BLOCKED.value:
+            return _("Closed")
```
vergama, line 443: Also need the locale message strings:
- fi: "Suljettu"
- sv: "Stängt"

email_notification/sender/email_notification_context.py
ranta: Changed these few rows back.

reservations/models.py
ranta: Added translations for this, but something to consider is that we are currently missing ~99% of translations. :)

vergama (approved):

PR #1075, opened 2024-01-19
Add option for one reservation per day

matti-lamppu:
## 🛠️ Changelog

- Add an option for reservation units where that reservation unit's reservations will always block the whole day, overriding any buffer time settings on the reservation unit

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3026](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3026), 
- [TILA-3027](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3027)


api/graphql/types/reservations/serializers/staff_reservation_modify_serializers.py
```diff
@@ -145,22 +147,28 @@ def validate(self, data):
...
+        end = data.get("end", self.instance.end).astimezone(DEFAULT_TIMEZONE)
 
-        new_buffer_before = data.get("buffer_time_before", None)
-        new_buffer_after = data.get("buffer_time_after", None)
+        new_buffer_before: datetime.timedelta | None = data.get("buffer_time_before", None)
+        new_buffer_after: datetime.timedelta | None = data.get("buffer_time_after", None)
 
+        reservation_unit: ReservationUnit
         for reservation_unit in self.instance.reservation_unit.all():
+            # Can't set buffers for whole day reservations
```
vergama, line 158: `StaffReservationModify` seems to be wrong of old, here it is not be possible to try to change the time information of the reservation, just the other details, so the existing buffers should be always kept.
That is, time and buffers always through `StaffReservationAdjustTime`; so like `begin` and `end`, `buffer_time_before` and `buffer_time_after` should not be mutable here. (Consequently the `overlap/buffer_times/intervals` validations are bogus here, I guess.)

I suppose the reason for the old implementation might have been that the option was reserved that staff could adjust just the buffers without moving the reservation itself to another time, but it would be better to do that with `StaffReservationAdjustTime`, especially now that the setting `reservation_block_whole_day` may apply.

api/graphql/types/reservations/serializers/adjust_time_serializers.py
```diff
@@ -32,7 +31,16 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["state"].readonly = True
 
-    def save(self, **kwargs):
+    def save(self) -> Reservation:
+        kwargs: dict[str, Any] = {}
+        for res_unit in self.instance.reservation_unit.all():
```
vergama, line 36: When user adjusts the time of the reservation, the buffers should always be re-calculated according to the current reservation unit settings (whether or not blocking the whole day), just like when creating a new reservation.

vergama (changes requested):
Two comments arising from the old code.

api/graphql/types/reservations/serializers/adjust_time_serializers.py
matti-lamppu: OK, is this a new rule, since I don't think reservations currently do this? I suppose there isn't much harm in it?

matti-lamppu: The only place I see this as being meaningful is when creating new reservations. Let's say that users make reservations with one buffer, which is then changed, and afterward, more users begin making reservations. The reservable time slots appear shorter or longer depending on whether the previous reservations users have edited their reservations or not.

reservation_units/utils/reservation_unit_reservation_scheduler.py
```diff
@@ -80,15 +80,11 @@ def is_reservation_unit_open(self, start_datetime: datetime, end_datetime: datet
...
-        interval_minutes: int,
-    ) -> set[datetime]:
+    def get_reservation_unit_possible_start_times(self, selected_date: date, interval_minutes: int) -> set[datetime]:
         resource = self.reservation_unit.origin_hauki_resource
         reservable_time_spans: QuerySet[ReservableTimeSpan] = resource.reservable_time_spans.filter(
-            start_datetime__date=selected_date,
-            end_datetime__date=selected_date,
+            start_datetime__date__lte=selected_date,
+            end_datetime__date__gte=selected_date,
         )
```
ranta, line 88: What was this change supposed to do? 🤔

actions/reservation_unit.py
```diff
@@ -119,3 +121,16 @@ def _update_hauki_resource(self, hauki_resource_data):
...
+            return time_as_timedelta(reservation_begin)
+        return self.reservation_unit.buffer_time_before or datetime.timedelta()
+
+    def get_actual_after_buffer(self, reservation_end: datetime.datetime | datetime.time) -> datetime.timedelta:
+        if self.reservation_unit.reservation_block_whole_day:
+            delta = time_as_timedelta(reservation_end)
+            if delta == datetime.timedelta():  # midnight
+                return delta
+            return datetime.timedelta(hours=24) - delta
+        return self.reservation_unit.buffer_time_after or datetime.timedelta()
```
ranta, line 136: Add comments why these helpers are needed

actions/reservation.py
```diff
@@ -0,0 +1,31 @@
...
+        return buffer_time_before
+
+    def get_actual_after_buffer(self) -> datetime.timedelta:
+        buffer_time_after: datetime.timedelta = self.reservation.buffer_time_after or datetime.timedelta()
+        reservation_unit: ReservationUnit
+        for reservation_unit in self.reservation.reservation_unit.all():
+            after = reservation_unit.actions.get_actual_after_buffer(self.reservation.end)
+            if after > buffer_time_after:
+                buffer_time_after = after
+        return buffer_time_after
```
ranta, line 31: Add comments why these helpers are needed

api/graphql/types/reservations/serializers/staff_reservation_modify_serializers.py
matti-lamppu: Need to check with frontend that these can be removed, but otherwise, I'm for removing them.

reservation_units/utils/reservation_unit_reservation_scheduler.py
matti-lamppu: This function did not pick up multi-day reservable timespans if they overlapped with the selected date, leading to no possible start times being found. Found this during testing, but didn't make a specific test for it.

api/graphql/types/reservations/serializers/staff_reservation_modify_serializers.py
matti-lamppu: Asked Joonatan, and they are not used, so removing is safe.

actions/reservation.py
matti-lamppu: I don't think these were used in the end. The idea was to calculate the reservation's actual current reservable time, and not the one at the time of creation, but based on previous tests, it seems that the set values should be honored (e.g. there was a case where staff reservations had 0 buffer set despite unit settings). @vergama might want to comment on what should be done here.

actions/reservation_unit.py
matti-lamppu: Done

api/graphql/types/reservations/serializers/staff_reservation_modify_serializers.py
matti-lamppu: Actually, the field is read-only, so couldn't have been used to write, but the checks were still there for some reason... Removing them.

matti-lamppu: _Actually_ actually, I just asked about the begin and end fields, mistakingly, and the buffers ARE mutated in the `staffReservationModify` endpoint. Removing them would require making 2 mutations with the current setup, so I suggest keeping this in until we can better refactor the reservation endpoint. @vergama

api/graphql/types/reservations/serializers/adjust_time_serializers.py
vergama: As it is/was, these two scenarios would give different end results for customer users:
A:
1. User makes reservation (say Monday 12-13.00 on date X) to a res.unit with 30 min buffer settings.
2. Res.unit settings are changed and the buffer lengths are increased to 60 mins. (Any existing reservations, including the above, still have their 30 min individual buffers, of course.)
3. User cancels the reservation, and then makes a new one for the following Thursday (date X + 2) where there is a free slot; the new one has the 60 min buffer settings applied to it.

B: same as A.1-2 plus:
3. Rather than cancelling the reservation, the user adjusts the time to the Thursday; it keeps the 30 buffers that the reservation originally had.

But that doesn't make sense since time adjustment is kind of shorthand for cancel & re-reserve, and by now it is known that the staff needs an hour to do whatever is needed for setup etc.
So the old logic of keeping the previously applied buffer times (whether zero or not), unless the res.unit "blocks the whole day", must be wrong.

matti-lamppu: That example is true, but there are still issues with buffers not being applied unless the user edits the reservation. The implication is that there is an "activation date" for the new buffer, even if it's "right now", and that all reservations after that time conform to the new buffer.

If new reservations should always conform to the reservation unit's buffer, what's the point of moving the buffer to the reservation?

matti-lamppu: One reason I can think of is admin reservations, where the buffer can be overridden. Perhaps there should be "buffer overrides" on the reservations, rather than always including a buffer?

actions/reservation.py
vergama: To clarify:

If something other than the time of an existing reservation is modified: the already set buffer values on it are honored.

If a customer user changes the time of a reservation, the new version should get the buffers set again, according to the current reservation unit settings (whether it is "whole day blocking" or has specific buffer lengths) -- since that corresponds to cancelling the existing reservation and creating a new one, as far as its target time is considered.

If a staff user changes the time of a reservation:
* Reservation unit is (at least now) "whole day blocking": the staff user has no choice and the buffers are set anew accordingly (as the implementation now is).
* Reservation unit has some specific buffer lengths (zero or otherwise, same or not as when the reservation was originally created): staff user has a choice of what buffers lengths to apply to the new version of the reservation (need not be constrained in backend); UI is supposed to guide the choice between 0 or current res.unit level setting ("no buffer" vs. "same as customer user would get") -- the last part doesn't actually quite work so but that is a frontend issue.

api/graphql/types/reservations/serializers/adjust_time_serializers.py
matti-lamppu: Anyways, removed the condition to skip updating buffers, so now every reservation will get new buffers on time adjust.

vergama: That's a fair point; admin reservations are probably the main reason. But secondarily, once a reservation exists and is not subsequently moved, it may be desirable to "lock" its buffers, too, and to be able to make fixed plans according to those.
Say a reservation has a 60 min buffer for cleanup afterwards, but the res.unit setting is later decreased to 30 mins for new reservations. But that specific cleanup in the real life may already have been scheduled, so don't want to allow another reservation *now* to be made only 30 mins after it.

(Whereas if a reservation is moved to another time altogether, all bets are off and I can't think of why its original buffers would have relevance over the current res.unit settings. I think that was just not accounted for when this serializer was written.)

matti-lamppu: Things to think about for the reservations refactor. I guess we can merge this and think about the edge case later?

actions/reservation.py
vergama: Yes.
At least for now it is valid that it is so (the existing buffers saved to the reservations are used, even if they no longer correspond to current and corresponding buffer settings at the reservation unit level).

ranta (approved):

PR #1085, opened 2024-01-24
Cache poetry installation between CI runs

matti-lamppu:
## 🛠️ Changelog

- ~~Cache Postgres `test_tvp` database between runs to avoid re-running migrations~~
   - ~~Improves CI time by 15s ~ 1 min (bottlenecked by CPU-intensive operations)~~
   - ~~Can cause issues if migrations are changed in multiple PRs. Currently, you'd need to delete the [cache-postgres-v1](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/actions/caches) cache instance and rerun the pipeline if this happens. Maybe there is a way to detect if migrations have changed?~~
- Cache poetry installation between runs
    - Improves CI time by 10s-30s

## 🧪 Test plan

- Chech that migrations run correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
     name: "Run tests and SonarCloud scan"
     runs-on: ubuntu-latest
     env:
-      POETRY_VERSION: 1.6.1
+      POETRY_VERSION: 1.7.1
 
     services:
       postgres:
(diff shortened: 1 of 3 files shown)


matti-lamppu:
@ranta I wanted to make our CI a bit faster by removing the need to do migrations at the start of tests, but it seems like there aren't that many gains to have from doing that. This would currently also introduce an issue, where we would need to clear the cached Postgres data if migrations are changed in multiple PRs at the same time, which could become annoying. What do you think?

matti-lamppu:
If we cache migrations per PR, then the initial run would still be "slow", i.e start without migrations.

My idea was that we would check whether migrations have changed in the PR and select whether to use/save to the cache or not based on that. My worry is that this is a bit too much complexity for little gain, and the real solution is to write faster tests (less db use, no elastic, etc.), merge migrations, or just to live with what we have.

ranta (approved):

PR #1129, opened 2024-03-11
Django 5 fixes

ranta:
## 🛠️ Changelog

- Fix Django Admin logout button
- Remove removed `USE_L10N` setting
- Use `SESSION_SERIALIZER` from `django-helusers`

## 🧪 Test plan

- Automated tests
- Try logging in locally with tunnistamo credentials

## 🎫 Tickets

- None


tilavarauspalvelu/settings.py
```diff
@@ -376,7 +376,7 @@
 
 LOGIN_REDIRECT_URL = "/admin/"
 LOGOUT_REDIRECT_URL = "/admin/"
-SESSION_SERIALIZER = "tilavarauspalvelu.utils.serializers.PickleSerializer"
+SESSION_SERIALIZER = "helusers.sessions.TunnistamoOIDCSerializer"
```
matti-lamppu, line 379: How does this affect current sessions? Do you get logged out or does it just work? Or are there complications?

matti-lamppu (approved):
Should check how the sessions behave after the serializer change, though I don't think there is much we can do other than be aware of what the impact is.

tilavarauspalvelu/settings.py
ranta: Should not affect current sessions at all (tested only locally)

PR #1143, opened 2024-04-04
Reservation Min/Max duration must be a multiple of Start Interval

ranta:
## 🛠️ Changelog

- Add validation for min/max reservation duration to ReservationUnit GQL endpoint

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3213](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3231)

[TILA-3213]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/graphql/types/reservation_unit/serializers.py
```diff
@@ -118,6 +121,53 @@ def validate(self, data: dict[str, Any]) -> dict[str, Any]:
...
         return data
 
+    def validate_reservation_duration_fields(self, data: dict[str, Any]) -> None:
+        """
+        Validates:
+        - min_reservation_duration
+        - max_reservation_duration
+        - reservation_start_interval
+        """
+        min_duration: timedelta | None = data.get("min_reservation_duration")
```
matti-lamppu, line 131: Use `self.get_or_default("min_reservation_duration", data)` (part of `NestingModelSerializer`), which will give you the current instance value if "min_reservation_duration" is not given in the update-mutation data. It'll also use the model field default value in create-mutation if the field is not required and not given there (though if the field has not default, this will be a [Sentinel](https://en.wikipedia.org/wiki/Sentinel_value) value "NotProvided", which is falsy)

matti-lamppu (changes requested):
Use `get_or_default` and make sure that the user cannot save inconsistent data during updates either.

api/graphql/types/reservation_unit/serializers.py
ranta: Changed those `get`s to `get_or_default`s

matti-lamppu (approved):

PR #1146, opened 2024-04-08
Existing user reservations count field for ReservationUnit gql endpoint

ranta:
## 🛠️ Changelog

- Refactor some reservation tests
  - Remove flakiness on tests when they are ran at night
- Move pruning settings that should be in settings to `settings.py`
- Add a field to tell how many reservations user already has for a reservation unit, which the frontend can use for validation

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3235](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3235)

[TILA-3235]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/graphql/types/reservation_unit/types.py
```diff
@@ -206,3 +209,12 @@ def resolve_reservable_time_spans(
...
         ]
+
+    def resolve_num_active_user_reservations(root: ReservationUnit, info: GQLInfo) -> int:
+        """
+        Number of active reservations made by the user to this ReservationUnit.
+        This is used to determine if the user can make a new reservation based on the max_reservations_per_user.
+        """
+        if not info.context.user.is_authenticated:
+            return 0
+        return root.reservation_set.filter(user=info.context.user).active().count()
```
matti-lamppu, line 220: The issue with this kind of resolver is that it will break optimization since it's evaluated for each reservation unit separately. Of course, this can't be a simple annotation either, since it needs to know about the current request user. If this is only used for single items, it can be fine, since then there is only a single additional query, but if it's used for lists, then it should be optimized somehow.

tests/helpers.py
```diff
@@ -96,3 +101,19 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
         return wrapper
 
     return decorator
+
+
+def next_hour(hours: int = 0, *, minutes: int = 0, days: int = 0) -> datetime.datetime:
```
matti-lamppu, line 106: Maybe argument names should be `plus_hours`, `plus_minutes`, and `plus_days` since we always get the next hour, and these just add more to that? I would also make all of them keyword-only for clarity.

tests/test_graphql_api/test_reservation/test_create.py
```diff
@@ -113,18 +116,18 @@ def test_reservation__create__use_longest_buffer_times_if_reserving_multiple_res
     #
     reservation_unit_1 = ReservationUnitFactory.create_reservable_now(
         sku="foo",
-        buffer_time_before=datetime.timedelta(),
-        buffer_time_after=datetime.timedelta(),
+        buffer_time_before=timedelta(),
```
matti-lamppu, line 119: Wasn't `from datetime import ...` supposed to become a banned import?

tests/test_graphql_api/test_reservation_unit/test_query.py
```diff
@@ -690,3 +691,64 @@ def test_reservation_unit__query__payment_product(graphql):
...
+
+
+def test_reservation_unit__query__num_active_user_reservations__user_unauthenticated(graphql):
+    reservation_unit = ReservationUnitFactory.create()
+
+    begin = next_hour(1)
+    end = begin + datetime.timedelta(hours=1)
+
+    # Reservation with a user
+    user = get_user_model().objects.create_user(username="other_user", email="other_user@django.com")
```
matti-lamppu, line 744: Use the `UserFactory` here.

matti-lamppu (approved):
Some comments, no real issues though.

tests/test_graphql_api/test_reservation/test_create.py
ranta: It's not yet banned _(and having shorter lines allowed for cleaner formatting on some places in the file 😄)_

api/graphql/types/reservation_unit/types.py
ranta: Yes, this is intended to be only be used for single reservation unit queries by the frontend.
While nothing is preventing it to be used for other queries too, I came to the same conclusion that this should be fine since its just one additional query and optimising it might be a difficult

matti-lamppu: I'll think about adding some mechanism to the optimizer library for this, but this is ok for now.

tests/test_graphql_api/test_reservation/test_create.py
matti-lamppu: _yet_. I guess it's fine.

PR #1154, opened 2024-04-17
Add SQL logging

matti-lamppu:
## 🛠️ Changelog

Adds SQL logging middleware.

This middleware can be used for logging SQL queries made to the backend in any HTTP request. This information can then be used to make sure the queries are optimized, which queries are most used, how long do the queries take, which queries require indexes, etc.

The middleware is off by default and can must be turned on with the `QUERY_LOGGING_ENABLED` env variable. The logger can be further tuned with the `QUERY_LOGGING_SKIP_ROUTES` setting to skip logging for some routes (e.g. `/admin` to not log any admin requests).

Logs are saved at the end of the request using a celery task, which is handled by the celery worker in production. This should make sure that the impact of saving logs for each request is minimal.

## 🧪 Test plan

The middleware can be tested locally by setting the `QUERY_LOGGING_ENABLED` env variable to `True` and the `CELERY_TASK_ALWAYS_EAGER = True` in `local_settings.py`. After this, logs should start generating for all requests. The admin panel should contain a section for "SQL logs", where the logs can be filtered further for analysis. Further development for processing the logs can be done in the future.


vergama:
Saving the request body without obfuscation might be a problem in production env.
Since they are in practice JSON data for GQL, maybe masking `variables` part would be sufficient?
(And the body simply not included in the logged data, if the request doesn't parse like that.)

matti-lamppu:
@vergama Changed logging so that the body is only captured for GraphQL requests, and only the query part is parsed out. This should remove any mutation variables from the logs.

vergama (approved):
Nice

ranta (approved):
Nice!
Might be also good to add `QUERY_LOGGING_ENABLED=false` to `.env.example`

matti-lamppu:
Merged PRs #1155, #1160, #1157, and #1177 to this PR so that they can be merged to main together.

PR #1155, opened 2024-04-17
Add Helsinki profile data endpoint

matti-lamppu:
## 🛠️ Changelog

- Adds an endpoint to the GraphQL schema for fetching user info from Helsinki profile.

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3164](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3164)

[TILA-3164]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/graphql/types/helsinki_profile/types.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @classmethod
+    def get_data(
+        cls,
+        info: GQLInfo,
+        *,
+        application_id: int | None = None,
+        reservation_id: int | None = None,
+    ) -> UserProfileInfo:
+        if not can_view_users(info.context.user):
```
vergama, line 43: Can't limit by `can_view_users()` since the _viewer_ and _handler_ roles do not have that permission; only by that have permission on the application or reservation.

matti-lamppu: Oh, I forgot to change this.

users/helauth/parsers.py
```diff
@@ -40,11 +50,43 @@ def parse(self) -> ReservationPrefillInfo:
...
+            phone=self.get_phone(),
+            birthday=self.get_birthday(),
+            ssn=self.get_social_security_number(),
+            street_address=address.get("address"),
+            postal_code=address.get("postalCode"),
+            city=address.get("city"),
+            municipality_code=self.get_municipality_code(),
+            municipality_name=self.get_municipality_name(),
+            login_method=LoginMethod.PROFILE,
+            is_strong_login=True,  # Helsinki profile login is always strong
```
vergama, line 74: Not necessarily in the future, could be password-only weak profile also in practice, if login with that is allowed at some point.
First thought was that this could be filled in from _user.id_token_ like here (whether its `loa` is `substantial`): https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1155/files#diff-3b2de07e9621e0695c8cd04e23191a197382943f8866bc9052a8d5352d85e6e1R71
**However**, that would simply reflect the last login method of the user in cases where profile has been created with both authentication methods (both `amr = suomi_fi` and `amr = helsinki_tunnus` i.e. password associated the same email in Helsinki-profile, where token for both gives the same `sub` claim but the latter has `loa = low` ).
**But** looks like it's possible to figure out that if `verifiedPersonalInformation` is not returned in the profile response despite being queried, then it's ultimately a weak/password-only profile.

users/helauth/clients.py
```diff
@@ -44,24 +49,219 @@ class HelsinkiProfileClient(BaseExternalServiceClient):
...
-        request_data = {"query": cls._reservation_prefill_query()}
+        request_data = {"query": query}
         headers = {"Authorization": f"Bearer {token}"}
 
         response = cls.post(url=url, json=request_data, headers=headers)
         response_data = cls.response_json(response)
 
-        my_profile_data: MyProfileData = get_nested(response_data, "data", "myProfile", default={})
-        return ReservationPrefillParser(my_profile_data).parse()
+        return get_nested(response_data, "data", "myProfile", default={})
```
vergama, line 160: Is it `myProfile` in the response if other user's profile is queried?

users/helauth/clients.py
```diff
@@ -44,24 +49,219 @@ class HelsinkiProfileClient(BaseExternalServiceClient):
...
+
+    @classproperty
+    def reservation_prefill_query(cls) -> str:
+        return (
+            """
+            query {
+                myProfile {
+                    firstName
+                    lastName
+                    ...MyPrimaryEmail
```
vergama, line 185: Nitpicking but could remove (or rename) the `My` prefix from the fragment names now that they can also be used to query another person's profile.

users/helauth/clients.py
matti-lamppu: Yes, this should be configurable.

users/helauth/clients.py
matti-lamppu: Makes sense, I'll remove them.

users/helauth/parsers.py
matti-lamppu: > But looks like it's possible to figure out that if verifiedPersonalInformation is not returned in the profile response despite being queried, then it's ultimately a weak/password-only profile.

Where did you get this information? Is it simply null in this case, or will there be an error? Currently, if the request to profile has any errors, our endpoint will fail.

vergama: null as part of the returned profile but the response also contains an error: "No permission to read verified personal information."
But come to think of it, I'll have to check if the response would be different when the profile has no suomi_fi association at all and could not be returned.

matti-lamppu: Counting on an error to tell us whether a profile is a certain kind seems like a bad idea. Their GraphQL documentation also states the following:

> **verifiedPersonalInformation**: **VerifiedPersonalInformationNode**
Personal information that has been verified to be true. Can result into `PERMISSION_DENIED_ERROR` if the requester has no required privileges to access this information.

This tells me that the requester's permissions are insufficient for querying this information, not that the requested user does no have it.

matti-lamppu: What is the use of `is_strong_login` on this endpoint anyway? Is it to indicate whether the user has strong login associated with their Helsinki profile account at all? In this case, it should be Helsinki profile providing that information to us. I added a comment on this to make a change once weak auth is implemented there, but before that, I don't think we need to do anything more here.

api/graphql/types/helsinki_profile/types.py
```diff
@@ -0,0 +1,144 @@
...
+            msg = f"Application with id {application_id} not found."
+            extensions = {"code": error_codes.HELSINKI_PROFILE_APPLICATION_USER_NOT_FOUND}
+            raise GraphQLError(msg, extensions=extensions)
+
+        if not can_read_application(info.context.user, application):
+            raise GQLNodePermissionDeniedError
+
+        user: User | None = application.user
+        if user is None:
+            msg = f"Application with id {application_id} does have a user."
```
vergama, line 105: does -> does not

users/helauth/parsers.py
```diff
@@ -40,11 +50,46 @@ def parse(self) -> ReservationPrefillInfo:
...
 
-    def get_last_name(self) -> str | None:
-        return self.data.get("lastName")
+    def parse_birthday_info(self) -> BirthdayInfo:
+        return BirthdayInfo(
+            id=self.data.get("id"),
+            birthday=self.get_birthday(),
+        )
+
+    def parse_user_profile_info(self, *, user: User) -> UserProfileInfo:
```
vergama, line 59: In the case of `parse_reservation_prefill_data` it is indeed desirable to prefer the profile level (user editable) name & address data as now, but in the added `parse_user_profile_info`, if `verifiedPersonalInformation` is present, then firstName/lastName and permanent(Foreign)Address should be attempted first, so that handlers will see the official DVV data of a person if it is available.

users/helauth/parsers.py
vergama: Yes to indicate whether a strong login is associated with the profile.
Could be simply based on self.data.get("verifiedPersonalInformation") is not None?

matti-lamppu: Added this to the parser.

users/helauth/parsers.py
matti-lamppu: Done

vergama (approved):
"That's Numberwang!"

ranta (approved):

PR #1160, opened 2024-04-18
Improve test data generation

matti-lamppu:
## 🛠️ Changelog

- Do not add spaces to reservation units randomly. Now each reservation unit has its own unit and space, and the units only space is also the same space as the reservation units'.
- Add additional reservation units with spaces in a hierarchy, Name them accordingly.
- Remove unnecessary space without unit and resource without space generation

## 🧪 Test plan

- Frontend will notify for any additional changes.


common/management/commands/create_test_data.py
```diff
@@ -580,7 +581,7 @@ def _create_service_sector_roles(
     text_entering="Creating units...",
     text_exiting="Units created!",
 )
-def _create_units(*, number: int = 15) -> list[Unit]:
+def _create_units(*, number: int = 300) -> list[Unit]:
```
ranta, line 584: Was this supposed to be `30`? `300` units is a lot 😄

ranta (approved):

common/management/commands/create_test_data.py
matti-lamppu: Previously, we generated 15 units and 300 reservation units, and used a unit randomly for each of them. Now we generate one unit per reservation unit, which for frontend testing was deemed easier.

PR #1171, opened 2024-04-24
Add new recuring reservation endpoint

matti-lamppu:
## 🛠️ Changelog

- Adds a new endpoint for creating a reservation series with its reservations in a since call
    - Validates that there are no overlapping reservations
    - Can optionally validate that reservations would fall within opening hours
    - Can skip certain dates if those dates already have a reservation where one would be created
- Refactors existing recurring reservations endpoint to use new serializer and error patterns

## 🧪 Test plan

- Automated tests

## 🎫 Tickets

- [TILA-3287](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3287)


common/date_utils.py
```diff
@@ -315,3 +317,39 @@ def merge_time_slots(time_slots: list[TimeSlot]) -> list[TimeSlot]:
...
+    :param end_date: Until which date to generate dates?
+    :param start_time: Start time of a period.
+    :param end_time: End time of a period.
+    :param interval: Days between each period.
+    :param tzinfo: Timezone information for the datetimes (if times are not datetime-aware yet).
+    """
+    if end_date < start_date:
+        msg = "End date cannot be before start date."
+        raise ValueError(msg)
+    if end_time < start_time:
```
vergama, line 347: Start and end time cannot be equal either.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +59,316 @@ class Meta:
...
-                "Begin date cannot be after end date.",
-                ValidationErrorCodes.RESERVATION_BEGIN_AFTER_END,
-            )
+            msg = "Begin date cannot be after end date."
+            raise ValidationError(msg, code=error_codes.RESERVATION_BEGIN_DATE_AFTER_END_DATE)
 
-        user = self.context.get("request").user
-        if user.is_anonymous:
-            user = None
+        if end_time < begin_time:
```
vergama, line 80: <= since can't allow zero-length reservations

common/date_utils.py
matti-lamppu: Fixed, and added tests for the utility

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: Fixed this and added tests.

matti-lamppu:
@vergama Can you define which fields for reservations should be definable, if not all?

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +59,316 @@ class Meta:
...
+            "end_time": {"required": False},
+            "reservation_unit": {"read_only": False},
+            "recurrence_in_days": {"required": False},
+        }
+
+
+class ReservationSeriesSerializer(RecurringReservationCreateSerializer):
+    """Create the recurring reservation with all its reservations."""
+
+    reservation_type = serializers.ChoiceField(
```
vergama, line 142: Also need to be able to supply other reservation details beside their type (optional, only the type is mandatory) -- also includes buffer time values, cf. https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/api/graphql/types/reservation/serializers/staff_create_serializers.py#L30

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +59,316 @@ class Meta:
...
+            reservation = Reservation(
+                name=instance.name,
+                description=instance.description,
+                begin=period["begin"],
+                end=period["end"],
+                user=reservation_user,
+                type=reservation_type,
+                recurring_reservation=instance,
+                age_group=instance.age_group,
+                buffer_time_before=instance.reservation_unit.buffer_time_before,
```
vergama, line 296: Buffer times should be according to request payload (UI defaults now to 0 but allows handler to select to use values which in practice are those defined on the reservation unit).

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: These fields are not included in the staff admin create, but are in the model: `sku`, `cancel_details`, `handling_details`, `price`, `price_net`,  `non_subsidised_price`, `non_subsidised_price_net`, `tax_percentage_value`, `deny_reason`, and `cancel_reason`. `user` is also not included, it always uses request user, which I don't think is correct.

vergama: Yes, these are not needed (either possibly set at update of individual reservation later or in the case of the price stuff not relevant for reservations created by stuff, at least for now): sku, cancel_details, handling_details, price, price_net, non_subsidised_price, non_subsidised_price_net, tax_percentage_value, deny_reason, and cancel_reason.
Other than those, `user` can indeed be supplied in the payload.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: Only request payload, so shouldn't use reservation unit values if buffers are omitted from the request?

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: For some reason, `unit_price` is updateable (only price field), is this correct?

vergama: No, `unit_price ` shouldn't be updateable.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +47,366 @@ class Meta:
...
+        reservations: list[Reservation] = []
+        through_models: list[models.Model] = []
+
+        for period in non_overlapping:
+            reservation = Reservation(
+                begin=period["begin"],
+                end=period["end"],
+                recurring_reservation=instance,
+                age_group=instance.age_group,
+                **reservation_details,
```
vergama, line 335: `buffer_before/after` needs to account for `reservation_unit.reservation_block_whole_day` (on basis of the date of each reservation): cf. https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/api/graphql/types/reservation/serializers/staff_create_serializers.py#L150C16-L150C61

api/graphql/types/recurring_reservation/serializers.py
vergama: Hmm... _StaffCreate_ (and _StaffAdjust_) says "Null/undefined value means buffer from reservation unit is used" but that only seems to apply for the validations, not what is in the end saved on the reservation... and when creating the series of reservations UI does not send zero values now but omits them from the request (only when adjusting the time of an existing reservation as staff member zero value is explicitly passed for buffer if it is not wanted on the reservation, but that is necessary so the old values that applied to the old version of the reservation are not kept).

Of course that could be changed and the UI needs to be updated anyway... but since the default scenario is that for staff created reservations buffers are not wanted, can be so that omitting them won't default to the res.unit settings and instead request payload needs to define the buffer values if they are wanted.

tests/test_graphql_api/test_recurring_reservation/test_create_series.py
```diff
@@ -0,0 +1,557 @@
...
+    assert reservations[0].state == ReservationStateChoice.CONFIRMED
+    assert reservations[0].type == ReservationTypeChoice.BEHALF
+    assert reservations[0].working_memo == "memo"
+    assert reservations[0].begin.astimezone(DEFAULT_TIMEZONE) == begin
+    assert reservations[0].end.astimezone(DEFAULT_TIMEZONE) == end
+    assert reservations[0].buffer_time_before == datetime.timedelta(minutes=15)
+    assert reservations[0].buffer_time_after == datetime.timedelta(minutes=30)
+    assert reservations[0].handled_at == datetime.datetime(2023, 1, 1, tzinfo=DEFAULT_TIMEZONE)
+    assert reservations[0].confirmed_at == datetime.datetime(2023, 1, 2, tzinfo=DEFAULT_TIMEZONE)
+    assert reservations[0].unit_price == Decimal("12.34")
```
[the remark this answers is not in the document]
matti-lamppu: Removed this.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: Removed this.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: Added support for this.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -126,3 +134,241 @@ class Meta(RecurringReservationCreateSerializer.Meta):
...
+                # Would the reservation overlap with any existing reservations?
+                # Also checks that buffers for the reservation will also not overlap.
+                if any(timespan.overlaps_with(reserved) for reserved in reserved_timespans):
+                    overlapping.append(
+                        ReservationPeriodJSON(
+                            begin=begin.isoformat(timespec="seconds"),
+                            end=end.isoformat(timespec="seconds"),
+                        )
+                    )
+                    continue
```
ranta, line 294: > Also checks that buffers for the reservation will also not overlap

This does not check the buffers for the existing reservations.
The `overlaps_with` method does check the buffers of the calling timespan (`timespan`), uses the buffers of the other timespan (`reserved`)
Above you set the `buffer_time_after and `..._before` to `timezone`, but the timespans returned from `get_reserved_timespans` in `reserved_timespans` don't have any buffer times set.

You should also check if any of the `reserved` overlap with with `timespan` to make sure the buffers don't overlap that way either.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -126,3 +134,241 @@ class Meta(RecurringReservationCreateSerializer.Meta):
...
+                    is_reservable=False,
+                )
+            )
+
+        return time_spans
+
+    def get_reservable_timespans(self, instance: RecurringReservation) -> list[TimeSpanElement]:
+        begin_time: datetime.time = instance.begin_time  # type: ignore[assignment]
+        end_time: datetime.time = instance.end_time  # type: ignore[assignment]
+        resource = instance.reservation_unit.origin_hauki_resource
```
ranta, line 409: `resource` -> `hauki_resource`
(to not confuse with `resources.models.Resource`)

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +47,372 @@ class Meta:
...
+                reservation=reservation,
+                reservationunit=instance.reservation_unit,
+            )
+            reservations.append(reservation)
+            through_models.append(through)
+
+        Reservation.objects.bulk_create(reservations)
+        ThroughModel.objects.bulk_create(through_models)
+
+    def get_reserved_timespans(
```
ranta, line 353: Consider using `Reservation.objects.get_affecting_reservations_as_closed_time_spans` and `merge_overlapping_time_span_elements` to replace this method.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +47,372 @@ class Meta:
...
+
+        # Create both the recurring reservation and the reservations in a transaction.
+        # This way if we get, e.g., overlapping reservations, the whole operation is rolled back.
+        with transaction.atomic():
+            instance = super().save()
+            self.create_reservations(instance, reservation_details, skip, check_opening_hours)
+
+        return instance
+
+    def create_reservations(
```
ranta, line 232: This method is missing some required validation checks for when to allow creating reservations.
For example, see:
- [ReservationUnitFirstReservableTimeHelper._get_hard_closed_time_spans](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/reservation_units/utils/first_reservable_time_helper.py#L377)
- [ReservationUnitFirstReservableTimeHelper._get_soft_closed_time_spans](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/reservation_units/utils/first_reservable_time_helper.py#L429)

ranta (changes requested):

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +47,372 @@ class Meta:
...
+        end_time: datetime.time = self.get_or_default("end_time", data)
+        reservation_unit: ReservationUnit = self.get_or_default("reservation_unit", data)
 
         if end_date < begin_date:
-            raise ValidationErrorWithCode(
-                "Begin date cannot be after end date.",
-                ValidationErrorCodes.RESERVATION_BEGIN_AFTER_END,
-            )
+            msg = "Begin date cannot be after end date."
+            raise ValidationError(msg, code=error_codes.RESERVATION_BEGIN_DATE_AFTER_END_DATE)
```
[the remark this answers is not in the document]
vergama: Max + 3 years into the future (inclusive of the same day of year as today) was the original requirement.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +51,78 @@ class Meta:
...
+
         return recurrence_in_days
+
+    @staticmethod
+    def validate_start_interval(reservation_unit: ReservationUnit, begin_time: datetime.time) -> None:
+        interval_minutes = ReservationStartInterval(reservation_unit.reservation_start_interval).as_number
+
+        # Staff reservations ignore start intervals longer than 30 minutes
+        if interval_minutes != 15:
+            interval_minutes = 30
```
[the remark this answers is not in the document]
matti-lamppu: Done.

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +51,78 @@ class Meta:
...
+
+        # For staff reservations, we don't need to care about opening hours,
+        # so we can just check start interval from the beginning of the day.
+        for hour in range(24):
+            for minute in range(0, 60, interval_minutes):
+                start_time = datetime.time(hour=hour, minute=minute, tzinfo=DEFAULT_TIMEZONE)
+                if start_time == begin_time:
+                    return
```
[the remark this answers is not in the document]
matti-lamppu: Good simplification. Done.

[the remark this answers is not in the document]
matti-lamppu: Actually, this needs to also check seconds and microseconds, but it'll still be simpler.

actions/reservation_unit.py
```diff
@@ -247,3 +248,46 @@ def reservation_units_with_common_hierarchy(self) -> models.QuerySet:
...
+
+        possible_start_times = set()
+        start_time = local_start_of_day(from_date)
+        interval_timedelta = datetime.timedelta(minutes=interval_minutes)
+
+        for time_span in reservable_time_spans:
+            # If the time span starts on the previous day, start from the beginning of selected date instead.
+            # Also handles cases where the time spans have a break smaller than the interval.
+            start_time = max(time_span.start_datetime, start_time).astimezone(DEFAULT_TIMEZONE)
+
```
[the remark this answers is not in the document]
matti-lamppu: Didn't write this code, just moved it from [ReservationUnitReservationScheduler](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/1c15f11e0a3d0f44905dc22dffbf515579f4d795/reservation_units/utils/reservation_unit_reservation_scheduler.py#L84-L104). Apparently, [you wrote this comment](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/commit/5a176a7e282210707e1cf16522c658e67a479ed4#diff-f82d96ba1233f30aea4d4131c7b6a48c4e75b47a0209a4c85230af19b7d4cda1) 😂

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -126,3 +134,241 @@ class Meta(RecurringReservationCreateSerializer.Meta):
...
+            self.create_reservations(instance, reservation_type, reservation_user, skip, check_opening_hours)
+
+        return instance
+
+    def create_reservations(
+        self,
+        instance: RecurringReservation,
+        reservation_type: ReservationTypeChoice,
+        reservation_user: User,
+        skip: list[datetime.date],
```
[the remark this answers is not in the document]
matti-lamppu: Done.

api/graphql/types/recurring_reservation/serializers.py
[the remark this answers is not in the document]
matti-lamppu: Added this check, and a regression tests.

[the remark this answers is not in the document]
ranta: @vergama We only have opening hours until 2 years into the future, so if creating recurring reservations more than te 2 years into the future, we would ALWAYS need to skip checking for opening hours

actions/reservation_unit.py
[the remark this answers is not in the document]
ranta: Nice to find ~_bugs_~ **_new features_** in my own code :DD

I wrote that comment 6 months ago, it's not possible for me to have any memory left of why I wrote that 🤣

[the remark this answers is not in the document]
ranta: Maybe just remove that comment line and add a note that this may not always work as espected with multi-day ReservableTimeSpans 🤷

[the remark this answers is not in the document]
ranta: Actually here is the method with a bit better validation that works past midnight and has a basic minimum duration check
(This still doesn't validate for many things which affect the actual reservability)

```py
def get_reservation_unit_possible_start_times(self, selected_date: date, interval_minutes: int) -> set[datetime]:
    if self.reservation_unit.origin_hauki_resource is None:
        return set()

    reservable_time_spans: ReservableTimeSpanQuerySet = ReservableTimeSpan.objects.filter(
        resource=self.reservation_unit.origin_hauki_resource,
        start_datetime__date__lte=selected_date,
        end_datetime__date__gte=selected_date,
    )

    possible_start_times = set()
    interval_timedelta = timedelta(minutes=interval_minutes)

    for time_span in reservable_time_spans:
        start_time = time_span.start_datetime.astimezone(DEFAULT_TIMEZONE)
        # Get all possible start times by looping through the time span until the time span or day ends
        while start_time < time_span.end_datetime:
            if start_time.date() < selected_date:
                continue
            if start_time - time_span.end_datetime < interval_timedelta:
                break
            possible_start_times.add(start_time)
            start_time += interval_timedelta
    return possible_start_times
```

api/graphql/types/recurring_reservation/serializers.py
```diff
@@ -38,52 +51,78 @@ class Meta:
...
-            )
+            msg = "Begin date cannot be after end date."
+            raise ValidationError(msg, code=error_codes.RESERVATION_BEGIN_DATE_AFTER_END_DATE)
 
-        user = self.context.get("request").user
-        if user.is_anonymous:
-            user = None
+        if end_time < begin_time:
+            msg = "Begin time cannot be after end time."
+            raise ValidationError(msg, code=error_codes.RESERVATION_BEGIN_TIME_AFTER_END_TIME)
```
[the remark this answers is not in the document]
matti-lamppu: Added a condition to this check that only checks this if the dates are the same.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: I think since recurring reservations follow staff reservation creation rules, we don't need to validate these ones. @vergama?

vergama: Yes, they do not limit staff reservations.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: Yes, this simplifies things. Fixed.

api/graphql/types/recurring_reservation/serializers.py
matti-lamppu: After using `get_affecting_reservations_as_closed_time_spans`, I'm checking overlap like this:

```python
# Would the reservation timespan overlap with any closing timespans
# that exist due to existing reservations? Checks for:
# 1) Unbuffered reservation timespan overlapping with any buffered closed timespan
# 2) Unbuffered closed timespan overlapping with any buffered reservation timespan
# 3) Unbuffered reservation timespan overlapping with any unbuffered blocked timespan
if (
    any(reservation_timespan.overlaps_with(closed) for closed in closed_timespans)
    or any(closed.overlaps_with(reservation_timespan) for closed in closed_timespans)
    or any(reservation_timespan.overlaps_with(blocked) for blocked in blocked_timespans)
):
    results.overlapping.append(ReservationPeriod(begin=begin, end=end))
    continue
```

vergama (approved):
Fine-tuning of the buffer logic can be taken care of in https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1185

actions/recurring_reservation.py
```diff
@@ -0,0 +1,185 @@
...
+                # Would the reservation timespan overlap with any closing timespans
+                # that exist due to existing reservations? Checks for:
+                # 1) Unbuffered reservation timespan overlapping with any buffered closed timespan
+                # 2) Unbuffered closed timespan overlapping with any buffered reservation timespan
+                # 3) Unbuffered reservation timespan overlapping with any unbuffered blocked timespan
+                if (
+                    any(reservation_timespan.overlaps_with(closed) for closed in closed_timespans)
+                    or any(closed.overlaps_with(reservation_timespan) for closed in closed_timespans)
+                    or any(reservation_timespan.overlaps_with(blocked) for blocked in blocked_timespans)
+                ):
```
ranta, line 119: `closed` and `blocked` shadows the variable names declared earlier.
This is really **not an issue that must be fixed**, since the original variables aren't used later, but maybe just a bit misleading? 😅

ranta (approved):

PR #1184, opened 2024-05-02
Remove all service sector checks from backend

matti-lamppu:
## 🛠️ Changelog

- Removes all service sector related checks and tests from backend. This is a part of removing service sector from the backend entirely
- Replaces some service sector permissions with equivalent unit group permission checks

> Its hard to say if this removes _all_ checks, since there can be all kinds of string references about, but most common access patterns have been checked with grep.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3298][TILA]

[TILA]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3298

[TILA-3298]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


permissions/helpers.py
```diff
@@ -630,7 +535,4 @@ def get_service_sectors_where_can_view_reservations(user: AnyUser) -> models.Que
     if has_general_permission(user, general_permission):
         return ServiceSector.objects.all().values("pk")
 
-    service_sector_ids = [
-        pk for pk, perms in user.service_sector_permissions.items() if service_sector_permission.value in perms
-    ]
-    return ServiceSector.objects.filter(id__in=service_sector_ids).values("pk")
+    return ServiceSector.objects.none()
```
vergama, line 538: Looks like `get_service_sectors_where_can_view_reservations` can be dropped altogether if it is removed also from RecurringReservationNode: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/b1a1e9d644ce26ee12911d04ca3405915f3061fc/api/graphql/types/recurring_reservation/types.py#L55

permissions/getters.py
```diff
@@ -46,6 +41,4 @@ def get_service_sectors_with_permission(user: AnyUser, *, permission: str) -> mo
     if has_general_permission(user, permission):
         return ServiceSector.objects.all().values("pk")
 
-    service_sector_ids = [pk for pk, perms in user.service_sector_permissions.items() if permission in perms]
-
-    return ServiceSector.objects.filter(models.Q(id__in=service_sector_ids)).values("pk")
+    return ServiceSector.objects.none()
```
vergama, line 44: `get_service_sectors_with_permission` can be dropped if `ReservationFilterSet.get_only_with_handling_permission` was the only place which used it.

permissions/helpers.py
```diff
@@ -95,17 +81,15 @@ def can_modify_reservation_unit(user: AnyUser, reservation_unit: ReservationUnit
     return can_manage_units_reservation_units(user, reservation_unit.unit)
 
 
-def can_manage_service_sectors_applications(user: AnyUser, service_sector: ServiceSector) -> bool:
+def can_manage_service_sectors_applications(user: AnyUser) -> bool:
```
vergama, line 84: This can be removed if the calls to `can_manage_service_sectors_applications` are changed in `ReservationUnitOptionPermission` and `AllocatedTimeSlotPermission`, so that update is permitted on them by virtue of the user having `CAN_HANDLE_APPLICATIONS` on the unit (of the reservation unit) of the option being updated or allocated.

api/graphql/types/allocated_time_slot/permissions.py
```diff
@@ -36,17 +36,12 @@ def has_create_permission(cls, user: AnyUser, input_data: dict[str, Any]) -> boo
...
-            .first()
-        )
+        option: ReservationUnitOption | None = ReservationUnitOption.objects.filter(pk=option_pk).first()
         if not option:
             msg = f"Reservation Unit Option with pk {option_pk} does not exist."
             raise GQLCodeError(msg, code=error_codes.ENTITY_NOT_FOUND)
 
-        sector = option.application_section.application.application_round.service_sector
-        return can_manage_service_sectors_applications(user, sector)
+        return can_manage_service_sectors_applications(user)
```
vergama, line 44: Let's replace the `can_manage_service_sectors_applications` check with the user having the `CAN_HANDLE_APPLICATIONS` permission on the unit (of the reservation unit) of the option which is being allocated.

api/graphql/types/allocated_time_slot/permissions.py
```diff
@@ -57,5 +52,4 @@ def has_delete_permission(cls, instance: AllocatedTimeSlot, user: AnyUser, input
         if not user.has_staff_permissions:
             return False
 
-        sector = instance.reservation_unit_option.application_section.application.application_round.service_sector
-        return can_manage_service_sectors_applications(user, sector)
+        return can_manage_service_sectors_applications(user)
```
vergama, line 55: Same comment as create, we can do away with `can_manage_service_sectors_applications` here too.

api/graphql/types/reservation_unit_option/permissions.py
```diff
@@ -29,5 +29,4 @@ def has_update_permission(cls, instance: ReservationUnitOption, user: AnyUser, i
         if not user.has_staff_permissions:
             return False
 
-        sector = instance.application_section.application.application_round.service_sector
-        return can_manage_service_sectors_applications(user, sector)
+        return can_manage_service_sectors_applications(user)
```
vergama, line 32: Similar to `AllocatedTimeSlotPermission`, the proper check for update permission on `ReservationUnitOption `would now be that the user (if not super user) must have the `CAN_HANDLE_APPLICATIONS` permission on the unit of the option, either from unit or general roles.
(Then can_manage_service_sectors_applications can be dropped from the permission helpers. Or perhaps replace with a new helper, `can_allocate_application` which is scoped at unit?])

Although in this case, should the user of the application be permitted as well, if the application period hasn't passed?

permissions/helpers.py
matti-lamppu: Yes, done.

permissions/helpers.py
matti-lamppu: Done.

api/graphql/types/allocated_time_slot/permissions.py
matti-lamppu: Done.

api/graphql/types/allocated_time_slot/permissions.py
matti-lamppu: Done.

api/graphql/types/reservation_unit_option/permissions.py
matti-lamppu: Update operations on the reservation unit options are limited to rejecting or locking the option, which a regular user shouldn't be doing. Creation is handled differently, as options are handled in bulk together with application sections, and thus follow its permissions.

I modified the check like all the rests.

vergama (approved):
Looks good, can't think of anything else.

PR #1234, opened 2024-06-12
Add fields to reservation statistics

matti-lamppu:
## 🛠️ Changelog
[//]: # "Describe the changes in this pull request here."

- Adds fields as described in the JIRA ticket
- Refactors and groups statistics model fields

## 🧪 Test plan
[//]: # "Help your fellow reviewer and write a short description of the fastest way to test your changes."

- Automated tests

## 🚧 Dependencies
[//]: # "Is this PR dependent on other changes outside this repository? Describe the changes, or add links to them."
[//]: # "e.g. This PR breaks X and is waiting for frontend changes before merging."

- None

## 🎫 Tickets
[//]: # "This pull request resolves all or part of the following ticket(s)."

- [TILA-3424][TILA]

[TILA]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3424

[TILA-3424]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1252, opened 2024-06-26
Speed up first reservable time calculation

matti-lamppu:
## 🛠️ Changelog

Speeds up first reservable time calculation in a two different ways:

1. Pre-calculates affecting reservations with a materialized view. The view should be updated with a scheduled task.

2. Run calculation for fewer units based on pagination arguments. Use a custom `QuerySet.iterator()` to fetch chunks of reservation units and their affecting reservations

The above optimizations achieve <1s fetching with our test data when calculating FRT, regardless of `showOnlyReservable` or any pagination args.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Setup a periodic job in Django to refresh AffectingTimeSpans in platta envs.

## 🎫 Tickets

- [TILA-3436][TILA]

[TILA]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3436

[TILA-3436]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matti-lamppu:
Idea: Maybe there should be a flag set in cache which indicates that the materialized view refresh is valid, in case the job fails for some reason. This would then force a refresh in the endpoint if the value is missing.

reservations/migrations/0073_affectingtimespan.py
```diff
@@ -0,0 +1,121 @@
...
+                        related_name="affecting_time_span",
+                        serialize=False,
+                        to="reservations.reservation",
+                    ),
+                ),
+                (
+                    "affected_reservation_unit_ids",
+                    django.contrib.postgres.fields.ArrayField(base_field=models.IntegerField(), size=None),
+                ),
+                ("is_reservable", models.BooleanField()),
```
vergama, line 106: Should this be `is_blocking`?

matti-lamppu: Yes, Fixed this.

reservations/migrations/0073_affectingtimespan.py
```diff
@@ -0,0 +1,121 @@
...
+                        )
+                    ) res
+                    GROUP BY
+                        res.reservation_id,
+                        res.buffered_start_datetime,
+                        res.buffered_end_datetime,
+                        res.buffer_time_before,
+                        res.buffer_time_after,
+                        res.is_blocking
+                    ORDER BY res.buffered_start_datetime, res.reservation_id;
```
vergama, line 51: Does it matter [that](https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html):
> If there is an ORDER BY clause in the materialized view's defining query, the original contents of the materialized view will be ordered that way; but REFRESH MATERIALIZED VIEW does not guarantee to preserve that ordering.

Not because of `AffectingTimeSpan.Meta.ordering`?

vergama:
> Idea: Maybe there should be a flag set in cache which indicates that the materialized view refresh is valid, in case the job fails for some reason. This would then force a refresh in the endpoint if the value is missing.

Or value of when the view was last updated; force refresh if too old?
Some complications may be caused by parallel endpoint hits encountering invalid mat.view, in which case the refresh should be synchronized...

reservations/migrations/0073_affectingtimespan.py
matti-lamppu: Oh, good to know. Still, I think this should be fine since AffectingTimeSpans are ordered chronologically when they are merged in `FirstReservableTimeHelper._get_elements_for_reservation_units`.

vergama (approved):
"The dude abides."

matti-lamppu:
Added a value to cache for when the view was last updated. The FRT calculation will check this and refresh the view if it is too old. Refreshing the view with `AffectingTimeSpan.refresh()` updates the cached value.

Also, added error handling logic to the `.refresh()` methods on both materialized views to raise an error if the refresh fails during local development, but log to sentry in other environments.

ranta (approved):

matti-lamppu:
Merged #1234, #1274, #1261 to this PR so that they can be merged to main together.

PR #1289, opened 2024-08-13
Fix FRT pagination optimization & use caching to make it faster

matti-lamppu:
## 🛠️ Changelog

- Full explanation will be written as an ADR.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Required frontend changes to make pagination work, but can be merged before those are done, and it will still make the page much faster to load due to cached results.

## 🎫 Tickets

- [TILA-3480](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3480)

[TILA-3480]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reservation_units/utils/first_reservable_time_helper/first_reservable_time_helper.py
+++ b/reservation_units/utils/first_reservable_time_helper/first_reservable_time_helper.py
@@ -211,7 +211,7 @@ class FirstReservableTimeHelper:
         ##############
 
         if pagination_args is not None:
-            pagination_args["size"] = graphene_settings.RELAY_CONNECTION_MAX_LIMIT
+            pagination_args["size"] = reservation_unit_queryset.count()
             qs_slice = calculate_queryset_slice(**pagination_args)
             self.start_offset = qs_slice.start
             self.stop_offset = qs_slice.stop
@@ -245,7 +245,7 @@ class FirstReservableTimeHelper:
 
         self.first_reservable_times = {}
         self.reservation_unit_closed_statuses = {}
-        self.cached_value_validity = {}
+        self.cached_value_validity: dict[int, datetime.datetime] = {}
 
         # Closed time spans that are shared by all ReservationUnits
         self.shared_hard_closed_time_spans = self._get_shared_hard_closed_time_spans()
@@ -277,20 +277,23 @@ class FirstReservableTimeHelper:
             # If we already have cached FRT results for enough reservation units to fill the page
             # AND all the previous pages, then we don't need to calculate anything.
             # NOTE: This does not support different orderings of the same queryset!
-            required = (
+            # TODO: This doesn't work for the last page if it's not full!
+            cached = (
                 sum(1 for frt in self.first_reservable_times.values() if frt is not None)
                 if self.show_only_reservable
                 else len(self.first_reservable_times)
             )
-            if required >= self.stop_offset:
+            if cached >= self.stop_offset:
                 return
 
-            # Otherwise, we can still skip looping through the previous pages.
-            qs = qs[self.start_offset :]
-            self.start_offset = 0  # Remove offset since we skip the previous pages
+            # Otherwise, we should still have valid results for the previous pages.
+            # We can start calculating after the last cached result.
+            qs = qs[len(self.first_reservable_times) :]
+            # We also don't need to skip any results that are not already cached.
+            self.start_offset = 0
 
         # If we don't have valid results, and this is not the first page,
-        # we should fetch the current and previous pages in on chunk. The next chunk
+        # we should fetch the current and previous pages in one chunk. The next chunk
         # will also be bigger (if needed), but likely small enough (<100) not to cause any problems.
         elif self.start_offset > 0:
             self.chunk_size = self.stop_offset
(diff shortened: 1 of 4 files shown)


ranta (approved):
Really nice stuff 🚀

reservation_units/utils/first_reservable_time_helper/first_reservable_time_helper.py
```diff
@@ -219,34 +269,85 @@ def calculate_all_first_reservable_times(self) -> None:
...
-
-        for reservation_unit in qs.hooked_iterator(self._get_elements_for_reservation_units, chunk_size=size):
+            cache.delete(self.cache_key)
+
+        has_valid_results_for_previous_pages = self._read_cached_results()
+        if has_valid_results_for_previous_pages:
+            # If we already have cached FRT results for enough reservation units to fill the page
+            # AND all the previous pages, then we don't need to calculate anything.
+            # NOTE: This does not support different orderings of the same queryset!
+            if len(self.first_reservable_times) >= self.stop_offset:
```
vergama, line 279: Is the length check sufficient; i.e., there cannot be gaps in the data (but so that it extends to further pages because someone else had queried that far)?
I guess it at least may assume that the pages are queried in order from the API -- if so, I suppose we can make that assumption (at least for now because the frontend search page is the use case) -- unless the cache being shared between users might lead to some tricky situation I haven't thought through.

matti-lamppu: It should be, since cached are unique per ordering too, and the entire cache gets invalidated an recalculated if even one of the items is no longer valid (i.e. items in page 1 become invalid -> must calculate all pages). Maybe there is some slight edge case when new reservation units are added, but those are very marginal cases that get corrected quickly.

vergama: I was thinking about the case that if there's a page in the middle (say page 3) which has not been queried (but now page 4 is being queried) and its data has never been cached but enough later pages have been cached so that the length check passes. But I think this would only come to pass if someone(s) queries the pages out of sequence (e.g. 1, 2, 5, 4, 3) directly from the API, bypassing the frontend, and need not be supported for now.

matti-lamppu: Yes, in regular use, you never query pages out of order. In any case, I don't think there is a way to know in that situation that the latter items are from page 4 and not 3 without calculating all of them.

reservation_units/utils/first_reservable_time_helper/first_reservable_time_helper.py
```diff
@@ -219,34 +269,90 @@ def calculate_all_first_reservable_times(self) -> None:
...
+            # current page only if there is a first reservable time.
+            if not self.show_only_reservable or first_reservable_time is not None:
+                results += 1
+
+            # This only really matters when showing only reservable reservation units,
+            # since we might need to fetch more than the first chunk size.
+            if results >= self.needed_reservation_units:
+                break
+
+        self._cache_results()
```
vergama, line 321: Does this not effectively extend the caching period for the ru.pks (of the previously queried pages) that were retrieved from the cache during the processing of this request -- so that if there are successive queries with the same parameters (and hence cache key, such as the default search) continuously made faster than the cache expiration time (the more likely due to multiple users in production), then the cache never expires?

vergama: ... or I suppose at least one client must encounter AffectingTimeSpan being invalid, and that clears the cache, though.

matti-lamppu: It's not the cache itself that dictates whether the results are invalid. There is a `valid_unit` parameter attached to each cached result, which is checked on each request to validate a specific value. Then even if one of those values is invalid, the whole cache is recalculated.

vergama: There is but the `valid_until` value is set again, to the same value for the previous page items as the items from the current page which are newly added to the cache (if I'm reading it correctly)?
Even if so, it may not matter in practical terms.

vergama: Nothing else pops out. Nice tests.

matti-lamppu:
Here are the current profiling results from `py-spy`:

profile

And from `pyinstrument`:

image

Seems like the current bottleneck is still in preparing `AffectingTimeSpan`s for overlap operations, since there is a lot of `TimeSpanElement` creation and merging happening when we fetch results for a chunk of reservation units. Actually pulling the data from the database doesn't take that much time, and it's the conversion that takes about 2 seconds on my machine with the test data.

What we could do here is to calculate the `TimeSpanElement`s on demand when we are trying to find the FRT, but that would be a bigger code change, so we can leave that for later.

vergama:
> What we could do here is to calculate the `TimeSpanElement`s on demand when we are trying to find the FRT (i.e. a week at a time), but that would be a bigger code change, so we can leave that for later.

Agree. 
That's some heavy lifting going on in those merge operations but as long as the database is not the bottleneck, I'm sure it can be optimized later.

vergama (approved):
_Pretty, pretty, pretty.... Pretty good!_

PR #1306, opened 2024-08-29
Log text searches & add admin panel view for listing them

matti-lamppu:
## 🛠️ Changelog

- Log text searches by where the search happens and at what time.
- Display searches from the last 30 days in the admin panel.

## 🧪 Test plan

- Automated tests
- Check admin panel views

## 🚧 Dependencies

- None

## 🎫 Tickets

- None


ranta (approved):

common/utils.py
```diff
@@ -148,6 +154,11 @@ def translate_for_user(text: Promise, user: User) -> str:
         return str(text)
 
 
+def get_text_search_language(request: HttpRequest) -> Literal["finnish", "english", "swedish"]:
+    lang_code: Literal["fi", "en", "sv"] = get_language_from_request(request)
```
vergama, line 158: Ultimately relying on the `Accept-Language` header won't work since the browser may be in another locale than the selected frontend language (or at least it is debatable which one should be respected) -- but does this also examine the `Referer` header?

vergama (approved):

common/utils.py
matti-lamppu: Frontend should be setting `Accept-Language` on requests according to the language selected. In any case, this is not yet used, since there weren't any translatable fields in the current data.

PR #1304, opened 2024-08-28
Convert Reservation net price fields into properties

ranta:
## 🛠️ Changelog
[//]: # "Describe the changes in this pull request here."

- Convert Reservation net price fields into properties
- This should not require changes from the frontend, as the fields should be simply ignored if they are sent

## 🧪 Test plan
[//]: # "Help your fellow reviewer and write a short description of the fastest way to test your changes."

- Automated tests
- Might be good to manually test as well just in case :D

## 🚧 Dependencies
[//]: # "Is this PR dependent on other changes outside this repository? Describe the changes, or add links to them."
[//]: # "e.g. This PR breaks X and is waiting for frontend changes before merging."

- None

## 🎫 Tickets
[//]: # "This pull request resolves all or part of the following ticket(s)."

- [TILA-3485](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3485)

[TILA-3485]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1317, opened 2024-09-04
Deprecate ReservationUnitPricing fields

ranta:
## 🛠️ Changelog

- Deprecate `ReservationUnitPricing.status` field
- Deprecate `ReservationUnitPricing.pricing_type` field
- Add `ReservationUnitPricing.is_activated_on_begins` field
- Block publishing a reservation unit with invalid pricings

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3486](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3486)


api/graphql/types/reservation_unit/serializers.py
```diff
@@ -206,34 +202,52 @@ def validate_for_publish(self, data: dict[str, Any]) -> None:
...
 
-            if (
-                ReservationUnitPricingHelper.is_future(pricing)
-                and current_future_pricing
-                and current_future_pricing.pk != pricing.get("pk", 0)
-            ):
-                raise GraphQLError("FUTURE pricing is already defined. Only one FUTURE pricing is allowed")
+        # At least one given pricing must be currently active (begin date is today or earlier)
+        today = local_date()
+        has_active_pricing = any(p.get("begins") <= today for p in pricings)
```
matti-lamppu, line 227: I'm guessing pricings always have a begins date? Otherwise this will result in a "None <= date" comparison, which will error.

api/graphql/types/reservation_unit/serializers.py
```diff
@@ -206,34 +202,52 @@ def validate_for_publish(self, data: dict[str, Any]) -> None:
...
+        if len(pricing_dates) != len(set(pricing_dates)):
+            msg = "Reservation unit can have only one pricing per date."
+            raise ValidationError(msg, code=error_codes.RESERVATION_UNIT_PRICINGS_DUPLICATE_DATE)
+
+        # Highest price cannot be less than the lowest price
+        for pricing in pricings:
+            highest_price = pricing.get("highest_price")
+            lowest_price = pricing.get("lowest_price")
+
+            if lowest_price is None and highest_price is None:
```
matti-lamppu, line 243: Is a pricing without any price valid?

api/graphql/types/reservation_unit/serializers.py
```diff
@@ -301,25 +315,13 @@ def update_hauki(instance: ReservationUnit) -> None:
...
-                reservation_unit=reservation_unit,
-                status=PricingStatus.PRICING_STATUS_ACTIVE,
-            ).delete()
-
-        if not ReservationUnitPricingHelper.contains_status(PricingStatus.PRICING_STATUS_FUTURE, pricings):
-            ReservationUnitPricing.objects.filter(
-                reservation_unit=reservation_unit,
-                status=PricingStatus.PRICING_STATUS_FUTURE,
-            ).delete()
+        pricing_pks = [pricing.get("pk") for pricing in pricings if "pk" in pricing]
```
matti-lamppu, line 318: Should probably add a transaction for the whole serializer save operation, so that this cannot succeed partially.

reservation_units/migrations/0105_refactor_pricing_state.py
```diff
@@ -0,0 +1,49 @@
...
+        .annotate(count=Count("begins"))
+        .filter(count__gt=1)
+    )
+    for duplicate in duplicates_qs:
+        # Delete all but the latest pricing with the same reservation_unit_id and begins, as it will be the active one
+        for pricing in ReservationUnitPricing.objects.filter(
+            reservation_unit_id=duplicate["reservation_unit_id"],
+            begins=duplicate["begins"],
+        ).order_by("-pk")[1:]:
+            pricing.delete()
```
matti-lamppu, line 22: I think this will work, but there is a more succinct way to express this using window functions:
```python
ReservationUnitPricing.objects.annotate(
    row=Window(
        RowNumber(),
        partition_by=("reservation_unit", "begins"),
        order_by="-pk",
    ),
).filter(row__gt=1).delete()
```

This will annotate a row number for each pricing so that row number starts from 1 for each unique reservation unit and begins date pair, starting from the last created. Therefore, we can just bulk remove anything that has a row number greater than 1, since they would be duplicate.

Should test this still if you're going to use it.

reservation_units/migrations/0105_refactor_pricing_state.py
```diff
@@ -0,0 +1,49 @@
...
+        ),
+        migrations.AddConstraint(
+            model_name="reservationunitpricing",
+            constraint=models.UniqueConstraint(
+                fields=("reservation_unit", "begins"),
+                name="reservation_unit_begin_date_unique_together",
+                violation_error_message="Pricing for this reservation unit already exists for this date.",
+            ),
+        ),
+    ]
```
matti-lamppu, line 49: Not 100% sure, but there might be a problem with removing duplicates AND adding the constraint in the same migration, IF there is actually a duplicate. I remember running into this at some point, and having to move them to different migrations, but I could be wrong.

matti-lamppu (changes requested):
Check comments

api/graphql/types/reservation_unit/serializers.py
ranta: Yep `begins` is required. Added a test for this.

api/graphql/types/reservation_unit/serializers.py
ranta: Good catch!

reservation_units/migrations/0105_refactor_pricing_state.py
ranta: Will separate them just in case. Tested this locally with some duplicates and this migration raised no issues

reservation_units/migrations/0105_refactor_pricing_state.py
ranta: Essentially the same thing, but this looks a lot cleaner :D

api/graphql/types/reservation_unit/serializers.py
matti-lamppu: Maybe a comment here would be nice?

reservation_units/migrations/0105_refactor_pricing_state.py
ranta: Too bad that that this doesn't work as-is due to database level restrictions:
```
django.db.utils.ProgrammingError: window functions are not allowed in WHERE
LINE 1: DELETE FROM "reservation_unit_pricing" WHERE ROW_NUMBER() OV...
```
I decided to keep the original duplicate removal since it works fine and in the end doesn't really matter that much 😅

matti-lamppu (approved):

api/graphql/types/reservation_unit/serializers.py
```diff
@@ -299,28 +314,19 @@ def update_hauki(instance: ReservationUnit) -> None:
...
-            ReservationUnitPricing.objects.filter(
-                reservation_unit=reservation_unit,
-                status=PricingStatus.PRICING_STATUS_FUTURE,
+    def handle_pricings(pricings: list[dict[Any, Any]], reservation_unit: ReservationUnit) -> None:
+        with atomic():
+            # Delete pricings that are not in the payload
+            pricing_pks = [pricing.get("pk") for pricing in pricings if "pk" in pricing]
+            ReservationUnitPricing.objects.filter(reservation_unit=reservation_unit).exclude(
+                pk__in=pricing_pks
             ).delete()
```
vergama, line 323: Doesn't this now delete past (inactive) pricings too?

api/graphql/types/reservation/serializers/confirm_serializers.py
```diff
@@ -91,8 +90,8 @@ def validate(self, data, prefill_from_profile=False):
             payment_type = data.get("payment_type", "").upper()
             reservation_unit = self.instance.reservation_unit.first()
 
-            active_price = ReservationUnitPricingHelper.get_active_price(reservation_unit)
-            if active_price.pricing_type == PricingType.PAID or self.instance.price_net > 0:
+            active_pricing = reservation_unit.actions.get_active_pricing()
+            if active_pricing.highest_price > 0 or self.instance.price_net > 0:
```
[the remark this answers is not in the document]
ranta: There are also some more places where the `price_net` can be changed to `price`, but let's maybe do that later since https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1304 is not merged yet.

actions/reservation_unit.py
```diff
@@ -340,3 +340,18 @@ def is_valid_staff_start_interval(self, begin_time: datetime.time) -> bool:
...
+    def get_active_pricing(self, by_date: datetime.date | None = None) -> ReservationUnitPricing | None:
+        """Returns the active pricing for the reservation unit."""
+        today = local_date()
+        if by_date is None:
+            by_date = today
+
+        return (
+            self.reservation_unit.pricings.filter(
+                models.Q(begins__lte=by_date, is_activated_on_begins=False)
+                | models.Q(begins__lte=today, is_activated_on_begins=True)
```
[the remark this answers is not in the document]
ranta: Yep, will simplify

api/graphql/types/reservation_unit/serializers.py
ranta: Yes, if they are not included in the pricings sent by the frontend.
Should I add a check here to prevent the deletion of any past pricings?

reservation_units/tasks.py
```diff
@@ -58,11 +47,11 @@ def update_reservation_unit_pricings_tax_percentage(
     # Last pricing for each reservation unit before the change date
     latest_pricings = (
         ReservationUnitPricing.objects.filter(
-            Q(begins__lte=change_date, pricing_type=PricingType.FREE)  # Ignore FREE pricings after the change date
-            | Q(pricing_type=PricingType.PAID)
+            Q(begins__lte=change_date, highest_price=0)  # Ignore FREE pricings after the change date
+            | Q(highest_price__gt=0)
         )
-        .filter(status__in=(PricingStatus.PRICING_STATUS_ACTIVE, PricingStatus.PRICING_STATUS_FUTURE))
         .exclude(reservation_unit__payment_accounting__company_code__in=ignored_company_codes)
```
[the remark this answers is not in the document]
ranta: ?

api/graphql/types/reservation_unit/serializers.py
vergama: That would be better since the frontend doesn't support transparent editing of the pricings and it is desirably to ensure that the inactive pricings are kept 'for the record'.

reservation_units/tasks.py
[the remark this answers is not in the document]
vergama: Commented on the wrong line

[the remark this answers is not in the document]
ranta: Added a check and comment for this

matti-lamppu (approved):

PR #1323, opened 2024-09-09
Add support for Keycloak authentication

matti-lamppu:
## 🛠️ Changelog

- Replaces Tunnistamo login system with Keycloak one.
- Migrates existing Tunnistamo user data to the new Keycloak user
- Adds support for logout redirect urls
- Logout all users when this change goes live via a migration
- Rewrite the Tunnistamo client as Keycloak client and use it in the Helsinki Profile client.

## 🧪 Test plan

- Automated tests
- Manual tests

## 🚧 Dependencies

- To support new logout redirect, the frontend must include a `redirect_to` parameter in the logout post request and redirect to one of the [configured logout endpoints](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/8981610507/Palveluintegraatio+HP+-+Tilavarauspalvelu+eli+uusi+Varaamo) (OIDC post logout redirect URIs) based on the UI. https://github.com/City-of-Helsinki/tilavarauspalvelu-ui/pull/1420
- Must reconfigure settings in Azure -> OpenShift when merging!!!

## 🎫 Tickets

- [TILA-2805](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2805)

[TILA-2805]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


users/helauth/pipeline.py
```diff
@@ -102,3 +102,61 @@ def update_user_from_profile(request: WSGIRequest, *, user: User | None = None)
...
+    request: WSGIRequest,
+    response: OIDCResponse,
+    user: User | None = None,
+    **kwargs: Unpack[ExtraKwargs],
+) -> dict[str, Any]:
+    if user is None:
+        return {"user": user}
+
+    id_token = IDToken.from_string(response["id_token"])
+    if id_token is not None and id_token.is_ad_login and id_token.iss.endswith("helsinki-tunnistus"):
```
vergama, line 118: Should perhaps further limit this to `@hel.fi` and `@*.hel.fi` emails (there's at least `@edu.hel.fi`), which would also prevent it being applied to the 70 or so AD users with an empty email (at least in the Tunnistamo token, not sure about Keycloak) -- in production, they seem to consist of persons with Sarastia in AD groups -- plus Eemeli -- that then need to be manually migrated for roles (and possibly reservations, if they had the Reserver role).

vergama: Although come to think of it, in test env. we may want to test with AD guest users (with nevertheless non-empty email and if so best to enforce the email domains that can be migrated / default to none, instead of any, if not specifically configured for the environment).

users/migrations/0016_clear_sessions.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.0.8 on 2024-08-26 22:44
+import os
+
+from django.contrib.sessions.backends.cache import KEY_PREFIX
+from django.core.cache import cache
+from django.db import migrations
+
+
+def clear_sessions(apps, schema_editor):
+    if os.getenv("DJANGO_SETTINGS_ENVIRONMENT") not in ["AutomatedTests", "CI"]:
```
matti-lamppu, line 10: Should not be skipped in the CI Environment since migrations are run in Platta using that environment.

api/helauth/views.py
```diff
@@ -0,0 +1,38 @@
...
+        end_session_url = redirect_to
+    else:
+        end_session_url = update_query_params(end_session_url, post_logout_redirect_uri=redirect_to)
+
+    logout(request)
+
+    if was_authenticated:
+        messages.success(request, _("You have been successfully logged out."))
+
+    return HttpResponseRedirect(end_session_url)
```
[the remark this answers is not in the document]
matti-lamppu: Not a real issue.

users/helauth/pipeline.py
matti-lamppu: I'll add checks for invalid/empty emails.

`id_token.is_ad_login` now validates that the authentication methods reference is `helsinki_adfs`, `helsinkiad`, or `helsinkiazuread`. Isn't this enough to determine that the migration should be attempted? Or is there another `amr` for the `edu` people?

users/migrations/0016_clear_sessions.py
matti-lamppu: Fixed.

users/helauth/pipeline.py
vergama: There is currently one production AD user with an **@edu.hel.fi** email who has a staff role (unit _Reserver_).

(In theory it is possible that there are other similar users created based on AD authentication via Tunnistamo but role has never been granted, or has been removed. But I suppose that is unlikely, unless a user has used the login method by mistake. It is also possible that some of the Kasko Sarastia AD users without email in Tunnistamo token are really edu people as well, but then those users cannot be migrated anyway.)

So at least the authentication method scenario with presumably new `amr` value coming from Keycloak indeed needs to be supported (if only to identify the login method as AD for other purposes; migrating a single user's roles manually would not be a big deal in itself). Currently with Tunnistamo the amr would be still be `'helsinki_adfs'`.

As for users who have used their **@edu.hel.fi** address for `suomi.fi` authentication and profile, they can apparently keep using those profiles (and contrary to previous information, create new profiles with such emails) as long as they don't wish to authenticate with (edu) AD, in which case the profile's email would need to be changed first (but we might not know about that change if they haven't logged in as that user afterwards).

users/helauth/pipeline.py
```diff
@@ -102,3 +102,66 @@ def update_user_from_profile(request: WSGIRequest, *, user: User | None = None)
...
+
+    return {"user": user}
+
+
+@SentryLogger.log_if_raises("Keycloak migration: Failed to migrate user from tunnistamo to keycloak")
+def migrate_from_tunnistamo_to_keycloak(*, email: str) -> None:
+    # Look at the two most recent AD users with the same email address (no profile ID).
+    # There can be more, but the most recent one is the new Keycloak user and the other one
+    # is the old Tunnistamo user after the auth-migration (when auth moved to backend), which we want to migrate.
+    users: list[User] = list(User.objects.filter(email=email, profile_id="").order_by("date_joined")[:2])
```
vergama, line 134: Haven't thought this entirely through yet, but `profile_id=""` might not be a sufficient check against some cases where the same email has been used for both profile (suomifi) and AD; i.e. the `old_user` maybe could be a profile one with this info unset (because no recent login)?
If so, then perhaps it can be weeded out by having no staff/superuser flags set nor roles assigned.

users/helauth/pipeline.py
matti-lamppu: > presumably new `amr` value coming from Keycloak indeed needs to be supported

Should be checked in staging and production, but at least on dev the `amr` is `helsinkiad` for my account, which is currently supported in `ADLoginAMR`.

matti-lamppu: > There is currently one production AD user with an @edu.hel.fi email who has a staff role (unit Reserver).

Added `eduad` as a possible `ADLoginAMR` value to support this users migration, however if it doesn't work then they can be manually migrated since there is just one.

users/helauth/pipeline.py
```diff
@@ -102,3 +102,66 @@ def update_user_from_profile(request: WSGIRequest, *, user: User | None = None)
...
+    if len(users) < 2:
+        return
+
+    old_user = users[0]
+    new_user = users[1]
+
+    # Don't run migration more than once.
+    if not old_user.is_active:
```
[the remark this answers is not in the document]
matti-lamppu: I don't believe this is necessary. Duplicate AD users should be from the time of the previous migration (3 clients to centralized to backend), in which case the penultimate user (more recent `date_joined`) is from after the migration (the latest user being the new Keycloack user). This user is only created if the user has logged in since the migration. Older users probably aren't using the system anymore.

users/helauth/pipeline.py
matti-lamppu: If the same email has been used for both `AD` and `Suomi.fi` logins, these would have created two different users in our system. See [this `AD` user](https://tilavaraus.hel.fi/admin/users/user/3852/change/) vs. [this `Suomi.fi` user](https://tilavaraus.hel.fi/admin/users/user/5866/change/).  Furthermore, our logging middleware makes sure that `profile_id` is filled for all profile users on each login.

users/helauth/pipeline.py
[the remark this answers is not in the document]
vergama: That is generally but not entirely uniformly true, as there are some duplicate AD users with the same email address:
- (A) both identities have been created _subsequent_ to the centralization -- in fact the newest such incarnation appears to have been created last week for an AD user who had previously logged in around March 2024 (with another identity) -- however, examination proves that they are not an issue, as the user incarnation with more recent `date_joined` value is clearly valid;
- (B) or both identities have been created _prior_ to that date (this is the rarer case), and it turns out that for few of them the more recent identity is now a bogus one. (This can be found out in more ways than one.)

**TL;DR**: the present check `if not old_user.is_active` should nevertheless be sufficient, together with the filtering out of non-empty profile ids further above. I went through the three dozen or so such user pairs in production environment (same hel.fi email, both still flagged active, profile id empty in both) and removed the active flag from one of them.

**Longer version**:
Tunnistamo token doesn't contain AD username but the difference in AD groups seems to indicate that for set **A**, person's role is now different and maybe these are cases where an external AD account has been recreated; although it doesn't really matter what the ultimate reason is.
Of set **B**, most cases (6/9) were explained by the more recent  `date_joined` (thus with greater `user.id`) having been used strictly for Django-admin login before the centralization, but few (3/9) could not be readily explained since the other version had never logged in subsequent to that previous migration. It can be speculated that either there had been a Tunnistamo malfunction and it had given out different sub claim / mis-identity for a time (this kind of thing has happened) or suomifi-login had been performed by the user once (by mistake). Whatever the situation, the bogus versions are no longer flagged as active, and will be skipped accordingly.

users/helauth/pipeline.py
vergama: Exactly, it creates two users, so then the question is, whether empty profile id is a fool-proof check to ignore the putative profile variants (because it wasn't always saved). Examination of already existing user pairs sharing the same `hel.fi` email turned up just a handful of cases where this wasn't clear, but after inactivating their dubious incarnations (where `date_joined` was more recent), this should be okay. (See further in other comment.)

users/helauth/pipeline.py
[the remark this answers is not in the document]
matti-lamppu: If there are any cases where there are currently two users, and the most recently created one is not the one that should be migrated, then this won't work for those users, since we aren't filtering out inactive users. Currently this just takes the two most recent users, and if the most recent one is inactive, then we just don't perform the migration even once. The new user doesn't get their permissions from the actual user, which is was the one before the inactive one. Did I understand correctly that there were some like that?

users/helauth/pipeline.py
matti-lamppu: The profile ID filling has been there for at least last September, so I think that all users that still use the system should have that set.

ranta (approved):

tests/test_gdpr_api/helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    # Ensure values are strings and not bytes
+    for name in ["n", "e"]:
+        value = key.public_key[name]
+        if isinstance(value, bytes):
+            key.public_key[name] = value.decode("utf-8")
+
+    return key
+
+
+RSA_PRIVATE_KEY = """-----BEGIN PRIVATE KEY-----
```
vergama, line 78: Could rename this to something like `RSA_PRIVATE_KEY_FOR_TESTING` so it is less likely to cause a false positive for security issue (private key committed to repo) when looked at disregarding the surrounding context of unit tests.

users/helauth/pipeline.py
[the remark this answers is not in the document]
vergama: With the `is_active` check now in the user query this should be fine.
In practical terms existing production environment duplicates of users with an `hel.fi` email have been set as inactive in all cases as revealed by the following SQL (if either one already was not):
`select u1.email, u2.id as newer_id, u1.id as older_id, u2.is_active as newer_active, u1.is_active as older_active, u2.username as newer_username, u1.username as older_username, u2.date_joined as newer_joined, u2.last_login as newer_last_login, u1.date_joined as older_joined, u1.last_login as older_last_login from public.user u1, public.user u2 where u1.id < u2.id and u1.email = u2.email and u1.email like '%hel.fi' and u1.profile_id = '' and u2.profile_id = '' order by u2.id desc, u1.id desc;`
In most cases the earlier version of the user (which dated to time before the authentication centralization) is in any case inactive; but in few cases it is the newer user that has been set as inactive, either because it had only been used for Django login before that date or was some other bogus variant which had never logged since that date.

**So**: if the migration query now still finds more than two active users ('new', 'old', plus 'older'), it would be because someone has continued using the system or otherwise has different sub-claim from Tunnistamo in later September (possible, there's one such case since yesterday) -- and that should be okay, since in those cases `order by date_joined desc` can be assumed to be fetch the two valid ones.

vergama (approved):

PR #1385, opened 2024-10-11
Fix email templates & ical file

matti-lamppu:
## 🛠️ Changelog

- Fixes email templates containing superfluous `%}` sings
- Removes unused email templates
- Fixes an issue for a like in staff email templates
- Adds timezone information to ical files
- Translate ical files according to user language
- Compile Swedish translations in docker container
- Removes some erroneous Swedish translations
- Example ical file:  [reservation_calendar.ics.txt](https://github.com/user-attachments/files/17340378/reservation_calendar.ics.txt)

## 🧪 Test plan

- Automated tests
- Manual tests:
    - Check ical file in Outlook (my licence expired...)
    - Check email contents (automated tests coming later)

## 🚧 Dependencies

- None

## 🎫 Tickets

- None


tilavarauspalvelu/models/reservation/actions.py
```diff
@@ -43,14 +43,57 @@ def get_actual_after_buffer(self) -> datetime.timedelta:
...
+        cal.add_component(self._get_ical_event(language=language))
+
+        return cal.to_ical()
+
+    def _get_ical_timezone(self) -> Timezone:
+        """Adds timezone information to the ical calendar event."""
+        timezone = Timezone()
+        timezone.add(TimezoneProperty.TZID, settings.TIME_ZONE)
+
+        # Taken from outlook generated iCal files
```
vergama, line 66: Could extend the comment that these assumptions are valid for timezone which observe European Summer Time as currently defined (EU directive 2000/84/EC).
If in the future EU eventually implements the proposal to abolish DST, then an UNTIL definition needs to be added to the RRULEs (so that the icals will still be generated properly across the transition period).

tilavarauspalvelu/models/reservation/actions.py
```diff
@@ -43,14 +43,57 @@ def get_actual_after_buffer(self) -> datetime.timedelta:
...
+        daylight = TimezoneDaylight()
+        daylight.add(name=TimezoneRuleProperty.DTSTART, value=daylight_start)
+        daylight.add(
+            name=TimezoneRuleProperty.RRULE,
+            value={"FREQ": "YEARLY", "BYDAY": "-1SU", "BYMONTH": 3},
+        )
+        daylight.add(name=TimezoneRuleProperty.TZOFFSETFROM, value=datetime.timedelta(hours=2))
+        daylight.add(name=TimezoneRuleProperty.TZOFFSETTO, value=datetime.timedelta(hours=3))
+
+        timezone.add_component(standard)
```
vergama, line 88: I guess strictly speaking these components should only be added if TZID is "Europe/*" -- just in case at some future point the ical files are being generated in UTC but nobody touches the VTimezone part which would then be inappropriate.
(There are maybe ways to programmatically deduce if a given timezone observes DST and with which transition points but that seems like over-complicating things... inasmuch as in practice it will hardly be a non-European timezone, and even if it were, the DST rules might be different anyhow; would need a library and extensive units tests in that case to handle the logic.)

vergama (approved):
I think this will work for practical purposes, at least for the time being. Comments could be slightly clarified the assumptions made for VTimezone.

tilavarauspalvelu/models/reservation/actions.py
matti-lamppu: The calendar files are always generated in the Helsinki timezone, even if the time saved to database would be in some other timezone (although it's always converted in UTC). I don't think we need to consider this, since this is just for Helsinki City.

tilavarauspalvelu/models/reservation/actions.py
matti-lamppu: Added clarifying comments to timezone use.

PR #1388, opened 2024-10-17
Improve email integration

matti-lamppu:
## 🛠️ Changelog

- Refactors the email integration (moved to `tilavarauspalvelu/integrations/email`)
- Separates responsibilities for rendering templates, finding recipients, determining email language, building template contexts, and sending the emails to make testing them simpler
- Improves email template rendering by making django template tags available in them
- Removes redundant `EmailTemplate` model, but retains the ability to view templates in the admin panel, as well as the email tester while making some improvements to them (i.e. viewing different language versions)
- Adds the ability to test translations based on `.po` files without having to compile the translations to `.mo` files
- Adds checks to CI to make sure translations are available in Swedish for defined contexts (currently ical and email)
- Adds better tests for email templates that make sure all templates render correctly, translations work correctly, and that the correct checks are in place when trying to send different kinds of emails

## 🧪 Test plan

- Automated tests
- Manual tests are required to check the email templates still work as they should
- Manual tests should check that emails are sent in general

## 🚧 Dependencies

- Should rename `SEND_RESERVATION_NOTIFICATION_EMAILS` to `SEND_EMAILS` in azure library

## 🎫 Tickets

- [TILA-3598](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3598) [this PR is a requisite for this ticket]
- [TILA-3601](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3601)

[TILA-3598]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/integrations/email/main.py
```diff
@@ -0,0 +1,506 @@
...
+            text_content=render_text(email_type=email_type, context=context),
+            html_content=render_html(email_type=email_type, context=context),
+        )
+
+    @staticmethod
+    def send_application_in_allocation_emails() -> None:
+        """Sends emails to applicants when application round has entered the allocation phase."""
+        applications = Application.objects.should_send_in_allocation_email().order_by("created_date")
+        if not applications:
+            # TODO: Should we log this?
```
vergama, line 89: If this is run as a periodic task, then the count will be zero most of the time.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -0,0 +1,506 @@
...
+
+        send_multiple_emails_in_batches_task.delay(emails=emails)
+        applications.update(in_allocation_notification_sent_date=local_datetime())
+
+    @staticmethod
+    def send_application_handled_emails() -> None:
+        """Sends email to applicants when application round has its allocation results available."""
+        applications = Application.objects.should_send_handled_email().order_by("created_date")
+        if not applications:
+            # TODO: Should we log this?
```
vergama, line 125: Good to log as this would be unexpected.

tilavarauspalvelu/integrations/email/template_context/reservation.py
```diff
@@ -0,0 +1,681 @@
...
+    *,
+    language: Lang,
+    **data: Any,
+) -> EmailContext:
+    if reservation is not None:
+        begin_datetime = reservation.begin.astimezone(DEFAULT_TIMEZONE)
+        end_datetime = reservation.end.astimezone(DEFAULT_TIMEZONE)
+        data: dict[str, Any] = {
+            "email_recipient_name": reservation.actions.get_email_reservee_name(),
+            "subsidised_price": reservation.actions.calculate_full_price(begin_datetime, end_datetime, subsidised=True),
```
vergama, line 295: The old version works (a bit obscurely) so that `subsidised_price` will be different to `reservation.price` only if discount has been applied for: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/175161b2419810643679c430dd99b6f6f961f0c9/tilavarauspalvelu/utils/email/email_builder_reservation.py#L209
That is because if it is has _not_ been applied for, then the final price _should not_ be changed from the default highest price in the handling (even if it theoretically could be changed by a handler).

The condition (`if subsidised_price < price`) in the `reservation_price_range` template partials has the benefit, though, that if there is no range defined in the reservation unit's pricing (but the form has allowed applying for a discount and the user has), then it won't print out a bogus price range such as 15,00 - 15,00 € -- unlikely as that combination of data perhaps is.

Even so, maybe the clearest solution (in this specific type of email context) would be to always calculate `subsidised_price` but in the `reservation_price_range` templates rewrite the condition as e.g. `if price_can_be_subsidised`, while price_can_be_subsidised would be here set to true if `reservation.applying_for_free_of_charge & subsidised_price <price`.

vergama (changes requested):
Looks very good, just one doubt about the detail of the price range in the emails for reservation which requires handling.

tilavarauspalvelu/integrations/email/main.py
matti-lamppu: Removed.

tilavarauspalvelu/integrations/email/template_context/reservation.py
matti-lamppu: Modified the logic to suit this and added tests

vergama (approved):
Clever stuff with html2text, I thought the outlook compatible html might be bad enough that tests would need to compare with the pre mjml-2-html stage.

PR #1390, opened 2024-10-17
Anonymize old users

matti-lamppu:
## 🛠️ Changelog

- Anonymize users that have not logged in for a period defined by `ANONYMIZE_USER_IF_LAST_LOGIN_IS_OLDER_THAN_DAYS`
- Prevent anonymization similarly to gdpr endpoint if there are open reservations, applications, or payments

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3590](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3590)

[TILA-3590]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/models/user/actions.py
```diff
@@ -1,11 +1,164 @@
...
+
+        GeneralRole.objects.filter(user=self.user).delete()
+        UnitRole.objects.filter(user=self.user).delete()
+        UserSocialAuth.objects.filter(user=self.user).delete()
+
+    def anonymize_user_reservations(self) -> None:
+        reservations = Reservation.objects.filter(user=self.user).exclude(
+            type__in=[ReservationTypeChoice.BLOCKED, ReservationTypeChoice.STAFF],
+        )
+        for reservation in reservations:
```
vergama, line 73: Strictly speaking not part of this ticket spec. but if these (filtered) reservations belong to a recurring reservation, then the name & description strings of those parents should also be anonymized. (Not previously thought of since `SEASONAL` reservations did not exist.)

tilavarauspalvelu/models/user/actions.py
```diff
@@ -1,11 +1,164 @@
...
+            )
+            .exists()
+        )
+
+        has_open_applications = (
+            self.user.applications.all()
+            .has_status_in(
+                [
+                    ApplicationStatusChoice.RECEIVED.value,
+                    ApplicationStatusChoice.IN_ALLOCATION.value,
```
vergama, line 148: Come to think of it, `HANDLED` can be added here. Anonymization can well wait until any rounds are in the final state.
(Should the application contain allocations, then the reservation check alone would prevent its anonymization, but if all sections were rejected then it would not -- and it might be confusing handlers to encounter anonymized applications while the application round results have not yet been sent to applicants; even if their sections have been rejected and therefore need not be allocated anything even if the round would be reset in the previous state. It will also normally ensure that a `sent_date` has already been set on anonymized applications.)

tilavarauspalvelu/models/user/queryset.py
```diff
@@ -53,5 +54,18 @@ def should_deactivate_permissions(self) -> models.QuerySet:
...
         )
 
+    def anonymize_inactive_users(self) -> None:
+        cutoff = local_date() - datetime.timedelta(days=settings.ANONYMIZE_USER_IF_LAST_LOGIN_IS_OLDER_THAN_DAYS)
+
+        # All users who haven't logged in in a while but haven't been anonymized yet
+        users = self.exclude(
+            first_name=ANONYMIZED_FIRST_NAME,
+            last_name=ANONYMIZED_LAST_NAME,
+        ).filter(last_login__date__lt=cutoff)
```
vergama, line 64: I realised that in production there are many users who don't have `last_login` set (because they precede the authentication change) but are less than 2 years, or even just about less than 1 year old... so need to also check that `user.date_joined < cutoff` lest they be prematurely anonymized.

vergama (changes requested):
Couple of details not come to mind but in general should work.

tilavarauspalvelu/models/user/actions.py
matti-lamppu: Added anonymization for these, and made existing anonymization more performant by using `bulk_update`.

tilavarauspalvelu/models/user/queryset.py
matti-lamppu: Corrected the filter.

vergama (approved):

PR #1433, opened 2024-11-14
Development improvements

ranta:
## 🛠️ Changelog

- Improve env variable setup for developers
- Downgrade `django-debug-toolbar` and lock its version due to `django-graphiql-debug-toolbar` breakage

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- None


local_settings_example.py
```diff
@@ -7,13 +7,24 @@
 class LocalMixin:
     """Add custom local settings here."""
 
-    SOCIAL_AUTH_TUNNISTAMO_ALLOWED_REDIRECT_HOSTS = ["localhost:3000", "localhost:3001"]
```
matti-lamppu, line 10: Did you test with the frontend if these can be removed? I think there might be some issues with redirects in local setups since the frontends are technically in a different domain (not true in Platta)

local_settings_example.py
```diff
@@ -7,13 +7,24 @@
...
+    # On macOS, you might need to set these paths manually (https://gis.stackexchange.com/a/416882)
+    # GDAL_LIBRARY_PATH = "/opt/homebrew/lib/libgdal.dylib"
+    # GEOS_LIBRARY_PATH = "/opt/homebrew/lib/libgeos_c.dylib"
+
+    # CELERY_ENABLED = False  # If False, run tasks synchronously without a worker, but scheduled tasks will run.
     CELERY_TASK_ALWAYS_EAGER = True
 
+    GRAPHQL_CODEGEN_ENABLED = False  # Set to True to enable codegen for frontend by disabling CSRF protection
```
matti-lamppu, line 17: Frontend devs don't really use Local interpreters, so this should preferably be in `DockerMixin`.

matti-lamppu (approved):

local_settings_example.py
ranta: I saw this commit, and thought that it should not be used anymore: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/commit/40520c6da67c4883739f6f5bc501baac6c466da8

But after testing locally, when logging in I get redirected to `/admin`, instead of back to the frontend. Even though nothing with this string is found in search, it still is used deep in the `social_core` package 😅

PR #1449, opened 2024-11-25
Update Ruff and configuration

ranta:
## 🛠️ Changelog

- See commits for details

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-2836](https://helsinkisolutionoffice.atlassian.net/browse/TILA-2836)

[TILA-2836]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-2836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/api/graphql/types/reservation/serializers/update_serializers.py
```diff
@@ -40,7 +41,7 @@ def validate(self, data, prefill_from_profile=False):
             msg = f"Setting the reservation state to '{getattr(new_state, 'value', new_state)}' is not allowed."
             raise ValidationErrorWithCode(msg, ValidationErrorCodes.STATE_CHANGE_NOT_ALLOWED)
 
-        data = super().validate(data, prefill_from_profile)
+        data = super().validate(data)
```
matti-lamppu, line 48: I think the behavior would change here, since `ReservationUpdateSerializer` inherits from `ReservationCreateSerializer`, which has `prefill_from_profile=True` by default, while here we would pass `False`. This would cause the reserve info to change to the profile info on update, which could be different if the user changed it before reservation confirmation.

Interestingly, `prefill_from_profile=False` was also included in the `ReservationConfirmSerializer`, but not passed down to to the superclass. Maybe this does work accidentally, since most handlers don't have permissions to read other user's profile info, so the prefill wouldn't override anything.

In any case, to fix this we would need to remove the inheritance between the two serializer and migrate the serializer methods to actions for resuability. This is a bit of work though, so maybe you can add some smaller fix before that.

pyproject.toml
```diff
@@ -204,22 +204,25 @@ lint.dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$|^args$|^kwargs$
 "local_settings*.py" = [
     "ERA001",   # Commented code is fine here.
 ]
+"utils/query_performance.py" = [
```
matti-lamppu, line 231: I think this doesn't exist anymore

tests/factories/recurring_reservation.py
```diff
@@ -69,7 +69,7 @@ def create_with_matching_reservations(cls, **kwargs: Any) -> RecurringReservatio
         space = SpaceFactory.create(unit=series.reservation_unit.unit)
         series.reservation_unit.spaces.add(space)
 
-        weekdays: list[int] = [int(val) for val in series.weekdays.split(",") if val != ""]
+        weekdays: list[int] = [int(val) for val in series.weekdays.split(",") if val]
```
matti-lamppu, line 72: This is wrong, it skips Mondays (value=0) since 0 is falsy.

tilavarauspalvelu/management/commands/data_creation/create_reservations.py
```diff
@@ -760,7 +760,7 @@ def _create_reservations_for_series(
     pricing: ReservationUnitPricing | None = series.reservation_unit.pricings.active().first()
     assert pricing is not None, "Reservation unit must have at least one pricing"
 
-    weekdays: list[int] = [int(val) for val in series.weekdays.split(",") if val != ""]
+    weekdays: list[int] = [int(val) for val in series.weekdays.split(",") if val]
```
matti-lamppu, line 763: This is wrong, it skips Mondays (value=0) since 0 is falsy.

tilavarauspalvelu/models/recurring_reservation/actions.py
```diff
@@ -163,7 +163,7 @@ def pre_calculate_slots(
         end_time: datetime.time = self.recurring_reservation.end_time
         reservation_unit = self.recurring_reservation.reservation_unit
 
-        weekdays: list[int] = [int(val) for val in self.recurring_reservation.weekdays.split(",") if val != ""]
+        weekdays: list[int] = [int(val) for val in self.recurring_reservation.weekdays.split(",") if val]
```
matti-lamppu, line 165: This is wrong, it skips Mondays (value=0) since 0 is falsy.

utils/sentry.py
```diff
@@ -17,7 +21,7 @@ def log_message(message: str, details: str | dict[str, Any] | None = None, level
             capture_message(message, level=level)
 
     @staticmethod
-    def log_exception(err: Exception, details: str, **extra) -> None:
+    def log_exception(err: Exception, details: str, **extra: dict[str, Any]) -> None:
```
matti-lamppu, line 24: `**kwargs` should be typed by their value, so `**extra: Any` for keyword arguments of `Any` type.

tests/factories/recurring_reservation.py
ranta: `0` is falsy, but `"0"` is truthy.
Since `series.weekdays` is always a string (we can't split an `int` anyway) this shouldn't affect the logic at all.

utils/sentry.py
ranta: Oopsie, nice catch 😄

tests/factories/recurring_reservation.py
matti-lamppu: Ah, that's right, my mistake.

tilavarauspalvelu/api/graphql/types/reservation/serializers/update_serializers.py
ranta: Excellent point!
Resolved this by creating a base class for these two serializers, and separating the prefill logic to only the create serializer.

matti-lamppu (approved):

tilavarauspalvelu/api/graphql/types/reservation/serializers/update_serializers.py
matti-lamppu: Yep, this will do.

matti-lamppu (approved):
LGTM 🚀

matti-lamppu:
I'm merging this just so that I can start working on top of it. Open another PR if you need to add more changed.

PR #1452, opened 2024-11-27
Remove deprecated features

matti-lamppu:
## 🛠️ Changelog

- Removes some deprecated models:
    - Service Sector
    - Service
    - Real Estate
    - Building
    - Introduction
    - Keyword
    - Keyword Group
    - Keyword Category
- Removes some deprecated fields on models
    - Resource buffer times
    - ReservationUnitCancellationRule needs handling boolean
- Fixed one nullable CharField issue
- Add `ext_id` UUID fields to reservations and recurring reservations for "keyless entry"
- Make translation check check for removed translations as well & fix printing

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Need to run frontend GraphQL generation https://github.com/City-of-Helsinki/tilavarauspalvelu-ui/pull/1551

## 🎫 Tickets

- [TILA-3695](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3695)

[TILA-3695]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/models/recurring_reservation/model.py
```diff
@@ -27,9 +27,9 @@
 
 
 class RecurringReservation(models.Model):
+    ext_id: uuid.UUID = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)  # ID for external systems
```
ranta, line 30: I think `public_uuid` (or just `uuid` or something similar) would fit better to reflect that it's an uuid field, not a "normal" id field, and that the uuid does not come from an external source, but rather that it's shared to the public

tilavarauspalvelu/migrations/0043_populate_uuid_values.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.db import migrations
+
+
+def gen_uuid(apps, schema_editor) -> None:
+    Reservation = apps.get_model("tilavarauspalvelu", "Reservation")
+    rows = list(Reservation.objects.all())
+    for row in rows:
+        row.ext_id = uuid.uuid4()
+    Reservation.objects.bulk_update(rows, ["ext_id"])
```
ranta, line 14: This migration should be unnecessary, due to `models.UUIDField(default=uuid.uuid4, ...)` already handling it

ranta (approved):

tilavarauspalvelu/migrations/0043_populate_uuid_values.py
matti-lamppu: This is straight from [Django's own instructions](https://docs.djangoproject.com/en/5.1/howto/writing-migrations/#migrations-that-add-unique-fields) on how to add unique fields. The middle step is required to make sure that new values are actually unique.

tilavarauspalvelu/models/recurring_reservation/model.py
matti-lamppu: `uuid` creates too many name collisions by itself (`import uuid` -> type field as `uuid.UUID`).  I think `ext_id` is fine since it means "external ID" = "ID used for external systems". `public_uuid` IMO has the same possibility for misunderstanding where the id comes from, but I did add a comment for that purpose (`# ID for external systems`)

matti-lamppu: And it's not like our regular IDs aren't "public", they just aren't as reliable for external systems to query with since they might change e.g. if the database is recreated.

vergama (approved):

tilavarauspalvelu/models/recurring_reservation/model.py
matti-lamppu: Changed the name to `ext_uuid` to reflect that it is a UUID.

PR #1462, opened 2024-12-11
Add "require adult reservee" checks

matti-lamppu:
## 🛠️ Changelog

- Add an option for requiring adult reservees for some reservation units by setting an option on the reservation unit
- Check user age in (tentative) reservation create endpoint if this setting is on
- Also check age when creating applications
- Allow AD users to bypass these checks, since we cannot check their birthdays. We have banned students from signing in via AD accounts, so all AD users should be of age.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3709](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3709)

[TILA-3709]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vergama (approved):

tilavarauspalvelu/models/user/actions.py
```diff
@@ -187,3 +188,27 @@ def can_anonymize(self) -> UserAnonymizationInfo:
...
+        if birthday is None:
+            return None
+        return relativedelta(local_date(), birthday).years
+
+    @property
+    def is_of_age(self) -> bool:
+        user_age = self.user_age
+        if user_age is None:
+            return False
+        return user_age >= settings.USER_IS_ADULT_AT_AGE
```
ranta, line 214: My intuition is saying that these would fit better directly to the `User` model, since they are properties and don't use data from any other source than itself 🤔 
(No need to make any changes if you prefer them here)

ranta (approved):

tilavarauspalvelu/models/user/actions.py
matti-lamppu: Maybe, but I didn't want to bloat the user model any more than it already is.

PR #1468, opened 2024-12-12
Seasonal Reservation Emails

ranta:
## 🛠️ Changelog

- Add a bunch of new emails when seasonal reservations/application sections/reservation series get modified/cancelled/rejected

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3668](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3668)

[TILA-3668]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/admin/email_template/forms.py
```diff
@@ -161,30 +189,30 @@ class ReservationBaseForm(
 
 class ApplicationHandledEmailTemplateTesterForm(BaseEmailTemplateForm):
     def to_context(self) -> EmailContext:
-        return get_context_for_application_handled(language=self.cleaned_data["language"])
+        return get_context_for_application_handled(**super().get_context_params())
```
matti-lamppu, line 213: You should just call `self.get_context_params()` in all of these implementations since you might want to modify the context params in the class itself, and this would now not enable that. Generally, you'd only want to use `super()` to call a the same method of a superclass, not any other method.

tilavarauspalvelu/integrations/email/template_context/application.py
```diff
@@ -1,28 +1,41 @@
...
 
 __all__ = [
     "get_context_for_application_handled",
     "get_context_for_application_in_allocation",
     "get_context_for_application_received",
+    "get_context_for_application_section_cancelled",
 ]
 
 
+# type: EmailType.APPLICATION_HANDLED ##################################################################################
```
matti-lamppu, line 36: `# type:` is a special syntax for comments used by python type checkers to add information about typing (see [PEP484](https://peps.python.org/pep-0484/), e.g. `# type: ignore` for indicating that type checkers should ignore type errors for some code segment). Therefore I would use some other keyword here, e.g. `# email type: APPLICATION_HANDLED`

tilavarauspalvelu/integrations/email/sending.py
```diff
@@ -1,67 +1,52 @@
...
-    """
-    Sends an email message to in batches.
-
-    :param recipients: The recipients of the email.
-    :param subject: The subject of the email.
-    :param text_content: The text content of the email.
-    :param html_content: The HTML content of the email.
-    :param attachments: An attachments in the email.
-    """
+def send_emails_in_batches_task(email_data: EmailData | dict[str, Any]) -> None:
```
matti-lamppu, line 25: I would remove `dict[str, Any]` from the typing so that when the function is used, typing would indicate that we should use the EmailData type instead of any dict being acceptable. I would then flip the guard clause to convert the celery-serialized dict to an `EmailData` instance, with a comment to explain it.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -80,81 +137,64 @@ def send_application_received_email(
...
+        email = EmailData.build(recipients, context, email_type)
+        send_emails_in_batches_task.delay(email_data=email)
 
     @staticmethod
-    def send_application_handled_emails() -> None:
-        """Sends email to applicants when application round has its allocation results available."""
-        applications = Application.objects.should_send_handled_email().order_by("created_date")
-        if not applications:
-            SentryLogger.log_message("Zero applications require the 'application handled' email to be sent")
+    def send_staff_notification_application_section_cancelled(application_section: ApplicationSection) -> None:
```
matti-lamppu, line 171: You added a separate section for staff emails, should this be there instead?

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -80,81 +137,64 @@ def send_application_received_email(
...
+        language: Lang | None = None,
+    ) -> None:
+        """Sends an email that the whole application section was cancelled by the user"""
+        reservation: Reservation | None = application_section.actions.get_last_reservation()
+        if reservation is None:
             return
 
-        recipients_by_language = get_recipients_for_applications_by_language(applications)
-        if not recipients_by_language:
+        recipients = get_reservation_email_recipients(reservation=reservation)
```
matti-lamppu, line 154: Should this fetch recipients from the application (`get_application_email_recipients(section.application)`) instead? I think they will be the same with how the series generation works out, but I think that would be more in line with the requirements.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -80,81 +137,64 @@ def send_application_received_email(
...
-        """Sends emails to applicants when application round has entered the allocation phase."""
-        applications = Application.objects.should_send_in_allocation_email().order_by("created_date")
-        if not applications:
+    def send_application_section_cancelled(
+        application_section: ApplicationSection,
+        *,
+        language: Lang | None = None,
+    ) -> None:
+        """Sends an email that the whole application section was cancelled by the user"""
+        reservation: Reservation | None = application_section.actions.get_last_reservation()
```
matti-lamppu, line 150: Should maybe have a check that the application status is `RESULTS_SENT`?

tilavarauspalvelu/integrations/email/main.py
matti-lamppu: Also applies to the staff email.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -445,56 +479,74 @@ def send_reservation_requires_payment_email(
...
-        :param language: The language of the email. Determine from reservation if not given.
-        """
-        # Prevent accidental sending of wrong email.
-        if reservation.state != ReservationStateChoice.DENIED:
+        reservation: Reservation | None = reservation_series.reservations.last()
+        if reservation is None:
+            return
+        if reservation.type != ReservationTypeChoice.SEASONAL:
+            return
+        if reservation_series.actions.get_application_section() is None:
```
matti-lamppu, line 500: I think this is an unnecessary check since you check `if reservation_series.allocated_time_slot is None` above, and the rest of the relations up to application section are not nullable.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -445,56 +479,74 @@ def send_reservation_requires_payment_email(
...
+        reservation_series: RecurringReservation,
+        *,
+        language: Lang | None = None,
+    ) -> None:
+        reservation: Reservation | None = reservation_series.reservations.last()
+        if reservation is None:
+            return
+        if reservation.type != ReservationTypeChoice.SEASONAL:
+            return
+        if reservation_series.actions.get_application_section() is None:
```
matti-lamppu, line 530: I would prefer this check that you have in `send_seasonal_reservation_modified_series_email`:

```python
if reservation_series.allocated_time_slot is None:
    return
```

tilavarauspalvelu/integrations/email/main.py
matti-lamppu: Applies to email contexts too.

tilavarauspalvelu/integrations/email/template_context/application.py
```diff
@@ -80,3 +99,136 @@ def get_context_for_application_received(*, language: Lang) -> EmailContext:
...
+                "weekday": WeekdayChoice.MONDAY.label,
+                "time": "13:00-15:00",
+                "url": get_staff_reservations_ext_link(reservation_id=1234),
+            },
+            {
+                "weekday": WeekdayChoice.TUESDAY.label,
+                "time": "21:00-22:00",
+                "url": get_staff_reservations_ext_link(reservation_id=5678),
+            },
+        ]
```
matti-lamppu, line 218: Tester data? Put it to the tester form method instead.

tilavarauspalvelu/admin/email_template/forms.py
```diff
@@ -162,384 +149,432 @@ class PendingInstructionsFormMixin(forms.Form):
...
+        )
+
+
+class StaffNotificationApplicationSectionCancelledTemplateTesterForm(EmailRecipientFormMixin, BaseEmailTemplateForm):
+    application_section_name = forms.CharField(initial="[HAKEMUKSEN OSAN NIMI]")
+    application_round_name = forms.CharField(initial="[KAUSIVARAUSKIERROKSEN NIMI]")
+    cancel_reason = forms.CharField(initial="[PERUUTUKSEN SYY]", widget=text_widget)
+
+    def to_context(self) -> EmailContext:
+        return get_context_for_application_section_cancelled(
```
matti-lamppu, line 574: Wrong method, should use `get_context_for_staff_notification_application_section_cancelled`.

tilavarauspalvelu/integrations/email/template_context/reservation.py
```diff
@@ -447,26 +704,82 @@ def get_context_for_reservation_requires_payment(
...
+            "email_recipient_name": reservation_series.actions.get_email_reservee_name(),
+            "rejection_reason": get_attr_by_language(reservation.deny_reason, "reason", language),
+            **params_for_reservation_series_info(reservation_series=reservation_series),
+            **params_for_application_section_info(application_section=section, language=language),
+        }
+
+    return {
+        "title": pgettext("Email", "Your seasonal booking has been cancelled"),
+        "text_reservation_rejected": pgettext(
+            "Email", "All space reservations included in your seasonal booking have been cancelled"
```
vergama, line 755: This seems to be different than in the defined email templates:
- **C** : staff cancels **one** series (possibly out of many) related to an application section: `"The space reservation included in your seasonal booking has been cancelled"` (same as emails A & B when a single reservation occurrence is cancelled/rejected but the weekday level time info clarifies how "reservation" is to be understood)
- **D** applicant cancels all series related to an application section: `"All space reservations included in your seasonal booking have been cancelled"`

tilavarauspalvelu/integrations/email/main.py
vergama: True, just in case somebody does this before the application round results have actually been sent to the applicants. (Even if at the moment the mutation which triggers this is only called from customer UI and should not be available there before that point -- but application section wide series rejection may well be later implemented for staff users.)

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -80,81 +137,64 @@ def send_application_received_email(
...
 
     @staticmethod
-    def send_application_handled_emails() -> None:
-        """Sends email to applicants when application round has its allocation results available."""
-        applications = Application.objects.should_send_handled_email().order_by("created_date")
-        if not applications:
-            SentryLogger.log_message("Zero applications require the 'application handled' email to be sent")
+    def send_staff_notification_application_section_cancelled(application_section: ApplicationSection) -> None:
+        """Sends an email to Staff that the whole application section was cancelled by the user"""
+        reservation: Reservation | None = application_section.actions.get_last_reservation()
```
vergama, line 173: This misses the possibility that the application section contain multiple recurring reservation series at different (reservation) units, for which the recipients may be different (or the same, so need to deduce the unique recipients across them).

tilavarauspalvelu/typing.py
```diff
@@ -40,7 +40,7 @@ class permission(classmethod): ...  # noqa: N801
 type Lang = Literal["fi", "sv", "en"]
 type TextSearchLang = Literal["finnish", "english", "swedish"]
 type M2MAction = Literal["pre_add", "post_add", "pre_remove", "post_remove", "pre_clear", "post_clear"]
-type EmailContext = dict[str, str | None]
+type EmailContext = dict[str, str | int | datetime.datetime | Decimal | None]
```
matti-lamppu, line 43: I wouldn't allow datetime in the email context, since it should probably be formatted in the local datetime-format before applying to the template. By default, datetimes would show up as ISO strings, which is not the best for customer-facing data.

matti-lamppu: Not sure about Decimals, likely you'd want to make sure that it's formatted for displaying prices, although you could do this in the template aswell.

tilavarauspalvelu/admin/email_template/forms.py
ranta: Yep, you're right, not sure what I was thinking writing it like this 😆

tilavarauspalvelu/admin/email_template/forms.py
```diff
@@ -162,384 +149,432 @@ class PendingInstructionsFormMixin(forms.Form):
...
 
-class ApplicationInAllocationEmailTemplateTesterForm(
-    LanguageFormMixin,
+class ReservationBaseForm(
+    ReservationBasicInfoFormMixin,
+    EmailRecipientFormMixin,
     BaseEmailTemplateForm,
 ):
+    @classmethod
+    def get_initial_data_from_reservation_unit(cls, instance: ReservationUnit, *, language: Lang) -> Self:
```
matti-lamppu, line 186: Return typing is wrong for this method in many of the subclasses. Should be `dict[str, Any]` like the superclass method.

tilavarauspalvelu/integrations/email/template_context/application.py
ranta: My intention here pretty much to abuse the `# type: ` comment to have the IDE recognise the type and create a "link" directly to the Enum. Changing it to `# email type:` breaks this "link", but I also wouldn't want to break the standards 🤔

matti-lamppu: You could use `Annotated[EmailContext, EmailType.APPLICATION_HANDLED]`, though I suppose it's a bit long. Well, it's not like there is anything wrong abusing the type comment like this so it's your call.

tilavarauspalvelu/integrations/email/template_context/reservation.py
ranta: Nice catch!

tilavarauspalvelu/integrations/email/template_context/application.py
matti-lamppu: Also, I think the fact that type comments create links like this might be a PyCharm specific thing, though I'm not sure.

ranta: It's what we both use, and since this is not a big issue and makes the dividers prettier, lets keep it as-is for now

tilavarauspalvelu/integrations/email/main.py
ranta: Added a fix to get staff recipients for all reservation units options in the application section

tilavarauspalvelu/typing.py
ranta: I think i added `datetime.datetime` there because I saw it being returned somewhere, but im not finding that anymore... 🤔 
Price we use only for Prices (I think) and they are already being formatted in the template.

tilavarauspalvelu/integrations/email/find_recipients.py
```diff
@@ -117,3 +117,40 @@ def get_reservation_staff_notification_recipients_by_language(
...
+        reservation_notification__in=notification_settings,
+    ).exclude(email="")
+
+    # Skip the reservation creator
+    if application_section.application.user:
+        users = users.exclude(pk=application_section.application.user.pk)
+
+    units = (
+        Unit.objects.prefetch_related("unit_groups")
+        .filter(reservation_units__reservation_unit_options__application_section=application_section)
```
vergama, line 146: The reservation units need to be fetched from the actual reservation series generated from the allocations, since original options set while applying may contain ones for which nothing was allocated in application handling.

ranta: Very good point, fixed this :)

tilavarauspalvelu/integrations/email/find_recipients.py
```diff
@@ -117,3 +117,44 @@ def get_reservation_staff_notification_recipients_by_language(
...
+def get_application_section_staff_notification_recipients_by_language(
+    application_section: ApplicationSection,
+) -> dict[Lang, set[str]]:
+    """
+    Get staff users who should receive email notifications for the given application.
+    Group recipients by their preferred language.
+    """
+    recipients_by_language: dict[Lang, set[str]] = defaultdict(set)
+
+    notification_settings = [ReservationNotification.ALL]
```
vergama, line 131: Can also be `ReservationNotification.ONLY_HANDLING_REQUIRED` beside `ALL`.

tilavarauspalvelu/integrations/email/find_recipients.py
```diff
@@ -117,3 +117,44 @@ def get_reservation_staff_notification_recipients_by_language(
...
+
+    # Only fetch users with active unit roles and appropriate notification settings.
+    users = User.objects.filter(
+        unit_roles__isnull=False,
+        unit_roles__role_active=True,
+        reservation_notification__in=notification_settings,
+    ).exclude(email="")
+
+    # Skip the reservation creator
```
matti-lamppu, line 140: Comment is copy-pasta.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -475,7 +473,8 @@ def send_seasonal_reservation_modified_series_email(
         if reservation.type != ReservationTypeChoice.SEASONAL:
             return
 
-        recipients = get_reservation_email_recipients(reservation=reservation)
+        application = reservation_series.actions.get_application_section().application
```
matti-lamppu, line 476: Could just use the ORM here: `reservation_series.allocated_time_slot.reservation_unit_option.application_section.application`.

tilavarauspalvelu/integrations/email/main.py
```diff
@@ -507,7 +506,8 @@ def send_seasonal_reservation_rejected_series_email(
...
```
matti-lamppu, line 509: Could just use the ORM here: `reservation_series.allocated_time_slot.reservation_unit_option.application_section.application`.

matti-lamppu (approved):
Only very minor comments from me, looks good otherwise.

PR #1485, opened 2025-01-08
Add access type for keyless entry

matti-lamppu:
## 🛠️ Changelog

- Add new `access_type` (="kulkuoikeus") to the reservation unit model for keyless entry
- Add new `access_type_start_date` and `access_type_end_date` to the reservation unit model for enabling onboarding to or offboarding from different access types
- Add a new PindoraClient for interfacing with the Pindora API
- Add `access_type` to the admin panel and check the reservation unit from Pindora if changed to "ACCESS_CODE"
- Add new fields to the GraphQL schema
- Add "access_type" filtering to the GraphQL schema

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3737](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3737)
- [TILA-3750](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3750)
- [TILA-3749](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3749)

[TILA-3737]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tilavarauspalvelu/integrations/keyless_entry/client.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @classmethod
+    def _get_headers(cls, headers: dict[str, Any] | None) -> dict[str, str]:
+        if not settings.PINDORA_API_KEY:
+            msg = "'PINDORA_API_KEY' environment variable must to be configured."
+            raise PindoraAPIConfigurationError(msg)
+
+        return {
+            "Pindora-Api-Key": str(settings.PINDORA_API_KEY),
+            "Accept": "application/vdn.varaamo-pindora.v1+json",
```
vergama, line 226: Just `application/json` in the OpenAPI description, after all.

vergama (approved):

tilavarauspalvelu/integrations/keyless_entry/client.py
matti-lamppu: Asked about this in Keyless Entry Trello, we can switch it if they have decided in some other versioning method.

tilavarauspalvelu/models/reservation_unit/model.py
```diff
@@ -88,6 +89,8 @@ class ReservationUnit(models.Model):
     max_reservation_duration: datetime.timedelta | None = models.DurationField(null=True, blank=True)
     buffer_time_before: datetime.timedelta = models.DurationField(default=datetime.timedelta(), blank=True)
     buffer_time_after: datetime.timedelta = models.DurationField(default=datetime.timedelta(), blank=True)
+    keyless_entry_start_date: datetime.date | None = models.DateField(null=True, blank=True)
+    keyless_entry_end_date: datetime.date | None = models.DateField(null=True, blank=True)
```
ranta, line 93: Should we have validation on these fields so that `keyless_entry_start_date` can't be after `keyless_entry_end_date`?

ranta (commented):
Something that should be clarified is the logic on how the logic between `keyless_entry_start_date`, `keyless_entry_end_date` and `method_of_entry` works.

If the method is set to `MethodOfEntry.KEYLESS` and start/end date is set, what is the method before/after the date?
I would assume it would always just be `KEYLESS`, but the ticket mentions:
> (Tämän aikavälin ulkopuolella kulku katsotaan kuitenkin avoimeksi.)

Having the start/end dates set when the method is `KEYLESS` would be a bit confusing and open to open for interpretation. I'm thinking if we should the block setting start/end dates if the method is set to `KEYLESS`?

---

The ticket also mentions:
> (Varausten tilastoihin tarvitaan kuitenkin boolean-tieto, oliko kyseessä pääsykoodin vaatinut varaus.)

Which is not implemented here. But it is also not in the acceptance criteria, so I would assume that this is coming later at some point?

---

In HAUKI there are some states (`HaukiResourceState`) that mention a key, do those relate to `WITH_KEY`/`KEYLESS` in any way? 🤔

tilavarauspalvelu/models/reservation_unit/model.py
matti-lamppu: I think these can work similarly to how `publish_begins` and `publish_ends` affect `publishing_state`, which means that end before begin if keyless entry was, e.g., disabled at some point and then re-enabled in the future.

---

Regarding how these dates affect the `method_of_entry`, it hasn't been fully specified here, but I'd assume that:

- If current method of entry is `OPEN_ACCESS` or `WITH_KEY`, and `keyless_entry_start_date` is in the future, and `keyless_entry_end_date` has not been set or is in the past, then the method of entry after that date is `KEYLESS`
- If current method of entry is `KEYLESS` and `keyless_entry_end_date` is set in the future, and `keyless_entry_start_date` has not been set or is in the past, then the method of entry after that date is `OPEN_ACCESS` (`WITH_KEY` cannot be specified with this system).
- If current method of entry is `OPEN_ACCESS` or `WITH_KEY`, and `keyless_entry_start_date` is in the future, but `keyless_entry_end_date` has also been set after `keyless_entry_start_date`, then the method of entry is `KEYLESS` after `keyless_entry_start_date` until `keyless_entry_end_date`, after which it will be either `OPEN_ACCESS` or `WITH_KEY` again.
- If current method of entry is `KEYLESS` and `keyless_entry_end_date` is set in the future, but `keyless_entry_start_date` has also been set after `keyless_entry_end_date`, then the method of entry is `OPEN_ACCESS` after `keyless_entry_end_date` until `keyless_entry_start_date`, after which it will be `KEYLESS` again.

I don't think there is other combinations. Perhaps this should be an action: `get_method_of_entry_at(moment: datetime.datetime) -> MethodOfEntry`?

matti-lamppu:
> The ticket also mentions:
> 
> > (Varausten tilastoihin tarvitaan kuitenkin boolean-tieto, oliko kyseessä pääsykoodin vaatinut varaus.)
> 
> Which is not implemented here. But it is also not in the acceptance criteria, so I would assume that this is coming later at some point?

This can be implemented when keyless entry is integrated to the reservations flow.

matti-lamppu:
> In HAUKI there are some states (HaukiResourceState) that mention a key, do those relate to WITH_KEY/KEYLESS in any way? 🤔

No relation.

tilavarauspalvelu/models/reservation_unit/model.py
matti-lamppu: For scenario 2, I guess you could swap the method of entry to WITH_KEY and set `keyless_entry_end_date` to the future, which would allow switching from `KEYLESS` to `WITH_KEY` on a certain date.

ranta: I don't really like the many choices and ways of interpretation here.

How I would like simplify this is:
1. Force both `keyless_entry_start_date` and `keyless_entry_end_date` to be `null` if the method is `KEYLESS`

or

2. Remove `KEYLESS` from the method options, and have it be fully defined by the start/end dates like you described above

vergama: Perhaps this would suffice:
- rename as `method_of_entry_start_date` & `method_of_entry_end_date` and validate `start < end`
- the selected method_of_entry applies between those dates (i.e., `[start, end)`) and any other time is considered `OPEN_ACCESS`
- either all three are set or are null

I doubt that it is needed to configure more than one upcoming entry method change at a time for a given reservation unit, or keep the history once an end date is past (in which if open access changes to something else again, it will be after that time, when new values can be set) -- so a series of activation date/method combinations (like the reservation unit pricings) might be overly complicated (also considering they would be maintained in Django admin).

ranta: That would suffice, if there is never a need to change from `KEYLESS` -> `WITH_KEY` on a set date (with this solution it's not supported or it would need to be manually changed at the time the change is needed)

I would assume the `OPEN_ACCESS` option needs to be removed from `method_of_entry`, since it doesn't really make sense there anymore 🤔

matti-lamppu: Alright, I make the following changes:

- Changed `keyless_entry_start_date` to `method_of_entry_start_date`
- Changed `keyless_entry_end_date` to `method_of_entry_end_date`
- Added constraint to model that `keyless_entry_start_date <= method_of_entry_end_date` if both are set
- Added a lookup property for calculating `current_method_of_entry` to expose from endpoints instead of `method_of_entry`
- Added a `ReservationUnit` action `get_method_of_entry_at` to calculate actual method of entry at some moment of time for use in creating reservations
- Added tests to verify that these properties work.

matti-lamppu: > I would assume the OPEN_ACCESS option needs to be removed from method_of_entry, since it doesn't really make sense there anymore 🤔

@ranta I decided to leave `OPEN_ACCESS` in the enum since it's easier to work with non-nullable fields.

vergama (approved):

ranta (approved):

matti-lamppu:
@ranta @vergama Added method of entry fields to GraphQL schema and a filter for filtering units that have certain method of entries on a given period (see new commits)

vergama (approved):

ranta (approved):

PR #1495, opened 2025-01-20
Refactor all reservation mutation APIs

matti-lamppu:
## 🛠️ Changelog
[//]: # "Describe the changes in this pull request here."

- Refactors all reservation mutation APIs to use newer serializers
- Adds new "ModelValidator" classes to Reservations and Reservation units where validation logic can be written to for better reusability. Can be used with `model_instance.validator.validation_method()`
- Makes some changes to the API:s of these mutations, mainly in order to make them simpler
    - Creating reservation for multiple reservation unit is no longer possible through the API
    - Drop reservation language from the APIs
    - Rename some related fields ending with "pk" to simply their name, e.g. "purposePk" -> "purpose"
    - For duration inputs, use integers representing durations in seconds instead of ISO formatted duration strings "HH:MM:SS" (mainly applies to staff mutations)
    - Drop some unused logic from update serializers that has been either replaced with more specific endpoints or dropped altogether
    - No longer allow creating recurring reservations using reservation API, should use custom reservation series API
    - Remove `ValidationErrorWithCode` and use ValidationErrors like in rest of endpoints. Changes how errors are returned in responses to match style elsewhere

## 🧪 Test plan
[//]: # "Help your fellow reviewer and write a short description of the fastest way to test your changes."

- Automated tests
- Should manually test paid reservations to make sure price calculation is correct

## 🚧 Dependencies
[//]: # "Is this PR dependent on other changes outside this repository? Describe the changes, or add links to them."
[//]: # "e.g. This PR breaks X and is waiting for frontend changes before merging."

- https://github.com/City-of-Helsinki/tilavarauspalvelu-ui/pull/1620

## 🎫 Tickets
[//]: # "This pull request resolves all or part of the following ticket(s)."

- None

PR #1499, opened 2025-01-22
Remove reservee language from reservations

matti-lamppu:
## 🛠️ Changelog

- Use `user.get_preferred_language()` instead

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Requires frontend to match API. Should likely merge #1495 first

## 🎫 Tickets

- [TILA-3706](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3706)

[TILA-3706]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/test_graphql_api/test_application_section/test_filtering.py
```diff
@@ -1049,17 +1050,19 @@ def test_application_section__filter__by_text_search__application_id(graphql):
...
     application_2 = ApplicationFactory.create_in_status_draft_no_sections(
+        id=2,
         organisation=None,
         contact_person=None,
         user=None,
     )
-    section_1 = ApplicationSectionFactory.create_in_status_unallocated(application=application_1, name="foo")
-    ApplicationSectionFactory.create_in_status_unallocated(application=application_2, name="bar")
+    section_1 = ApplicationSectionFactory.create_in_status_unallocated(id=3, application=application_1, name="foo")
+    ApplicationSectionFactory.create_in_status_unallocated(id=4, application=application_2, name="bar")
```
matti-lamppu, line 1065: Just fixing a flaky test here

vergama (approved):

tilavarauspalvelu/management/commands/data_creation/create_reservations.py
```diff
@@ -269,7 +268,7 @@ def _create_normal_reservations(
 
 def _build_payment_order(reservation: Reservation, payment_types_choices: list[PaymentType]) -> PaymentOrder:
     payment_order_builder = PaymentOrderBuilder().set(
-        language=reservation.reservee_language or Language.FI,
+        language=reservation.user.get_preferred_language(),
```
ranta, line 271: `reservation.user` is nullable, so accessing the `user` might be an issue in some cases

This comment applies for other changes as well

tests/test_graphql_api/test_application_section/test_filtering.py
ranta: I don't think the hardcoded IDs are needed here, maybe lets just remove them all together

tilavarauspalvelu/management/commands/data_creation/create_reservations.py
matti-lamppu: In practice this is not true since the user is always set for all reservations and only removed if the actual user is removed from the system (has `on_delete=SET_NULL`). I also think operations on reservations should fail if there is no user for it.

tests/test_graphql_api/test_application_section/test_filtering.py
matti-lamppu: They are needed, since application section text search accounts for application ids as well as section ids, so to remove the flakiness we need to give each of them unique ids so that they are not accidentally the same between application and application section (=source of flakiness)

PR #1511, opened 2025-01-28
Frontend Testing API

ranta:
## 🛠️ Changelog

- Framework for creating endpoints for Frontend Testing API
- Endpoint for creating testing data for a single reservation

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3746](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3746)
- [TILA-3775](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3775)

[TILA-3746]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TILA-3775]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -512,10 +512,10 @@ class Common(Environment):
     PERMISSION_NOTIFICATION_BEFORE_DAYS = values.IntegerValue(default=14)
     RAISE_ERROR_ON_REFRESH_FAILURE = False
 
+    ICAL_HASH_SECRET = values.StringValue()
+
     # Allows faking membership to certain AD groups for testing automatic role assignment
     FAKE_SUPERUSER_AD_GROUPS = values.ListValue(default=[])
-
-    ICAL_HASH_SECRET = values.StringValue()
 
 
 class EmptyDefaults:
@@ -573,6 +573,8 @@ class EmptyDefaults:
 
     EMAIL_VARAAMO_EXT_LINK = ""
     EMAIL_FEEDBACK_EXT_LINK = ""
+
+    FRONTEND_TESTING_API_ENABLED = True
 
 
 class Local(Common, overrides_from=LocalMixin):
@@ -902,6 +904,10 @@ class Platta(Common, use_environ=True):
     REDIS_PASSWORD: str = values.StringValue()
 
     DJANGO_REDIS_CONNECTION_FACTORY = "django_redis.pool.SentinelConnectionFactory"
+
+    # --- Misc settings ----------------------------------------------------------------------------------------------
+
+    FRONTEND_TESTING_API_ENABLED = False
 
     @classproperty
     def CACHES(cls):
--- a/tests/test_frontend_testing_api/test_reservation.py
+++ b/tests/test_frontend_testing_api/test_reservation.py
@@ -1,0 +1,178 @@
+from __future__ import annotations
+
+import datetime
+
+import pytest
+from rest_framework.reverse import reverse
+
+from tilavarauspalvelu.api.frontend_testing_api.helpers import unfreeze_at_end
+from tilavarauspalvelu.enums import OrderStatus, PaymentType, ReservationStateChoice
+from tilavarauspalvelu.models import PaymentOrder, Reservation, ReservationUnit, ReservationUnitPricing, Unit, User
+from utils.date_utils import local_datetime
+
+pytestmark = [
+    pytest.mark.django_db,
+]
+
+
+_ENDPOINT = "frontend_testing_api:reservation-list"
+
+
+def test_frontend_api__reservation__user__general_role(api_client):
+    data = {
+        "user": {"role": "HANDLER"},
+        "params": {},
+    }
+    response = api_client.post(reverse(_ENDPOINT), data=data, format="json")
+    assert response.status_code == 201, response.json()
+
+    user = User.objects.first()
+    assert user.permissions.can_manage_reservation(Reservation.objects.first(), reserver_needs_role=True)
+    assert user.permissions.has_general_role(role_choices=["HANDLER"])
+    assert not user.permissions.has_role_for_units_or_their_unit_groups(units=Unit.objects.all())
+
+
+def test_frontend_api__reservation__user__unit_role(api_client):
+    data = {
+        "user": {"role": "ADMIN", "is_unit_role": True},
+        "params": {},
+    }
+    response = api_client.post(reverse(_ENDPOINT), data=data, format="json")
+    assert response.status_code == 201, response.json()
+
+    user = User.objects.first()
+    assert user.permissions.can_manage_reservation(Reservation.objects.first(), reserver_needs_role=True)
+    assert not user.permissions.has_general_role()
+    assert user.permissions.has_role_for_units_or_their_unit_groups(units=Unit.objects.all())
+
+
+def test_frontend_api__reservation__user__not_provided_no_auth(api_client):
+    data = {
+        "params": {},
+    }
+    response = api_client.post(reverse(_ENDPOINT), data=data, format="json")
+    assert response.status_code == 201, response.json()
+
+    user = User.objects.first()
+    reservation = Reservation.objects.first()
+    assert user == reservation.user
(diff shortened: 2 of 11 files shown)


config/settings.py
```diff
@@ -863,6 +869,8 @@ class AutomatedTests(EmptyDefaults, Common, dotenv_path=None, overrides_from=Aut
     EMAIL_VARAAMO_EXT_LINK = "https://fake.varaamo.hel.fi"
     EMAIL_FEEDBACK_EXT_LINK = "https://fake.varaamo.hel.fi/feedback"
 
+    FRONTEND_TESTING_API_ENABLED = True
```
matti-lamppu, line 872: I would add this to the `EmptyDefaults` mixin instead.

config/settings.py
```diff
@@ -881,6 +889,8 @@ class CI(EmptyDefaults, Common, use_environ=True):
     # Migrations require the database
     DATABASES = values.DatabaseURLValue()
 
+    FRONTEND_TESTING_API_ENABLED = values.BooleanValue(default=True)
```
matti-lamppu, line 892: I would add `FRONTEND_TESTING_API_ENABLED=True` to the `EmptyDefaults` mixin instead.

tests/test_frontend_testing_api/test_reservation.py
```diff
@@ -0,0 +1,151 @@
...
+    assert reservation.begin == local_datetime(2022, 1, 2, 12)
+    assert reservation.state == ReservationStateChoice.CONFIRMED
+    assert reservation.recurring_reservation is None
+
+    pricing = ReservationUnitPricing.objects.first()
+    assert pricing.highest_price == 0
+
+    assert PaymentOrder.objects.count() == 0
+
+    stop_time_freeze()
```
matti-lamppu, line 89: This will not get run if the test runs into an error (like an assert failing), which would break the rest of the tests due to wrong local time begin set. Instead, use a decorator with a finally block:

```python
@contextmanager
def unfreeze_at_end():
    try:
        yield
    finally:
        stop_time_freeze()

@unfreeze_at_end
def test_frontend_api__reservation__set_datetime__default_options(api_client): ...
```

tests/test_frontend_testing_api/test_reservation.py
```diff
@@ -0,0 +1,151 @@
...
+    assert ReservationUnitPricing.objects.count() == 1
+    pricing = ReservationUnitPricing.objects.first()
+    assert pricing.highest_price == 10
+
+    assert PaymentOrder.objects.count() == 1
+    payment_order = PaymentOrder.objects.first()
+    assert payment_order.payment_type == PaymentType.ON_SITE
+    assert payment_order.status == OrderStatus.DRAFT
+
+    stop_time_freeze()
```
matti-lamppu, line 132: See comment: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1511/files#r1933342361

tilavarauspalvelu/api/frontend_testing_api/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+def stop_time_freeze() -> None:
+    """
+    Stop the time freeze if it is currently active.
+    This must be called after each test that uses `set_current_datetime`.
+    """
+    global freeze_factory  # noqa: PLW0603
+
+    if freeze_factory:
+        freeze_factory.stop()
+        freeze_factory = None
```
matti-lamppu, line 39: Now `setup_now_tt` can be run multiple times if `stop_time_freeze` is run, since it deletes the `freeze_factory` that is used to check if `setup_now_tt` should be run in `set_current_datetime`. This also creates a memory leak, since `setup_now_tt` creates a patch on top of `freezegun.api._freeze_time` each time it is run.

You could modify `setup_now_tt` to return the patch it creates and also save that globally here, and then stop that together with the `freeze_factory` in `stop_time_freeze`.

To note: `setup_now_tt` is called automatically at the start of the test run, and other tests expect this to be applied, so you need to be careful to not remove this first patch, only the one applied by `set_current_datetime`. i.e there `set_current_datetime` should have applied a patch on top of another patch, and you only want to remove the outer patch to prevent memory leaks.

tilavarauspalvelu/api/frontend_testing_api/reservation.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        state = attrs.get("state")
+        is_paid_reservation = attrs.get("is_paid_reservation")
+        is_paid_on_site = attrs.get("is_paid_on_site")
+
+        if not is_paid_reservation and any([
+            state == ReservationStateChoice.WAITING_FOR_PAYMENT,
+            is_paid_on_site,
+        ]):
```
matti-lamppu, line 47: I feel like this is simpler without the `any`:

```python
if not is_paid_reservation and (
    state == ReservationStateChoice.WAITING_FOR_PAYMENT
    or is_paid_on_site,
):
```

tilavarauspalvelu/api/frontend_testing_api/serializers.py
```diff
@@ -0,0 +1,104 @@
...
+        return attrs
+
+    def save(self, **kwargs: Any) -> None:
+        """Simplified version of BaseSerializer.save() without the need to return an instance of the created object"""
+        assert hasattr(self, "_errors"), "You must call `.is_valid()` before calling `.save()`."  # noqa: S101
+        assert not self.errors, "You cannot call `.save()` on a serializer with invalid data."  # noqa: S101
+
+        validated_data = {**self.validated_data, **kwargs}
+
+        self.instance = self.create(validated_data)
```
matti-lamppu, line 38: Why is this implemented?

tilavarauspalvelu/models/reservation_unit_pricing/actions.py
```diff
@@ -32,4 +33,4 @@ def calculate_reservation_price(self, duration: datetime.timedelta, *, subsidise
         duration_seconds = int(duration.total_seconds())
         duration_minutes = int(math.ceil(duration_seconds / 60 / 15) * 15)
 
-        return (price / price_unit.in_minutes) * duration_minutes
+        return round_decimal((price / price_unit.in_minutes) * duration_minutes, 2)
```
matti-lamppu, line 36: I don't think rounding should be done to the price here, it can be applied on relevant precision when needed, but this function should return the _actual_ price.

matti-lamppu (changes requested):
See comments

tilavarauspalvelu/api/frontend_testing_api/serializers.py
ranta: Initially added this to remove the need to return anything from `.create()` methods, but removed it now since it isn't really needed 😄

config/settings.py
ranta: Set this as `True` in `EmptyDefaults` and as `False` in Platta`

tilavarauspalvelu/api/frontend_testing_api/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+        now_tt_freeze_factory.stop()
+        freeze_factory = None
+
+
+@contextmanager
+def unfreeze_at_end() -> None:
+    """
+    Temporarily unfreeze the time for the duration of the context.
+
+    This should be used for all tests tht call `set_current_datetime`, so the time is not frozen for all later tests.
```
matti-lamppu, line 51: tht -> that

matti-lamppu (approved):

PR #1616, opened 2025-03-03
Admin UI access code series

ranta:
## 🛠️ Changelog

- Add Access code support for Reservation Series reservations
- Permission check and Confirmation dialog for changing the access code
- Refactor Admin UI Reservation page

## 🧪 Test plan

- Manual testing vs the ticket specs

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3839](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3839)

[TILA-3839]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+      initiallyOpen={false}
+    >
+      <div>
+        <SummaryHorizontal $isRecurring={!!reservation.recurringReservation}>
+          <DataWrapper label={t("RequestedReservation.accessCodeLabel")}>
+            {reservation.pindoraInfo?.accessCode ?? "-"}
+          </DataWrapper>
+
+          <DataWrapper label={t("RequestedReservation.accessCodeStatusLabel")}>
+            {reservation.pindoraInfo?.accessCodeIsActive
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:100]: In case of a recurring reservation the pindoraInfo needs to be fetched through that parent, since this particular reservation in the series can be one without an access code (e.g. date with direct access) or in the past (in which case GQL won't return the info through it).

joonatank (approved):
👍

apps/admin-ui/src/spa/reservations/[id]/index.tsx
```diff
@@ -524,169 +385,46 @@ function RequestedReservation({
...
-              reservationPk={reservation.pk ?? 0}
-              refetch={refetch}
-              initialValue={reservation.workingMemo ?? ""}
-            />
-          </Accordion>
-        </VisibleIfPermission>
+          onReservationUpdated={refetch}
+        />
+
         {reservation.accessType === AccessType.AccessCode && (
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/index.tsx:410]: If the reservation belongs to a recurringReservation, the check should be `recurringReservation.usedAcessTypes.includes(AccessType.AccessCode`) -- since the `usedAcessTypes` array contains all unique access types across its individual reservations and we wan't to show the section in the UI if any one them (potentially) uses a code.

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+              ? t("RequestedReservation.accessCodeStatusActive")
+              : t("RequestedReservation.accessCodeStatusInactive")}
+            {reservation.pindoraInfo?.accessCodeIsActive !==
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:103]: There will be lookup properties (available in the GQL, suitable prop name TBD, check with Matti) for both the `reservation` and `recurringReservation` models and for which the backend will deduce whether the "state" of the access code is correct or not:
- `reservation`: incorrect/false if (should have active access code generated but the code is inactive or missing) or (has active access code but it should be either inactive or removed)
- `recurringReservation`: incorrect if it contains reservations (of the type AccessCode but there is at least one in Confirmed state for which the time has not been scheduled to Pindora so is not covered by the validity of the code although it should) or (in another res.state or access type yet the time remains scheduled in Pindora so the code is valid for a reservation for which it should not)

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+              reservation.accessCodeShouldBeActive && <IconAlertCircleFill />}
+          </DataWrapper>
+
+          {reservation.recurringReservation && (
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:107]: In the recurring case, the data for `recurringReservation.pindoraInfo.accessCodeValidity` is an array which contains the begin/end times in Pindora for all the reservations in it that are supposed to have an access code; the first and last date can be looked up from it.

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+                  ? formatDate(reservation.recurringReservation.endDate)
+                  : "-"}
+              </DataWrapper>
+            </>
+          )}
+
+          <DataWrapper
+            label={t("RequestedReservation.accessCodeValidityLabel")}
+          >
+            {reservation.pindoraInfo
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:125]: In the recurring case, since `pindoraInfo.accessCodeValidity` is an array of 1-N occurrences, we shall display the time of the currently ongoing or next (future) one contained in it (cf. tooltip in Figma). Or lacking either, the last (past) one.

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const { hasPermission } = useCheckPermission({
+    units: [reservation.reservationUnits?.[0]?.unit?.pk ?? 0],
+    permission: UserPermissionChoice.CanManageReservations,
+  });
+
+  const handleExecuteMutation = async () => {
+    const payload = { variables: { input: { pk: reservation.pk ?? 0 } } };
+
+    try {
+      if (isAccessCodeBroken) {
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:166]: The mutations should be the opposite way.

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,251 @@
...
+  reservation: ReservationType;
+  onSuccess: () => void;
+}>) {
+  const { t, i18n } = useTranslation();
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const [changeAccessCodeMutation] = useChangeReservationAccessCodeMutation();
+  const [repairAccessCodeMutation] = useRepairReservationAccessCodeMutation();
+
+  const isAccessCodeBroken =
```
Comment (vergama) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:153]: The backend/GQL will be able to provide the info, at both the reservation and recurringReservation levels, whether the "state" of the access code is correct, all things considered (see another comment). It should then be possible to decide which action button to show (if the user has permissions) something like this:
1. the "state" is incorrect in one way or another: show the "repair" button.
2. correct and furthermore (at least one in the series or the) reservation uses access code (which can then be changed): `reservation/recurrringReservation.has shouldHaveActiveAccessCode == true` should be a sufficient check (since if they also did not have an active code, the "state" would be incorrect)
3. _else_: no button since nothing to fix nor is there an existing, correctly synchronized code to change.

vergama (changes requested):
Bunch or retro-active specification.

vergama (approved):
In **AccessCodeChangeRepairButton** when the reservation is part of _recurringReservation_, `isAccessCodeIsActiveCorrect` will need to be checked from the parent and likewise `isReservationEnded` (across its reservations).
But the mutations for the recurring variant are not yet available anyway, the display part looks correct.

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,370 @@
...
+      }
+    }
+    if (!validityBeginsTime) {
+      // If no next validity date found, use the last one
+      validityBeginsTime =
+        accessCodeValidity[accessCodeValidity.length - 1].accessCodeBeginsAt;
+      validityEndsTime =
+        accessCodeValidity[accessCodeValidity.length - 1].accessCodeEndsAt;
+    }
+  }
```
Comment (joonatank) [inline apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx:213]: can this be moved to a function? not a fan `let` and `for` loop in component body, allows accidental reassingment in JSX.

also not a fan of using `const now = new Date()` inside component body without `useMemo`, probably not causing issues here since it's only used for comparison (no hook calls / cbs). np in a free function

packages/common/src/components/form/ControlledDateInput.tsx
```diff
@@ -44,7 +44,7 @@ export function ControlledDateInput<T extends FieldValues>({
     <DateInput
       {...rest}
       id={id ?? `controlled-date-input__${name}`}
-      label={label ?? t(`ReservationDialog.${name}`)}
+      label={label ?? t(`common.${name}`)}
```
joonatank, line 47: note if you change the translation key

joonatank (approved):
👍

some stylistic / preference things that could be changed

apps/admin-ui/src/i18n/messages.ts
```diff
@@ -153,20 +152,24 @@ const translations: ITranslations = {
...
     inspect: ["Tarkastele"],
     cityOfHelsinki: ["Helsingin kaupunki"],
     chooseAction: ["valitse toiminto"],
     weekday: ["Viikonpäivä"],
     date: ["Päivämäärä"],
     time: ["Kellonaika"],
+    startingDate: ["Aloituspäivä"],
+    endingDate: ["Päättymispäivä"],
+    startTime: ["Aloitusaika"],
+    endTime: ["Päättymisaika"],
```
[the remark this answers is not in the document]
ranta: Made a ticket to migrate all translations at once sometime in the future

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+  justify-content: flex-end;
+`;
+
+const DataWrapperContainer = styled.div`
+  display: flex;
+  gap: var(--spacing-xs);
+  align-items: center;
+  white-space: nowrap;
+`;
```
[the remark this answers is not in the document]
Comment (ranta) [reply apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx]: Fixed

apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx
Comment (ranta) [reply apps/admin-ui/src/spa/reservations/[id]/ReservationKeylessEntrySection.tsx]: Fixed

PR #1634, opened 2025-03-14
Separate Celery worker and Celery beat Scheduler

matti-lamppu:
## 🛠️ Changelog

This is done for two reasons:
1. There should always be just one scheduler. Currently, we start a single worker with the scheduler, but this prevents us from adding more workers if necessary.
2. Since the worker is started in the background, we don't get its logs, which makes debugging harder. Separating the worker and the scheduler to different containers allows us to view both of their logs.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Need to setup Azure Pipelines to support two celery containers, otherwise background tasks stop working. This has been done in the [stop-worker-before-rollout](https://dev.azure.com/City-of-Helsinki/1dfa9ae6-397b-44b5-86f9-0faa3e3eba87/_git/tilavarauspalvelu-pipelines?version=GBstop-worker-before-rollout) branch in Azure Pipelines

## 🎫 Tickets

- None


matti-lamppu:
Came across this when debugging some allocation -> reservation generation logic. Should be fixed so that we can observe the worker logs in prod.

PR #1658, opened 2025-03-21
Display access type & code in an approved application's reservation listing

vincit-matu:
## 🛠️ Changelog

- displays the access type (and it's access code where applicable) in the reservations listing view for an approved seasonal appplication (see [ticket](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3842) for specs on what should be shown where)
- shows the access types and their validity periods in the instructions for each reservation series in the above mentioned listing view as requested in [TILA-3843](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3843)
- extracts `getReservationUnitAccessPeriods` to a reservationUnit helper function
- adds `ApplicationSectionReservationUnit` and `PindoraInfo` fragments

## 🧪 Test plan

- Manual testing
  - to see all permutations you need to end an application round, with has a suitable application in it, so that it has assigned reservations to said application)
  - necessary approved reservation series:
    - one to a reservation unit with accessCode as its access type
    - two to a reservation unit where the access type changes during the application rounds duration (can be done afterwards via Django) - one of them should have accessCode as one of the access types, the other should not
- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3842](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3842)
- [TILA-3843](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3843)

[TILA-3843]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TILA-3842]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ui/components/application/ApprovedReservations.tsx
+++ b/apps/ui/components/application/ApprovedReservations.tsx
@@ -1,22 +1,28 @@
+import React, { useState } from "react";
 import {
   type ApplicationSectionReservationFragment,
   useApplicationReservationsQuery,
-  type ReservationUnitNode,
   ReservationStateChoice,
+  AccessType,
+  ReservationUnitNode,
   type ApplicationNode,
+  ApplicationSectionReservationUnitFragment,
+  AccessTypeWithMultivalued,
+  PindoraSectionFragment,
+  PindoraReservationFragment,
+  ReservationUnitAccessTypeNode,
 } from "@/gql/gql-types";
 import {
   getApplicationReservationPath,
   getApplicationSectionPath,
   getReservationUnitPath,
 } from "@/modules/urls";
-import { breakpoints, fontMedium, fontRegular, H5 } from "common";
+import { breakpoints, fontBold, fontMedium, fontRegular, H5 } from "common";
 import {
   getTranslationSafe,
   toApiDate,
   toUIDate,
 } from "common/src/common/util";
-import { IconButton, StatusLabel } from "common/src/components";
 import {
   filterNonNullable,
   formatApiTimeInterval,
@@ -38,12 +44,10 @@ import {
   IconLocation,
   IconPen,
   Table,
+  Tooltip,
 } from "hds-react";
-import React, { useState } from "react";
 import { useTranslation } from "next-i18next";
 import styled from "styled-components";
-import { Sanitize } from "common/src/components/Sanitize";
-import { LinkLikeButton } from "common/styles/buttonCss";
 import { type TFunction } from "i18next";
 import { convertWeekday } from "common/src/conversion";
 import { ButtonLikeLink } from "../common/ButtonLikeLink";
@@ -59,8 +63,12 @@ import {
 import { formatDateTimeStrings } from "@/modules/util";
 import { PopupMenu } from "common/src/components/PopupMenu";
 import { useSearchParams } from "next/navigation";
+import { IconButton, StatusLabel } from "common/src/components";
 import type { StatusLabelType } from "common/src/tags";
 import { gql } from "@apollo/client";
+import { LinkLikeButton } from "common/styles/buttonCss";
+import { Sanitize } from "common/src/components/Sanitize";
+import { getReservationUnitAccessPeriods } from "@/modules/reservationUnit";
 
 const N_RESERVATIONS_TO_SHOW = 20;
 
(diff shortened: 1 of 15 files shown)


apps/ui/components/application/ApprovedReservations.tsx
```diff
@@ -753,30 +924,49 @@ function sectionToreservations(
         status: "rejected",
         isCancellableReason: "ALREADY_CANCELLED",
         pk: 0,
+        accessType: AccessType.Unrestricted,
+        accessTypeChanged: false,
```
joonatank, line 928: are these non nullable? would prefer nulls since nothing about the access should render if it's not an existing reservation.

apps/ui/components/application/ApprovedReservations.tsx
```diff
@@ -753,30 +924,49 @@ function sectionToreservations(
...
+        const accessCodeTime =
+          res.pindoraInfo &&
+          formatDateTimeStrings(
+            t,
+            {
+              begin: res.pindoraInfo?.accessCodeBeginsAt,
+              end: res.pindoraInfo?.accessCodeEndsAt,
+            },
+            r.allocatedTimeSlot
+          )?.time;
```
joonatank, line 955: don't like this. what is the type of this? string or boolean or undefined? e.g. don't like 
```
var && ${var}
``` 
syntax since it occasionally creates `bool | string` union. and the type is not easily readable. if you wanna use the syntax define the var type explicitly ex.
```
const accessCodeTime: string | null = ...
```
if using type inference then use ternaries / functions / closures instead.

either wrap this to a function / closure or use a intermediate variable so it's explicit that we are taking only the `time` portion.

the optionals `?` are unnecessary and hurt readability.

apps/ui/modules/reservationUnit.ts
```diff
@@ -626,3 +628,31 @@ function getNotReservableReason(
...
+};
+
+export function getReservationUnitAccessPeriods(
+  accessTypes: AccessTypeDurations[]
+) {
+  return accessTypes.reduceRight<{
+    nextEndDate: string | null;
+    array: AccessTypeDurations[];
+  }>(
+    (acc, aT) => {
```
joonatank, line 647: bit pedantic but it looks cleaner if you move the type definition from inline to `type Var` inside the function body. then the `reduceRight<T>(acc, aT)` all fits a single line.

joonatank (changes requested):

apps/ui/components/application/ApprovedReservations.tsx
vincit-matu: Hmm.. not entirely sure what you're referring to - but `accessType` should always exist for an approved (== existing) reservation as it's derived from the reservation unit's specified access type for the given day. And `accessTypeChanged` always exists as it's checked by the front and not a direct property coming from the reservation data per se (defaulting to `false` if the accessTypes.reduce()` doesn't show a reservation having a different `accessType` than the one before it).

apps/ui/modules/reservationUnit.ts
vincit-matu: Nah, you're right - also makes it more readable IMO. 👍

apps/ui/modules/queries/applicationRound.tsx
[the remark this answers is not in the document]
vincit-matu: Have no idea to be honest 😅 
I guess this must also be a result of some rebase or shuffling around I did in this branch before submitting the PR. Removed.

apps/ui/components/application/ApprovedReservations.tsx
```diff
@@ -601,20 +738,23 @@ function ReservationsTable({
...
+            >
               <span>{toUIDate(date)}</span>
               <OnlyForMobile>
                 {/* span removes whitespace */}
                 <pre style={{ display: "inline" }}>{" - "}</pre>
                 <span>{dayOfWeek}</span>
               </OnlyForMobile>
             </span>
-            {!isDisabled ? (
+            {isDisabled && (
```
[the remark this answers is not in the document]
vincit-matu: accidental, fixed

joonatank (approved):
👍

ranta (approved):

PR #1683, opened 2025-03-28
Don't include access codes in emails if it is not actually active

matti-lamppu:
## 🛠️ Changelog

- After fetching access code for emails, check if it is actually active in Pindora and don't include it if it isn't

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3938](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3938)

[TILA-3938]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/tilavarauspalvelu/integrations/email/template_context/common.py
+++ b/backend/tilavarauspalvelu/integrations/email/template_context/common.py
@@ -261,24 +261,25 @@ def params_for_price_range_info(*, reservation: Reservation) -> dict[str, Any]:
 
 
 def params_for_access_code_reservation(*, reservation: Reservation) -> dict[str, Any]:
+    empty_response = {
+        "access_code_is_used": False,
+        "access_code": "",
+        "access_code_validity_period": "",
+    }
+
     if not reservation.access_code_should_be_active:
-        return {
-            "access_code_is_used": False,
-            "access_code": "",
-            "access_code_validity_period": "",
-        }
+        return empty_response
+
+    empty_response["access_code_is_used"] = True
 
     try:
         response = PindoraClient.get_reservation(reservation=reservation)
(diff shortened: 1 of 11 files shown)


backend/tilavarauspalvelu/integrations/email/template_context/common.py
```diff
@@ -261,24 +261,23 @@ def params_for_price_range_info(*, reservation: Reservation) -> dict[str, Any]:
...
     except PindoraNotFoundError:
-        response = None  # Set as None to make mocking in tests easier
-    if not response:
-        # Reservation should have an access code, but it is not available.
-        return {
-            "access_code_is_used": True,
-            "access_code": "",
-            "access_code_validity_period": "",
-        }
+        return empty_response
```
vergama, line 276: I think the previous version was correct, should return this when the code is not found or is inactive in Pindora but Varaamo thinks it should be active (i.e., `reservation.access_code_should_be_active = True`):
`"access_code_is_used": True, "access_code": ""`
Thus the email will instruct the user about the access code requirement but the actual code value is not shown if it is (unexpectedly) not usable.

vergama (changes requested):
Check comment

backend/tilavarauspalvelu/integrations/email/template_context/common.py
matti-lamppu: Added the same fix to section access codes

vergama (approved):

PR #1696, opened 2025-04-02
Prevent simultaneous overlapping reservation creation

matti-lamppu:
## 🛠️ Changelog

- Make a second overlapping validation check during the endpoint, and remove the created reservation if an overlapping reservation is found
- Make second check also during staff create
- Also do overlapping checks during "return to handling" as this can change a "denied" (not going to happen) to "requires handling" (going to happen), which may make it overlap with newer reservations
- Also also do checks when rescheduling reservations

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3952](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3952)
- [TILA-3956](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3956)

[TILA-3952]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TILA-3956]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
+++ b/backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
@@ -16,7 +16,7 @@ from tilavarauspalvelu.integrations.keyless_entry import PindoraService
 from tilavarauspalvelu.integrations.sentry import SentryLogger
 from tilavarauspalvelu.models import Reservation, ReservationUnit
 from utils.date_utils import DEFAULT_TIMEZONE
-from utils.external_service.errors import ExternalServiceError, external_service_errors_as_validation_errors
+from utils.external_service.errors import ExternalServiceError
 
 if TYPE_CHECKING:
     from tilavarauspalvelu.models import User
@@ -118,16 +118,35 @@ class ReservationCreateSerializer(NestingModelSerializer):
         if prefill_info is not None:
             data.update({key: value for key, value in prefill_info.items() if value is not None})
 
-    @transaction.atomic()
     def create(self, validated_data: ReservationCreateData) -> Reservation:
-        # Must be able to link reservation unit to the reservation
-        reservation_unit: ReservationUnit = validated_data.pop("reservation_unit")
-        reservation: Reservation = super().create(validated_data)
-        reservation.reservation_units.set([reservation_unit])
+        with transaction.atomic():
+            # Must be able to link reservation unit to the reservation
+            reservation_unit: ReservationUnit = validated_data.pop("reservation_unit")
+            reservation: Reservation = super().create(validated_data)
+            reservation.reservation_units.set([reservation_unit])
 
-        # Pindora request must succeed, or the reservation is not created.
+        reservation.begin.astimezone(DEFAULT_TIMEZONE)
+        reservation.end.astimezone(DEFAULT_TIMEZONE)
+
+        # After creating the reservation, check again if there are any overlapping reservations.
+        # This can fail if two reservations are created for reservation units in the same
+        # space-resource hierarchy at almost the same time, meaning when we check for overlapping
+        # reservations during validation, neither of the reservations are yet created.
+        if reservation.actions.overlapping_reservations().exists():
+            reservation.delete()
+            msg = "Overlapping reservations were created at the same time."
+            raise ValidationError(msg, code=error_codes.OVERLAPPING_RESERVATIONS)
+
+        # Pindora request must succeed, otherwise the reservation is removed.
+        # Do this after the second overlapping reservation check, so that we don't need to
+        # remove the access code in Pindora from the removed reservation.
         if reservation.access_type == AccessType.ACCESS_CODE:
-            with external_service_errors_as_validation_errors(code=error_codes.PINDORA_ERROR):
+            try:
                 PindoraService.create_access_code(obj=reservation)
+
+            except Exception as error:
+                reservation.delete()
+                code = error_codes.PINDORA_ERROR if isinstance(error, ExternalServiceError) else None
+                raise ValidationError(str(error), code=code) from error
 
         return reservation
--- a/backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/requires_handling_serializers.py
+++ b/backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/requires_handling_serializers.py
@@ -1,17 +1,23 @@
 from __future__ import annotations
 
-from typing import Any, NotRequired, TypedDict
+from typing import TYPE_CHECKING, Any, NotRequired, TypedDict
 
 from graphene_django_extensions import NestingModelSerializer
 from graphene_django_extensions.fields import EnumFriendlyChoiceField
+from rest_framework.exceptions import ValidationError
 from rest_framework.fields import IntegerField
 
+from tilavarauspalvelu.api.graphql.extensions import error_codes
 from tilavarauspalvelu.enums import AccessType, ReservationStateChoice
 from tilavarauspalvelu.integrations.email.main import EmailService
 from tilavarauspalvelu.integrations.keyless_entry import PindoraService
 from tilavarauspalvelu.integrations.sentry import SentryLogger
 from tilavarauspalvelu.models import Reservation
+from utils.date_utils import DEFAULT_TIMEZONE
 from utils.external_service.errors import ExternalServiceError
+
+if TYPE_CHECKING:
+    from tilavarauspalvelu.models import ReservationUnit
 
 __all__ = [
     "ReservationRequiresHandlingSerializer",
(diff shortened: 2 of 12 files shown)


backend/tilavarauspalvelu/models/reservation/actions.py
```diff
@@ -290,3 +291,12 @@ def refund_paid_reservation(self) -> None:
...
         payment_order.save(update_fields=["refund_id", "status"])
+
+    def overlapping_reservations(self) -> ReservationQuerySet:
+        """Find all reservations that overlap with this reservation."""
+        reservation_unit = self.reservation.reservation_units.first()
+        return Reservation.objects.overlapping_reservations(
+            reservation_unit=reservation_unit,
+            begin=self.reservation.begin,
+            end=self.reservation.end,
+        ).exclude(id=self.reservation.id)
```
ranta, line 302: Since this is an action of Reservation, should we pass buffers from the Reservation to `overlapping_reservations`? If no buffers is passed, the ReservationUnit buffers are used. 🤔

ranta (approved):

backend/tilavarauspalvelu/models/reservation/actions.py
matti-lamppu: Fixed

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
```diff
@@ -118,16 +118,50 @@ def prefill_reservation_from_profile(self, data: ReservationCreateData) -> None:
...
+        try:
+            reservation_unit.validators.validate_no_overlapping_reservations(
+                begin=begin, end=end, ignore_ids=[reservation.id]
+            )
+        except ValidationError:
+            overlapping_reservations = reservation.actions.overlapping_reservations()
+            for overlapping_reservation in overlapping_reservations:
+                #
+                # Remove this reservation if it was created later than any of the overlapping reservations.
+                if reservation.created_at > overlapping_reservation.created_at:
```
vergama, line 144: To be pedantic, I suppose the comparison should strictly speaking be `≥` instead of `>` (is it at least theoretically possible that the timestamps are assigned the same value at the precision supported?) -- but then how to resolve which was first -- could the comparison then be by the `id` instead since the lower should be the first?

In any case I guess the "winner" hinges on which request gets to open the DB transaction first -- and that might not actually be the first request received for processing -- so the outcome is in that sense somewhat arbitrary anyway (outside this test setup which is deterministic).

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
```diff
@@ -118,16 +118,50 @@ def prefill_reservation_from_profile(self, data: ReservationCreateData) -> None:
...
+                    msg = "Overlapping reservations created at the same time. Removed one that was created later."
+                    details = {
+                        "reservation_unit": reservation_unit.pk,
```
vergama, line 147: Could also include the `overlapping_reservation.pk` in the details for Sentry.

vergama:
In fact having thought about this further, I believe that in case of detecting conflicting bookings, both need to be reverted by their respective requests, irrespective of the timestamps or PK values.
That should make the mutation(s) (more) foolproof where the only (?) adverse side effect is that: some of the time no reservation actually gets created.
Assuming then that there's no competing third user for the same time slot, neither of the two users (who both would thus receive the validation error) may figure out to check again if the slot might nevertheless still be available.
That is, however, a situation much like where a second user cancels an overlapping tentative reservation almost immediately and releases the timeslot. And the possibility could be alleviated at the UI level by fine-tuning the wording of the error notification.

(And if in practice these situations are liable to only happen when the bookable units are in high demand, it is likely enough that there is in fact a third user next in the line, who will then snap up the time slot.
Whether that is then an entirely fair outcome is another matter... but neither is the processing of the incoming requests now guaranteed to be performed strictly on a "first come, first served" basis anyway, as the execution happens in parallel outside the PK and current TS assignment in DB transactions.
In practice the time differences are negligible so it can be argued that the "winner" is random and thus fair enough, however the tie-breaking is done.)

In other words, **consider the following scenario** with competing requests **A** & **B** received within milliseconds of each other:
1. **B** begins transaction.
2. **A** begins transaction.
3. **B** is assigned PK and timestamp X.
4. **A** is assigned PK and timestamp Y = X + 1.
5. **A** commits.
6. **A** validates OK: the other reservation **B** not yet visible since uncommitted.
7. **A** succeeds in creating its reservation.
8. **B** commits.
9. **B** validates OK: reservation **A** is at this point returned but it has later TS than **B**: `(B.X > A.Y) <=> False`.
**B** would thus also succeed in creating a reservation. 

Now if the mutation simply reverted (B), in this specific case one reservation would get created (and for that A is just as valid a choice as B); while in other variations both may end up being reverted but it seems like a good trade-off if the possibility of double-booking would then be altogether eliminated. (Is it?)

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
matti-lamppu: Added this.

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
```diff
@@ -118,16 +118,50 @@ def prefill_reservation_from_profile(self, data: ReservationCreateData) -> None:
...
+        begin = reservation.begin.astimezone(DEFAULT_TIMEZONE)
+        end = reservation.end.astimezone(DEFAULT_TIMEZONE)
+
+        # After creating the reservation, check again if there are any overlapping reservations.
+        # This can fail if two reservations are created for reservation units in the same
+        # space-resource hierarchy at almost the same time, meaning when we check for overlapping
+        # reservations during validation, neither of the reservations are yet created.
+        try:
+            reservation_unit.validators.validate_no_overlapping_reservations(
+                begin=begin, end=end, ignore_ids=[reservation.id]
```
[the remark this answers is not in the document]
matti-lamppu: Changed this to call `reservation.actions.overlapping_reservations` which does this.

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/create_serializer.py
matti-lamppu: Changed this to remove regardless, according to your later comment

matti-lamppu:
@vergama Changed the validation to only make a single query, since making multiple queries could also mess this up, and changed logic to always remove the reservation as you suggested.

vergama (approved):
Reasoning now described in https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/9614819329/P+llek+isten+varausten+est+minen

matti-lamppu:
Added checks to "staff create" and "requires handling" endpoint

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/staff_create_serializers.py
```diff
@@ -160,6 +162,16 @@ def create(self, validated_data: StaffCreateReservationData) -> Reservation:
             reservation: Reservation = super().create(validated_data)
             reservation.reservation_units.set([reservation_unit])
 
+        # After creating the reservation, check again if there are any overlapping reservations.
+        # This can fail if two reservations are created for reservation units in the same
+        # space-resource hierarchy at almost the same time, meaning when we check for overlapping
+        # reservations during validation, neither of the reservations are yet created.
+        overlapping_reservations = list(reservation.actions.overlapping_reservations().values_list("pk", flat=True))
+        if overlapping_reservations:
```
ranta, line 170: Or simply just `reservation.actions.overlapping_reservations().exists()` ?

ranta (approved):

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/staff_create_serializers.py
matti-lamppu: Yes, previously had Sentry logging here but took it our as unnecessary.

backend/tilavarauspalvelu/api/graphql/types/reservation/serializers/requires_handling_serializers.py
```diff
@@ -45,12 +51,32 @@ class Meta:
         ]
 
     def validate(self, data: ReservationHandlingData) -> ReservationHandlingData:
+        begin = self.instance.begin.astimezone(DEFAULT_TIMEZONE)
+        end = self.instance.end.astimezone(DEFAULT_TIMEZONE)
+
+        if self.instance.state == ReservationStateChoice.DENIED:
```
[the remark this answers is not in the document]
matti-lamppu: Fixed.

matti-lamppu:
Added post-mutation checks to adjust time mutations as well

vergama (approved):

PR #1769, opened 2025-04-28
fix: make the purpose text able to wrap in view application page

vincit-matu:
## 🛠️ Changelog

- edit the list styling to keep the medium font weight in the info values, while having only ones within a `<span>` as `nowrap`.
- `nowrap` _could_ in principle be removed from all the application info values, but in the other ones there is good a case to be made for keeping them confined to be on a single line. Since purpose is a free text field it can be a lot longer, and it’s quite natural to have it wrap onto two or more lines if needed due to it being natural text and not "hard values" (== a number followed by a unit, or a range of dates..). The other values can not be long enough to warrant a second line by themselves. If they wouldn't fit on one line together with their label the entire value text should be on a new line.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3960](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3960)

[TILA-3960]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/ui/components/application/ApplicationSectionList.tsx
+++ b/apps/ui/components/application/ApplicationSectionList.tsx
@@ -20,6 +20,7 @@ import {
   InfoItemContainer,
   InfoItem,
   ScheduleDay,
+  RegularText,
 } from "./styled";
 import { SuitableTimeRangeFormValues } from "./form";
 import { WEEKDAYS } from "common/src/const";
@@ -33,6 +34,7 @@ import {
   Tooltip,
 } from "hds-react";
 import { getDayTimes } from "./module";
+import { NoWrap } from "common/styled";
 
 function formatDurationSeconds(seconds: number, t: TFunction): string {
   const hours = Math.floor(seconds / 3600);
@@ -80,11 +82,17 @@ function getLabelProps(
   }
 }
 
-function InfoListItem({ label, value }: { label: string; value: string }) {
+function InfoListItem({
+  label,
+  value,
+}: {
+  label: string;
+  value: string | JSX.Element;
+}) {
   return (
     <li>
       <h4>{`${label}: `}</h4>
-      <span>{value}</span>
+      {value}
     </li>
   );
 }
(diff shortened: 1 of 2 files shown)


apps/ui/components/application/styled.tsx
```diff
@@ -117,14 +117,17 @@ export const InfoItem = styled(Flex).attrs({
...
+
   li {
     h4 {
       display: inline;
       ${fontRegular};
     }
     span {
       white-space: nowrap;
-      ${fontMedium};
     }
```
joonatank, line 131: This isn't how we typically do this (yes it's existing code).

Remove the custom CSS that styles all `span` elements and instead use `NoWrap` to explicitly show what should be `nowrap` and not a regular `span`.

apps/ui/components/application/ApplicationSectionList.tsx
joonatank, line 1: replace `span` with `NoWrap` from 
```
import { NoWrap } from "common/styled";
```
so the styling is explicit in the JSX without reading the CSS.

apps/ui/components/application/ApplicationSectionList.tsx
```diff
@@ -194,7 +210,7 @@ function SingleApplicationSection({
             <ol>
               {filterNonNullable(reservationUnits).map((ru) => (
                 <li key={ru.pk}>
-                  {getTranslationSafe(ru, "name", lang).trim()}
+                  <h4>{getTranslationSafe(ru, "name", lang).trim()}</h4>
```
joonatank, line 213: this looks wrong. is there a body text for the `h4` section somewhere? if not and you just want the styling use `H4 as="span"`

apps/ui/components/application/ApplicationSectionList.tsx
```diff
@@ -80,11 +80,17 @@ function getLabelProps(
...
+function InfoListItem({
+  label,
+  value,
+}: {
+  label: string;
+  value: string | JSX.Element;
+}) {
   return (
     <li>
       <h4>{`${label}: `}</h4>
```
joonatank, line 92: don't think `h4` is appropriate here (creates a section instead of just defining a key - value pair), but wasn't part of the spec.

joonatank (changes requested):
replace the hierarchical css overrides with a specified NoWrap component instead.

apps/ui/components/application/ApplicationSectionList.tsx
vincit-matu: Yeah, this is a style issue. In other words your assumption (that the use of `<h4>` here is due to styling) is correct.
But it looks like using the custom(?) `H4` component fucks up the styles - however using simply `<h4 as="span">` seems to do the trick.

joonatank (approved):

ranta (approved):

PR #1783, opened 2025-05-06
Add "access code added" emails

matti-lamppu:
## 🛠️ Changelog

- Send emails when access codes are added or activated on reservations or seasonal bookings
- New email templates:
  - Reservation access code added (similar to "reservation access code changed", but with different text)
  - Seasonal booking access code added (similar to seasonal booking access code changed", but with different text)
- Only include series' that have access codes associated with them in access code emails
- Only include series' that were cancelled or rescheduled in those emails
- Include reservation unit information in series and application section emails
- Refactor context functions to use simpler typing

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3973](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3973)
- [TILA-4007](https://helsinkisolutionoffice.atlassian.net/browse/TILA-4007)

[TILA-3973]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vergama (approved):
Looks good; I think it's OK to skip sending the series emails if they have no reservations.

matti-lamppu:
@vergama My concern there was if we should send a reschedule email when rescheduling results in an empty series, though I'm not sure if that is allowed?

vergama:
> @vergama My concern there was if we should send a reschedule email when rescheduling results in an empty series, though I'm not sure if that is allowed?

At least in `send_seasonal_booking_denied_series_email` (L600) it seems like it should be OK, since if there are actually no reservations, then nothing could in fact have been rejected, so the email would seem quite pointless.

The rescheduling scenario (L663) may be more complicated, though... Currently the frontend should prevent that kind of rescheduling from taking place, but I guess it's at least in theory a possible edge case.
Should that be so, then the email might be desirable so that the user is somehow notified about changes having taken place. At least the user would then presumably look up the seasonal booking in the UI and be alerted to the fact that there are no reservations left (while the changed weekday / time range / etc. would be displayed in the table of allocations for the application section).

(This reminds me that the `prune_recurring_reservations_task` is now probably obsolete since the recurring reservation mutations, including creation, are atomic; if rescheduling ends up with no reservations, it's better to keep the info about the recurrence itself. Separate thing, though.)

ranta (approved):
Good stuff especially with context typing 🚀

PR #1806, opened 2025-05-14
Add some pending orders to handled paid reservation in test data

matti-lamppu:
## 🛠️ Changelog

- When creating reservations for local test data, if the reservation unit would be paid, but also handled and approved, but not paid on site, create its order as pending with a "handled payment due at" value

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-4031](https://helsinkisolutionoffice.atlassian.net/browse/TILA-4031)


backend/tilavarauspalvelu/management/commands/data_creation/create_reservations.py
```diff
@@ -191,25 +194,32 @@ def _create_normal_reservations(
...
-
-        if reservation_unit.require_reservation_handling:
-            reservation_state = weighted_choice(handling_state_choices, weights=[2, 5, 1])
-            if reservation_state in {ReservationStateChoice.CONFIRMED, ReservationStateChoice.DENIED}:
-                handled_at = local_datetime()
+        handled_payment_due_by: datetime.datetime | None = None
 
         for reservation_date in get_date_range(start_date, number=14):
+            if reservation_unit.require_reservation_handling:
+                reservation_state = next(handling_state_choices_cycle)
```
ranta, line 201: Why not use `weighted_choice(handling_state_choices, [2, 5, 1])` here? 🤔

ranta (approved):

backend/tilavarauspalvelu/management/commands/data_creation/create_reservations.py
matti-lamppu: Just a bit less randomness. Should be mostly the same.

vergama (changes requested):
Also need some handled reservations where the due by deadline of the payment is already past, with statuses then being either:
1. `EXPIRED` on the payment order and `CANCELLED` on its reservation -- the frontend can then be implemented against those regardless of the periodic tasks.
2. `PAID` (or `PAID_BY_INVOICE` as the case may be) and `CONFIMED`, respectively -- I guess it's good to have these variants as well. (They should have the mock webshop data, to boot.)
(Other combinations probably not necessary: denied but already paid, etc.)

matti-lamppu:
@ranta @vergama Rewrote logic for cancelling and denying reservations, as well as their payment orders. This should result in handled paid reservations in cancelled, expired, and paid states in the test data.

PR #1807, opened 2025-05-14
Single payment order per reservation

matti-lamppu:
## 🛠️ Changelog

- Make reservation to payment order relation one-to-one
- Fix some django admin panel views that lag due to too many reservations/reservation units/users in prod

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- Breaks frontend queries, need to refactor

## 🎫 Tickets

- [TILA-4032](https://helsinkisolutionoffice.atlassian.net/browse/TILA-4032)

[TILA-4032]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-4032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matti-lamppu:
Test and production has some reservations with multiple payment orders. Should find out why this is, and if this change will break some edge case I'm not aware of.

vergama (approved):
The migration logic should be OK: the dozen reservations in production with multiple orders are nearly all from the 1-2 April 2025 peak usage times with one paid version and one unpaid, expired version, -- or in two cases, explicitly cancelled, and in another two cases, two expired variants -- with most of them having been created within a short time.
In one case the time difference between their creation was more than 10 seconds but that doesn't seem to make a difference.

In any case, for all 12 reservations found in production (with duplicate webstore orders), the general situation is the same: user was able to initiate "confirming" a tentative reservation from DRAFT the WAITING_FOR_PAYMENT state twice about the same time (while the processing of the mutation may have taken an abnormally long time in the backend) and two webstore orders were generated as a result.
Even so, only one instance has been paid and the duplicates in un-PAID state are just bogus information which can be swept away.

PR #1828, opened 2025-05-23
Fix series and section `update_access_code_is_active` logic

matti-lamppu:
## 🛠️ Changelog

- Run reschedule in access code active status update to cover situations where access type changes from access code to something else in the middle of a series. In this case, some validity times should be removed, but the core remains active, so no email should be sent.

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- None


matti-lamppu:
@vergama @ranta Actually, this is probably still not correct. User should recieve an email if an access code is added to another reservation unit in their seasonal booking while they already have another unit which uses access codes.

PR #1867, opened 2025-06-11
Add non-null constraints to some Model fields

matti-lamppu:
## 🛠️ Changelog

- Makes some model fields non-null:
  - `Application.user`
  - `ReservationSeries.user`
  - `Reservation.user`

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3949](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3949)


backend/tests/test_graphql_api/test_application/test_filtering.py
```diff
@@ -232,14 +232,16 @@ def test_application__filter__by_text_search__section_name(graphql):
         organisation_name="",
         contact_person_first_name="",
         contact_person_last_name="",
-        user=None,
+        user__first_name="",
+        user__last_name="",
```
ranta, line 236: Would it be better to have this in the factory method as default values?

matti-lamppu: Nah, these are special case tests where we need to know that the other values don't contribute to the text search.

matti-lamppu:
> There is a ruff rule DJ001 that checks for those, but it doesn't recognise errors when the field is manually typed (e.g. name: str = models.CharField(null=True))

@ranta There are still a 4 nullable `CharFields` (`Unit.tprek_id`, `Unit.tprek_department_id`, `ReservationStatistics.reservation_type`, `ReservationStatistics.primary_unit_trpek_id`) and 2 nullable `CharFields` with choices (`Reservation.type`, `Reservation.cancel_reason`)

PR #1868, opened 2025-06-11
More model refactoring

matti-lamppu:
## 🛠️ Changelog

- Remove default `TaxPercentage` from `ReservationUnitPricing`

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

## 🎫 Tickets

- [TILA-3949](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3949)


vergama:
> Looking into production:
> Reservation.objects.filter(reservee_is_unregistered_association=False, reservee_id="", reservee_type="NONPROFIT").count()
> 2665
> Reservation.objects.filter(reservee_is_unregistered_association=True, reservee_type="NONPROFIT").exclude(reservee_id="").count()
> 9
> These states should be impossible, how to migrate?

I suspect that the rarer, latter case may be due to the form having first been filled (for a `NONPROFIT`) with a `reservee_id` value, then the checkbox has been checked but front has nevertheless included the value in the update mutation -- in which case the correct recourse would be to remove the bogus `reservee_id` in migration, since in UI the `reservee_is_unregistered_association` field is supposed to take precedence. Indeed that scenario can be reproduced even now in the test environment, but will investigate further nevertheless.

The more numerous first case is harder to explain... since we (obviously) can't make up the `reservee_id`s _ex nihil_ even if they should be there (i.e., assuming that in the past the customer UI would have failed to validate the fields correctly), the naive migration would be just to leave them as-is (with the `reservee_is_unregistered_association` field being removed).
One possibility is that all of them are staff-created since the form fields are not really mandatory in the Admin-ui, although one would maybe expect a bit smaller count of such reservations -- so needs some investigation as well.

backend/tilavarauspalvelu/models/reservation/actions.py
```diff
@@ -240,33 +240,20 @@ def get_state_on_reservation_confirmed(self, payment_type: PaymentType | None) -
...
 
         # Some fields are never mandatory for an individual reserver even if marked so in the metadata.
-        if reservee_type == CustomerTypeChoice.INDIVIDUAL:
-            qs = qs.exclude(field_name__in=["home_city", "reservee_id", "reservee_organisation_name"])
+        if reservee_type == ReserveeType.INDIVIDUAL:
+            qs = qs.exclude(field_name__in=["reservee_identifier", "reservee_organisation_name"])
 
-        # `reservee_id` is not mandatory for unregistered associations.
-        if reservee_is_unregistered_association:
-            qs = qs.exclude(field_name__in=["reservee_id"])
```
vergama, line 261: Shouldn't `reservee_id` still be removed from required fields if `type == NONPROFIT`?

vergama:
> Reservation.objects.filter(reservee_is_unregistered_association=False, reservee_id="", reservee_type="NONPROFIT").count()
> 2665

Only 3/2665 are `NORMAL` reservations and the rest are staff-created so that the fact that the field is not mandatory in the Admin-UI form must explain the greater part.
The 3 remaining cases appear to have been modified afterwards through Django admin. In both cases, it's then OK that `reservee_id` will be empty and that if viewed after the refactoring, treated as reservations for _unregistered_ nonprofit.

> Reservation.objects.filter(reservee_is_unregistered_association=True, reservee_type="NONPROFIT").exclude(reservee_id="").count()
> 9

At least some of them probably are explained by the user changing mind and checking "unregistered" on the UI form after first filling something in the org.id field. Yet about half have a plainly valid org.id value so it's hard to be sure; perhaps they too have been fiddled with in Django and/or there has been a bug in the UI, so one is wary of removing the ids, even though if they were now viewed in the Admin-UI the org.ids would not be shown.
Probably doesn't really matter one way or another since all 9 reservations are in the past, but I'd err on the side of caution and keeping the org.ids for all of them (despite this leading to the reservations being treated as for _registered_ nonprofits). One could make a distinction between the (more or less) valid ids and filler ones by keeping only those that match `*******-*` but that's probably needlessly over-complicating it.

vergama (changes requested):
Check comments (_re_ migration and otherwise)

backend/tilavarauspalvelu/models/reservation/actions.py
matti-lamppu: Yes, fixed.

vergama (approved):

PR #1953, opened 2025-08-13
Small fixes to Admin UI router refactor

ranta:
## 🛠️ Changelog

- Small non-critical fixes to Admin UI, see commit messages for details

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- None

### DIFF
--- a/apps/admin-ui/src/component/MainLander.tsx
+++ b/apps/admin-ui/src/component/MainLander.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import styled from "styled-components";
 import { useTranslation } from "next-i18next";
 import { signIn } from "common/src/browserHelpers";
@@ -8,6 +8,7 @@ import { breakpoints } from "common/src/const";
 import { HERO_IMAGE_URL } from "@/common/const";
 import { KorosHeading, Heading } from "./KorosHeading";
 import { getLocalizationLang } from "common/src/helpers";
+import { useRouter } from "next/router";
 
 const LoginBtn = styled(Button)`
   --background-color: var(--color-white);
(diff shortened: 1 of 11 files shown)


apps/admin-ui/src/component/MainLander.tsx
```diff
@@ -45,6 +46,15 @@ const Ingress = styled(H2)`
...
   const { t, i18n } = useTranslation();
 
+  // Redirect to root url if not already there, useful e.g. in /auth/logout page
+  const router = useRouter();
+  useEffect(() => {
+    if (router.pathname !== "/") {
+      router.replace("/");
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
joonatank, line 56: this should not require effect, but might require reworking `/auth/logout`, since it should redirect from either middleware / SSR after removing session cookie.

this is fine for now, but should be reworked later if this is added. (rework `/auth/logout` so this is not needed).

packages/common/src/components/SearchButton.tsx
```diff
@@ -27,6 +27,7 @@ export function SearchButton({ isLoading = false, labelKey = "filters:searchButt
   return (
     <Flex $direction="row" $justifyContent="flex-end">
       <Button
+        id="searchButton"
```
joonatank, line 30: I'd prefer if Risto could use something else for this, even if it's `data-testid`. HTML should not have duplicated element ids and hard coding it so that there is ever only one `SearchButton` component in the whole document is bad.

not a fan of over using id props either because it just moves the problem down the line. and there is no error reporting for duplicated ids even though it's not correct.

joonatank (commented):
Good 👍

apps/admin-ui/src/component/MainLander.tsx
ranta: The issue right now is that the `/auth/logout/` page is not rendered at all due to permission checks and user is not redirected to `/`, which this is a workaround for.

I will add a ticket to rework this later

joonatank: taking a quick look at `ui/auth/logout` why not just copy it? or add `...getCommonServerSideProps()` to the admin one (it's missing it). this should not be needed after.

PR #1972, opened 2025-08-25
Use new opening hours redirect-url

vincit-matu:
## 🛠️ Changelog

- Uses the new redirect url provided by the backend to edit opening hours (`/v1/edit_opening_hours`)
- Makes admin reservation-unit table `selectable` (using the inbuilt functionality provided by HDS)
- Adds `EditOpeningHoursBar` to the reservation unit page
- Moves `useToastIfQueryParam` to `common` so it can be used in `admin-ui`
- Uses `useToastIfQueryParam` to show possible errors from the new redirect endpoint on reservation unit page
- Small fixes to the `css` in our `Table` component, to show the empty heading (from HDS:s `selectable` Table) correctly

## 🧪 Test plan

- Automated tests

## 🚧 Dependencies

- This needs the functionality provided in #1967 (which it is based on) to work, otherwise the redirect url doesn't work. So don't merge to `main`, at least until #1967 is merged. 😅 

## 🎫 Tickets

- [TILA-4169](https://helsinkisolutionoffice.atlassian.net/browse/TILA-4169)

[TILA-4169]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-4169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/admin-ui/src/common/urls.ts
```diff
@@ -111,3 +111,37 @@ export function getNotificationUrl(pk: Maybe<number> | undefined): string {
   }
   return `${BANNER_NOTIFICATIONS_URL_PREFIX}/${pk}`;
 }
+
+export function getOpeningHoursUrl(
+  apiBaseUrl: string,
+  reservationUnitPk: number | number[] | null,
+  errorUrl: URL = new URL(window.location.pathname, window.location.origin)
```
joonatank, line 118: this is ran before the `window.location` check so it will throw error from here and not the first lines of the function.

I'd prefer to make this SSR compatible if possible and remove the default values here (move them to the caller and explicitly set them).

primary reasoning is that this doesn't require `window` to function outside of these defaults, but `window` will break on SSR.

apps/admin-ui/src/common/urls.ts
```diff
@@ -111,3 +111,37 @@ export function getNotificationUrl(pk: Maybe<number> | undefined): string {
...
+): string {
+  if (window?.location == null) {
+    throw new Error("window.location is not available, cannot build redirect url");
+  }
```
joonatank, line 122: this function has double error mechanics (see above comment on SSR). on nodejs / SSR it will throw, but invalid redirect urls on browser will return `""` and not throw.

not a huge issue but is a bit iffy (e.g. make it SSR compatible would fix it), throwing all errors would be fine also, but requires actually catching them client side. returning empty strings for all errors would also work (may cause hydration errors down the line) this would probably be the 2nd best approach (SSR compatibility being the best) if missing the link (e.g. empty string) is not a critical problem.

apps/admin-ui/src/lib/reservation-units/ReservationUnitsTable.tsx
```diff
@@ -1,6 +1,7 @@
-import React from "react";
+import type { SelectedRow } from "@/pages/reservation-units";
```
joonatank, line 1: weird import, should be the other way around (even though it's type). `@/pages` import from `@/lib` nobody imports from `@/pages`

apps/admin-ui/src/lib/reservation-units/ReservationUnitsTable.tsx
```diff
@@ -149,11 +158,18 @@ export function ReservationUnitsTable({
...
       rows={reservationUnits}
       cols={cols}
       initialSortingColumnKey={sort.startsWith("-") ? sort.slice(1) : sort}
       initialSortingOrder={sort.startsWith("-") ? "desc" : "asc"}
       isLoading={isLoading}
+      checkboxSelection
+      ariaLabelCheckboxSelection={t("common:select")}
+      selectedRows={selectedRows}
+      selectAllRowsText={"Valitse kaikki"}
+      clearSelectionsText={"Tyhjennä valinnat"}
```
joonatank, line 171: not a fan of leaving text strings inside code since we use translations for everything else (just a consistency thing).

apps/admin-ui/src/lib/reservation-units/[pk]/OpeningHoursSection.tsx
```diff
@@ -29,8 +31,8 @@ export function OpeningHoursSection({
           <p style={{ gridColumn: "1 / -1" }}>{t("reservationUnitEditor:openingHoursHelperTextHasLink")}</p>
           {/* TODO this should be external? i.e. standard a link */}
           <ButtonLikeLink
-            disabled={!reservationUnit?.haukiUrl}
-            href={reservationUnit?.haukiUrl ?? ""}
+            disabled={!apiBaseUrl || !reservationUnit?.pk}
+            href={getOpeningHoursUrl(apiBaseUrl, reservationUnit?.pk)}
```
Comment (joonatank) [inline apps/admin-ui/src/lib/reservation-units/[pk]/OpeningHoursSection.tsx:35]: prefer making a temporary variable out of the url and use that instead of having two checks that check different things. e.g. `getOpeningHoursUrl` can return empty string while `disabled={!apiBaseUrl || !reservationUnit?.pk} === disabled={false}`

apps/admin-ui/src/pages/reservation-units/[pk]/index.tsx
```diff
@@ -187,6 +192,12 @@ function ReservationUnitEditor({
   });
   const unit = reservationUnit?.unit ?? unitData?.unit;
 
+  useToastIfQueryParam({
+    key: ["error_code", "error_message"],
+    message: params.get("error_message") ?? t("reservationUnit:editErrorMessage"),
```
Comment (joonatank) [inline apps/admin-ui/src/pages/reservation-units/[pk]/index.tsx:197]: don't like this.

general logic inside a page where it feels like it should be inside the hook itself (and maybe requires refactoring other hook users).

can we say it's always `error_message` param and make it default so `useToastIfQueryParam` uses `error_message` param first, if no such param exists then it uses the `message` param given to the hook. is there a case where this default breaks something?

if it does, I'd rather have a param with a list of keys (empty by default) for what params to listen to for the error_message. then the key could also be extended inside the hook to include both the user defined key + `error_message` (or a list of keys to look for the message).

joonatank (changes requested):

apps/admin-ui/src/lib/reservation-units/ReservationUnitsTable.tsx
vincit-matu: Good catch - I completely agree

PR #2000, opened 2025-09-03
Consolidate all date formatting to common date-utils

vincit-matu:
## 🛠️ Changelog

- Introduces all new date formatting functions (`common/src/date-utils`)
- Consolidates all date formatting related operations in the codebase's front end

### Previously, date utility functions were scattered across multiple files:
- `apps/admin-ui/src/helpers/date.ts` - Admin-specific helpers
- `apps/admin-ui/src/common/util.ts` - Admin utilities with i18n
- `apps/ui/modules/util.ts` - UI utilities with different i18n approach
- `packages/common/src/common/util.ts` - Shared utilities (partial)
- use of `date-fns`'s `format()` (instead of utilizing helper-functions) scattered around the front end packages

This consolidation moves all date/time related helper functions to `common/src/date-utils/`, and refactors all such formatting to use said functions. It also attempts to use the i18n-functionality of `date-fns` for all standard adherent translations/formats, which reduces the need of `i18n-next`'s translation function with custom translation values (they are far more brittle == could introduce problems in the future).

### This consolidation provides:
- **Single source of truth** for date operations
- **Consistent error handling** across all functions  
- **Better TypeScript support** with proper types
- **Unified i18n approach** for formatting functions
- **Comprehensive documentation** with examples

### Architecture

```
date-utils/
├── index.ts          # Main exports
├── conversion.ts     # API/UI date format conversion
├── construction.ts   # DateTime construction helpers
├── formatting.ts     # Display formatting with i18n
└── types.ts          # Shared TypeScript types
```

Rename/get rid of functions named `to/from...()` for clarity, resulting in:
- functions starting with `format...()` return a displayable string (e.g. input a `Date` -> returns a `dd.mm.yyyy` format string)
- functions starting with `parse...()` can return a value in any required type (e.g. input a date `string` -> return a `Date` object)

## 🧪 Test plan

- Automated tests will probably catch some problems, but since there are so many changes in so many places it would probably be very useful to go through all use-case flows with the old and new versions side-by-side to be able to verify that the formatting is in fact identical to how it used to be.

## 🚧 Dependencies

- The branch is based on the `undine` switch branch. While the changes introduced in this PR would in principle work with the old GQL-libraries too, at least a bit of rebasing would be necessary to merge this to `main` without it containing the changes that switching to `undine` causes in the front end.

## 🎫 Tickets

- [TILA-3926](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3926)
- [TILA-4218](https://helsinkisolutionoffice.atlassian.net/browse/TILA-4218)

[TILA-3926]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/admin-ui/src/component/ReservationsList.tsx
+++ b/apps/admin-ui/src/component/ReservationsList.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { formatDateRange } from "common/src/date-utils";
 import styled from "styled-components";
 import { useTranslation } from "next-i18next";
 import { startOfDay } from "date-fns";
@@ -11,6 +10,7 @@ import { useModal } from "@/context/ModalContext";
 import { H6 } from "common/styled";
 import StatusLabel from "common/src/components/StatusLabel";
 import { gql } from "@apollo/client";
+import { formatDate } from "common/src/date-utils";
 
 export type NewReservationListItem = {
   date: Date;
@@ -218,6 +218,8 @@ export function ReservationList(props: Props | ExtendedProps) {
 
   const removed = items.filter((x) => x.isRemoved).length;
   const count = items.length - removed;
+  const formatReservationSlot = (item: NewReservationListItem) =>
+    `${formatDate(item.date, { includeWeekday: true })}, ${item.startTime}-${item.endTime}`;
   return (
     <ListWrapper data-testid="reservations-list" $isTall={isTall}>
       <TitleWrapper>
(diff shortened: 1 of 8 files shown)


packages/common/src/date-utils/types.ts
```diff
@@ -0,0 +1,38 @@
...
+export type TimeStruct = {
+  hours?: number | null;
+  minutes?: number | null;
+  seconds?: number | null;
+};
+
+export interface CommonDateOptions {
+  t?: TFunction;
+  locale?: LocalizationLanguages | "fi" | "sv" | "en";
+  formatString?: string;
```
joonatank, line 13: i'd prefer removing this so user can't do something stupid like `const dayString = formatDateTime(new Date(), "HH")` never mind `formatDateTime(new Date(), "")` reduces error surface.

instead using an enum / string literal to switch between predetermined formats.

joonatank (changes requested):

packages/common/src/date-utils/types.ts
vincit-matu: Removed. Using abstractions where necessary.

PR #2469, opened 2026-07-28
refactor: pre-commit auto update compatibility

voneiden:
Pre-commit's auto update applies unsafe fixes via ruff, which ends up
removing a field assignment line after the type definition.
Attempting to work around this by combining type definition with the
assignment on a single line.

The type definition is incorrect, but that is beyond the scope of this.

Refs: TILA-4416

The pre-commit autoudpate in question is here: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/2456

## Summary by CodeRabbit

* **Refactor**
  * Reformatted image field declarations to use inline type-annotated assignments for clearer consistency.
  * Kept the same upload targets and form/nullable behavior.
  * No user-visible behavior or configuration changes.

### DIFF
--- a/backend/tilavarauspalvelu/models/intended_use/model.py
+++ b/backend/tilavarauspalvelu/models/intended_use/model.py
@@ -30,8 +30,9 @@ class IntendedUse(models.Model):
 
     name: str = models.CharField(max_length=200)
 
-    image: ThumbnailerImageFieldFile | None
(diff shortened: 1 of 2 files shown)


danipran (approved):
Not a pre-commit issue but rather a style/ruff configuration issue, [unsafe fixes are on](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/cfff8303c356b330f2bd73fb823253dbcedcb755/backend/pyproject.toml#L95-L97) so pre-commit applies unsafe fixes. I think the fix is still correct, but the commit message should be modified (e.g. "style: ruff fixes" or something, idk)

PR #2481, opened 2026-09-03
Tila-4426 & Tila-4425 sonar reliability issues

timwessman:
Closes the 16 open reliability issues SonarCloud reports for
tilavarauspalvelu-staff-ui and tilavarauspalvelu-customer-ui. The
backend project was already clean.

Three commits, easiest reviewed one by one:

1. **`getOpeningHoursUrl` dropped its filter result** — the only real bug
   here. Non-positive reservation unit pks were still sent, and the
   emptiness check tested the unfiltered list. Adds tests for the
   function, which had none.
2. **`Number`, `Error` and `Map` renamed** — they shadowed the globals of
   the same name. The map component file is renamed to match, which also
   removes an import alias at its only consumer.
3. **Callbacks no longer passed directly to `Array.map`** — no behaviour
   change (none of them takes a second parameter), so this is rule
   compliance. Applied at all 37 sites in the frontend rather than only
   the 12 reported, to keep the pattern consistent.

Two things left alone on purpose: `.map(Number)`, because oxlint requires
native coercion functions to be passed directly, and test files, which
Sonar excludes from this rule.

Refs: TILA-4426

## Summary by CodeRabbit

- **Bug Fixes**
  - Opening-hours links now exclude invalid reservation unit IDs and return no link when no valid IDs are provided.
  - Reservation unit maps continue to display with the correct language and sizing, including the default height.

- **Tests**
  - Added coverage for opening-hours links, reservation unit maps, filter handling, and application-round time-slot calculations.

- **Refactor**
  - Improved consistency across option and data transformations without changing user-facing behavior.

### DIFF
--- a/frontend/apps/customer/next-env.d.ts
+++ b/frontend/apps/customer/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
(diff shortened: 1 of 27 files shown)


tjgofore (approved):

tjgofore:
LGTM. Quality gate still fails.

